PR #7, opened 2023-04-06
Add swagger API file 

daria305:
Defines the  Iota Core APIs list based on the Hornet APIs. [Issue](https://github.com/iotaledger/iota-core-roadmap/issues/1)


documentation/APIs/openapi3-core.yaml
jonastheis, line 1: `/info`: here we should check whether eg metrics is needed as it currently is.

`/blocks/{blockID}`: empty payload example is missing some fields (signature, commitment, etc)

`/blocks/{blockID}/metadata`: here the examples could explicitly distinguish between a block without and one with TX. 
`blockStateReason` should state explicitly that we will add more reasons here.

`/transactions/{transactionId}/included-block` and `/metadata` can't we just reuse the exact same responses from the corresponding `blocks` endpoints?

`/commitment/{commitmentId}`: Do we just return the plain commitment? Or do we add some more information like the merkle tree of the rootsID ? How else could an application do PoI or something?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+    </td>
+  </tr>
+  <tr>
+    <td>Issuer ID</td>
+    <td>ByteArray[32]</td>
+    <td>The issuer identifier</td>
+  </tr>
+  <tr>
+    <td>Issuing Time</td>
+    <td>int64</td>
```
thibault-martinez, line 128: Does this need to be a signed integer?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+    <td>The time the block is issued.</td>
+  </tr>
+  <tr>
+    <td>Slot Commitment Length</td>
```
thibault-martinez, line 132: What is this length for? A `Slot Commitment` size doesn't look dynamic.

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+        <summary>Slot Commitment</summary>
+        <blockquote>
+          Slot Commitment is an object that contains a summary of a slot. More descriptions are in Slot Commitment section.
+        </blockquote>
+      </details>
+    </td>
+  </tr>
+  <tr>
+    <td>Latest Finalized Slot</td>
+    <td>int64</td>
```
thibault-martinez, line 149: Does this need to be a signed integer?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+    <td>The slot index of latest finalized slot.</td>
+  </tr>
+  <tr>
+    <td>Payload Length</td>
+    <td>uint32</td>
+    <td>The length of the following payload in bytes. A length of 0 means no payload will be attached.</td>
+  </tr>
+  <tr>
+    <td valign="top">Payload <code>oneOf</code></td>
```
thibault-martinez, line 158: Should be `optOneOf` ?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+  <tr>
+    <td>Nonce</td>
+    <td>uint64</td>
+    <td>The nonce which lets this block fulfill the PoW requirement.</td>
+  </tr>
+</table>
+
+
+### Block ID
+Block ID denotes an identifier of a block, with type `ArrayBytes[40]`. It is calculated as the following steps, using BLAKE2b-256 hash function:
```
thibault-martinez, line 220: Out of curiosity, has blake3 been considered for protocol v3?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+
+* `content` is the serialized block **without** signature and nonce.
+* `slot_index` is the slot index of the `Issuing Time` of the block.
```
thibault-martinez, line 223: Is that the index in the Slot Commitment struct?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+* `signatureBytes` is the serialized Ed25519 signature.
```
thibault-martinez, line 224: Signature is supposed to be generic so maybe don't mention Ed25519 here?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+2. `id` = hash(Concat(`content_hash`, signatureBytes))
+3. And finally, `BlockID` = Concat(`id`, `slot_index`) 
+
+The string format of Block ID is hexadecimal encoding of Block ID with `0x` prefix.
+
+### Block Syntactic Validation Rules
+
+* Block size must not exceeds `MaxBlockSize` bytes, currently is `32768`.
+* `Protocol Version` must match the `Protocol Version` config parameter of the node.
+* Must have at least **1** `Strong Parents`.
```
thibault-martinez, line 237: Just to be sure, this means we can have no weak and no shallow like parents? To disambiguate, I'd give an actual allowed range for each of them.

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+* `Strong Parents`, `Weak Parents`, `Shallow Like Parents` must not exceeds `BlockMaxParents` (currently is **8**) per type.
+* `Strong Parents`, `Weak Parents`, `Shallow Like Parents` of parents must be lexically ordered.
```
thibault-martinez, line 239: Remove `of parents` ?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+* `Strong Parents`, `Weak Parents`, `Shallow Like Parents` must not have duplicates in each list.
+* `Weak Parents` must be disjunct to the rest of the parents.
```
thibault-martinez, line 241: So no weak parent at all should also be in strong or shallow like?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+* `Payload Type` must match one of the values described under Payloads.
+  Data Fields must be correctly parsable in the context of the `Payload Type`.
+  The payload itself must pass syntactic validation.
+* `Signature` must be valid, the input of a signature is concatation of Issuing Time bytes, Commitment ID bytes and Block bytes (without Signature and Nonce) by order.
```
thibault-martinez, line 245: Where is `Commitment ID` from?

thibault-martinez, line 245: Isn't `Issuing Time` included twice in the signature then ?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+## Slot Commitment
+
+Commitment is an object that contains a summary of a slot, and it is linked to the commitment of the previous slot, which forms a commitment chain. 
+
+Multiple sparse merkle trees are managed by per slot, which are:
+* *Tangle tree*: All accepted blocks within a slot.
+* *State mutation tree*: All accepted transactions within a slot.
+* *Activity tree*: All active accounts within a slot.
+* *State tree*: All accepted created/consumed outputs within a slot.
+* *Mana tree*: Mana of all accounts at the end of slot.
```
thibault-martinez, line 265:
```suggestion
* *Mana tree*: Mana of all accounts at the end of a slot.
```

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+
+<table>
+  <tr>
+    <th>Name</th>
+    <th>Type</th>
+    <th>Description</th>
+  </tr>
+  <tr>
+    <td>Index</td>
+    <td>int64</td>
```
thibault-martinez, line 281: signed integer?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+    <td>Root ID</td>
+    <td>ByteArray[32]</td>
+    <td>
+      A BLAKE2b-256 hash of concating multiple sparse merkle tree roots of a slot.
+      <i>Root ID = hash( Concat( hash(Concat(Tangle, TX)), hash(Concat(State, Mana)) ) )</i>  
+    </td>
+  </tr>
+  <tr>
+    <td>Cumulative Weight</td>
+    <td>int64</td>
```
thibault-martinez, line 301: signed integer?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+### Slot Commitment ID
+
+Slot Commitment ID denotes an identifier of a slot commitment, with type `ArrayBytes[40]`. It is calculated as the following steps, using BLAKE2b-256 hash function:
+
+* `content` is the serialized slot commitment.
+* `slot_index` is the slot index of the slot commitment.
+
+Caculation:
+1. `content_hash` = hash(`content`)
+2. And finally, `CommitmentID` = Concat(`content_hash`, `slot_index`)
```
thibault-martinez, line 315: For my understanding, what's the point of concatenating the slot_index here?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+  <tr>
+    <td>Previous Slot Commitment ID</td>
+    <td>ByteArray[40]</td>
+    <td>The commitment ID of the previous slot.</td>
+  </tr>
+  <tr>
+    <td>Root ID</td>
+    <td>ByteArray[32]</td>
+    <td>
+      A BLAKE2b-256 hash of concating multiple sparse merkle tree roots of a slot.
```
thibault-martinez, line 295:
```suggestion
      A BLAKE2b-256 hash of concatenating multiple sparse merkle tree roots of a slot.
```

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+        <th>Description</th>
+    </tr>
+    <tr>
+        <td>Tagged Data</td>
+        <td>5</td>
+        <td>A payload which holds a tag and associated data.</td>
+    </tr>
+    <tr>
+        <td>Transaction</td>
+        <td>6</td>
```
thibault-martinez, line 335: Do we not need a new payload type since this transaction is not compatible with the stardust one?

documentation/APIs/asyncapi3.yaml
```diff
@@ -0,0 +1,1211 @@
...
+          examples:
+            - '0xcf5f77d62285b9ed8d617729e9232ae346a328c1897f0939837198e93ec13e85'
+    subscribe:
+      tags:
+        - name: blocks
+      operationId: blockSpecificMetadata
+      description: Publishes metadata of a particular block whenever its metadata changes.
+      message:
+        $ref: '#/components/messages/BlockMetadata'
+  block-metadata/referenced:
```
luca-moser, line 140: I don't think this event makes sense because what does `referenced` mean in the context of IOTA 2.0? Previously `referenced` meant that a block was referenced by a milestone and thus "included" but it is unclear what it means here.

Perhaps consider creating events for "confirmed" and "finalised"?

luca-moser (changes requested):

documentation/APIs/core-models.md
thibault-martinez: is that `Slot Commitment ID` ?

documentation/APIs/openapi3-indexer.yaml
thibault-martinez, line 1: No Alias outputs yet?

documentation/APIs/core-models.md
jkrvivian: No, it's a different one. I'll clarify that in the context!

documentation/APIs/core-models.md
daria305: It's for data locality, we do similar thing for blockID, this way you always know where to look for the commitment/block in the database given it is organized per slot based on its ID

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+
+
+
+## Slot Index
+
+Timeline is divided into slots, and each slot has a corresponding slot index, which is a `int64`.
+To calculate the slot index of a timestamp, `genesisTimestamp` and the duration of a slot are needed.
+The slot index of timestamp `ts` is `(ts - genesisTimestamp)/duration + 1`.
+
+## Slot Commitment
```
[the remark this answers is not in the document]
jkrvivian: Slot Commitment is part of the `/block/issuance` API, clients can get it from the node, with no need to compute themselves.

documentation/APIs/openapi3-indexer.yaml
jkrvivian: I accidentally deleted it!

documentation/APIs/core-models.md
jkrvivian: It's a new type of identifier, from which we can derive the slot index directly from an ID.

documentation/APIs/core-models.md
thibault-martinez: Same as Commitment ID?

documentation/APIs/core-models.md
thibault-martinez: @jkrvivian I see that this was closed without answering or addressing it, can you let me know what's the reason for this field please?

documentation/APIs/core-models.md
jkrvivian: New version is introduced in transaction essence type.

thibault-martinez: Ah yeah we keep the transaction but bump the essence, I see. Thanks!

documentation/APIs/core-models.md
jkrvivian: I don't think so, we don't have this discussion around at the moment.

documentation/APIs/openapi3-core.yaml
```diff
@@ -0,0 +1,2920 @@
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/InternalErrorResponse'
+
+  
+  '/api/core/v3/transactions/{transactionId}/included-block':
+    get:
+      tags:
+        - UTXO
+      summary: Returns the included block of a transaction.
```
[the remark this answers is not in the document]
jkrvivian: The earliest block containing the tx that gets confirmed.
I also added the description to the file.

documentation/APIs/asyncapi3.yaml
jkrvivian: `referenced` removed, and added `confirmed` and `finalized`.

documentation/APIs/core-models.md
jkrvivian: Firstly, Block bytes are not the input of the Signature, it's `content_hash` instead, already fixed it in the doc. :sweat_smile:
 
And yes, we'll need the Issuing Time explicitly to perform validation in slot commitment chain fork.

thibault-martinez: What do you mean it's been fixed? I still see this:
> Signature must be valid, the input of a signature is concatation of Issuing Time bytes, Commitment ID bytes and Block bytes (without Signature and Nonce) by order.

thibault-martinez: Where is `Slot Commitment ID` from? I assume that's *not* the one from within the Block itself?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,576 @@
...
+* [Block](#block)
+  * [Block ID](#block-id)
+  * [Block Syntactic Validation Rules](#block-syntactic-validation-rules)
+* [Slot Index](#slot-index)
+* [Slot Commitment](#slot-commitment)
+  * [Slot Commitment ID](#slot-commitment-id)
+* [Payloads](#payloads)
+  * [Tagged Data Payload](#tagged-data-payload)
+  * [Transaction](#transaction)
+  * [Outputs and Unlock blocks](#outputs-and-unlock-blocks)
```
Thoralf-M, line 14:
```suggestion
  * [Outputs and Unlocks](#outputs-and-unlocks)
```
Should `blocks` be removed? I think we did it in Stardust everywhere except for the actual blocks

documentation/APIs/core-models.md
jkrvivian: It's the ID of Slot Commitment in a Block. I'll add this description.

thibault-martinez: Ok so it's `Previous Slot Commitment ID`

documentation/APIs/core-models.md
jkrvivian: Yeah, I removed this field, it's a fixed-size and not an optional structure, so the structure is serialized anyway.
I also double-check with the bytes from serix, we don't need the length.

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,581 @@
...
+### Block ID
+Block ID denotes an identifier of a block, with type `ArrayBytes[40]`. It is calculated as the following steps, using BLAKE2b-256 hash function:
+
+* `content` is the serialized block **without** signature and nonce.
+* `slot_index` is the slot index of the `Issuing Time` of the block.
+* `signatureBytes` is the serialized Ed25519 signature.
+
+Calculation:
+1. `content_hash` = hash(`content`)
+2. `id` = hash(Concat(`content_hash`, signatureBytes))
```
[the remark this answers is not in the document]
jkrvivian: We're having a discussion on this but might have an answer by next Tuesday. :cry:

documentation/APIs/core-models.md
thibault-martinez: Ok, thanks

documentation/APIs/core-models.md
jkrvivian: Nope, it's the ID calculated from the Slot Commitment structure of the block.

Quote from the Slot Commitment ID section:

Slot Commitment ID is calculated as the following steps, using BLAKE2b-256 hash function:

* `content` is the serialized slot commitment.
* `slot_index` is the slot index of the slot commitment (`Index` from the Slot Commitment Structure).

Caculation:
1. `content_hash` = hash(`content`)
2. And finally, `CommitmentID` = Concat(`content_hash`, `slot_index`)

jkrvivian: Let me add a Block Signature section to go through the details.

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,577 @@
...
+    <td>ByteArray[32]</td>
+    <td>The issuer identifier</td>
+  </tr>
+  <tr>
+    <td>Issuing Time</td>
+    <td>uint64</td>
+    <td>The time the block is issued. It's a Unix-like timestamp in nanosecond.</td>
+  </tr>
+  <tr>
+    <td valign="top">Slot Commitment <code>oneOf</code></td>
```
thibault-martinez, line 133: Unnecessary `oneOf`?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,588 @@
...
+1. `content_hash` = hash(`content`)
+2. `data` = Concat(`Issuing Time`, `slot_commitment_ID`, `content_hash`)
+3. And sign with the Issuer PubKey, `Signature` = Sign(`data`) 
+
+### Block ID
+Block ID denotes an identifier of a block, with type `ArrayBytes[40]`. It is calculated as the following steps, using BLAKE2b-256 hash function:
+
+* `content` is the serialized block **without** signature and nonce.
+* `slot_index` is the slot index of the `Issuing Time` of the block.
+   **Note**: It's **not** the same slot index as in the Slot Commitment.
```
thibault-martinez, line 205: So it means that we can't deduce the BlockID just by looking at the Block?

jkrvivian: We're having a discussion on this but might have an answer by next Wednesday. :cry:

documentation/APIs/openapi3-core.yaml
```diff
@@ -0,0 +1,2915 @@
...
+        - knownBlocks
+        - receivedBlocks
+        - receivedBlockRequests
+        - receivedHeartbeats
+        - sentBlocks
+        - sentBlockRequests
+        - sentHeartbeats
+        - droppedPackets
+
+    MigratedFundsEntry:
```
thibault-martinez, line 2345: Do we keep that stuff in 2.0?

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,588 @@
...
+</table>
+
+### Slot Commitment ID
+
+Slot Commitment ID denotes an identifier of a slot commitment, with type `ArrayBytes[40]`. It is calculated as the following steps, using BLAKE2b-256 hash function:
+
+* `content` is the serialized slot commitment.
+* `slot_index` is the slot index of the slot commitment.
+
+Caculation:
```
luca-moser, line 298:
```suggestion
Calculation:
```

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,588 @@
...
+### Block Signature
+This is how the signature is generated:
+
+* `content` is the serialized block **without** signature and nonce.
+* `slot_commitment_ID` is the identifier calculated from the Slot Commitment within the block, following steps described in [Slot Commitment ID](#slot-commitment-id) section.
+* `Issuing Time` is the block issuing time.
+
+Calculation:
+1. `content_hash` = hash(`content`)
+2. `data` = Concat(`Issuing Time`, `slot_commitment_ID`, `content_hash`)
```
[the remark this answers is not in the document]
jkrvivian: We want to be able to prove the time without the block itself, so the signature verification works exactly the same as the block. That way you can prove the block existed, the slot commitment is the correct one and the issuing time is valid.
And this is currently used to perform validation in slot commitment chain fork.

@thibault-martinez here's a more detailed response :stuck_out_tongue: 

I'll add it to the doc.

documentation/APIs/openapi3-core.yaml
jkrvivian: That should go away.

documentation/APIs/core-models.md
```diff
@@ -0,0 +1,590 @@
...
+3. And sign with the Issuer PubKey, `Signature` = Sign(`data`) 
+
+Including `Issuing Time` explicitly in the signature makes the signature verification works exactly the same as a block. So one can prove the block existed, the content and Slot Commitment are correct, and the Issuing Time is valid.
```
luca-moser, line 200:
```suggestion
Including `Issuing Time` explicitly in the signature makes the signature verification work exactly the same as a block's. So one can prove that the block existed, the content and Slot Commitment are correct, and the Issuing Time is valid.
```

luca-moser (changes requested):

luca-moser (approved):

thibault-martinez (approved):

documentation/APIs/openapi3-core-advanced.yaml
```diff
@@ -0,0 +1,2607 @@
...
+      required:
+        - type
+        - address
+
+    StorageDepositReturnUnlockCondition:
+      description: Can be unlocked by depositing return amount to return address via an output that only has Address Unlock Condition.
+      properties:
+        type:
+          type: integer
+          description: Set to value 1 to denote an Dust Deposit Return Unlock Condition.
```
Thoralf-M, line 1709:
```suggestion
          description: Set to value 1 to denote an Storage Deposit Return Unlock Condition.
```

thibault-martinez: a Storage *

documentation/APIs/openapi3-core-advanced.yaml
```diff
@@ -0,0 +1,2607 @@
...
+              * `8` - denotes that the return amount in a transaction is not fulfilled by the output side.
+              * `9` - denotes that the input unlock is invalid.
+              * `10` - denotes that the inputs commitment is invalid.
+              * `11` - denotes that an output contains a Sender with an ident (address) which is not unlocked.
+              * `12` - denotes that the chain state transition is invalid.
+              * `13` - denotes that ...
+              * `255` - denotes that the semantic validation failed.
+        reissuePayload:
+          type: boolean
+          description: Tells if the block should be issued again.
```
Thoralf-M, line 2418:
```suggestion
          description: Tells if the payload should be issued again.
```
?

documentation/APIs/openapi3-core-advanced.yaml
```diff
@@ -0,0 +1,2607 @@
...
+      description: Find an output by its identifier.
+      parameters:
+        - in: path
+          name: outputId
+          schema:
+            type: string
+          required: true
+          description: >-
+            Identifier of the output encoded in hex. An output is identified by
+            the concatenation of `transactionId+outputIndex`. Hex-encoded with 0x prefix.
```
Thoralf-M, line 413: I think the concatenation here is not clear `outputIndex` (u16) needs to be converted to little endian first
there are multiple places where the same text is used

jkrvivian: @Thoralf-M  Thanks for reviewing!! All the fixes go to `openapi3-core.yaml`, the advanced ones are drafts for the next version, and it's not aligned to the `openapi3-core.yaml` yet.

documentation/APIs/openapi3-core.yaml
thibault-martinez: Btw, can you confirm that all the following stardust types can be removed from the codebase completely please: `Receipt Milestone Option`,  `Treasury Transaction`, `Treasury Input`, `Treasury Output` and `Migrated Funds Entry`

jkrvivian: Yes, we don't have those in iota.go [serix-iota-core](https://github.com/iotaledger/iota.go/tree/serix-iota-core) branch.

PR #14, opened 2023-04-11
Fix go.mod and go.sum to compile without a go.work

karimodm:


.gitignore
```diff
@@ -26,6 +27,9 @@ dist/
 # Data directory 
 /data
 /peerdb
+
+go.work
+go.work.sum
```
jonastheis, line 32: Already in line 18,19

jonastheis (changes requested):
Would also be good if you can add the `get_hive` utility from GoShimmer to easily update this

karimodm:
All issues resolved.

PR #36, opened 2023-04-18
Add core API structure and goshimmer dashboard

### LINKED ISSUE #27: Implement block related APIs

### LINKED ISSUE #28: Port old dashboard

daria305:
Fixes #27 Fixes #28 
 - Initial version of dashboard and explorer, with only basic metrics enabled, like BPS, tips count, visualizer and block explorer


components/coreapi/blocks.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"github.com/labstack/echo/v4"
+	"github.com/pkg/errors"
+
+	"github.com/iotaledger/inx-app/pkg/httpserver"
+	"github.com/iotaledger/iota-core/pkg/restapi"
+	iotago "github.com/iotaledger/iota.go/v4"
+)
+
+func blockBytesByID(c echo.Context) ([]byte, error) {
```
jonastheis, line 12: there's a lot of duplicate code here. I think all 3 methods can be combined to always parse the blockIDParam and get it from the engine and afterwards extract the necessary information / format

components/dashboard/component.go
```diff
@@ -0,0 +1,213 @@
...
+
+func init() {
+	Component = &app.Component{
+		Name:      "Dashboard",
+		DepsFunc:  func(cDeps dependencies) { deps = cDeps },
+		Params:    params,
+		Configure: configure,
+		Run:       run,
+		IsEnabled: func(c *dig.Container) bool {
+			return true
```
jonastheis, line 33: I think we should have a parameter to disable it, just like in the coreapi plugin

jonastheis (changes requested):

components/dashboard/frontend/src/app/components/AliasOutputComponent.tsx.tsx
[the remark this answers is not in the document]
daria305: nah we just copy pasted this one from goshimmer, but I think it might stay we will have accounts outputs then we will just update it soon

jonastheis (approved):

PR #54, opened 2023-04-21
Feat/linter cleanup

alexsporn:


pkg/protocol/engine/booker/inmemorybooker/booker.go
```diff
@@ -166,8 +167,8 @@ func (b *Booker) trackWitnessWeight(votingBlock *blocks.Block) {
 	}
 }
 
-func (b *Booker) markInvalid(block *blocks.Block, reason error) {
-	b.setInvalidCallback(block, errors.Wrap(reason, "block marked as invalid in Booker"))
```
jonastheis, line 170: What if we need/want to attach to an invalid event? So that we get notified about invalid blocks?

jonastheis (approved):

pkg/protocol/engine/booker/inmemorybooker/booker.go
alexsporn: I added a `BlockInvalid` event to the booker

karimodm (approved):

PR #58, opened 2023-04-24
BlockIssuer with trivial reference provider

### LINKED ISSUE #56: Block issuer component

karimodm:
Closes #56


components/activity/issuer.go
```diff
@@ -1,54 +1,22 @@
...
-	pubKey := issuerKey.Public()
-
-	return iotago.AccountID(iotago.Ed25519AddressFromPubKey(pubKey[:]))
-}
-
 func issueActivityBlock() {
 	if !ParamsActivity.IgnoreBootstrapped && !deps.Protocol.MainEngineInstance().IsBootstrapped() {
 		Component.LogDebug("Not issuing activity block because node is not bootstrapped yet.")
 		return
 	}
```
jonastheis, line 11: Shouldn't this check be part of the issuer package?

pkg/protocol/blockissuer/blockfactory/events.go
```diff
@@ -0,0 +1,25 @@
+package blockfactory
+
+import (
+	"github.com/iotaledger/hive.go/runtime/event"
+	"github.com/iotaledger/iota-core/pkg/model"
+)
+
+// Events represents events happening on a block factory.
+type Events struct {
+	// Fired when a block is built including tips, sequence number and other metadata.
```
jonastheis, line 10: We don't have sequence number :P

components/protocol/component.go
```diff
@@ -58,6 +58,7 @@ func provide(c *dig.Container) error {
 		return protocol.New(
 			workerpool.NewGroup("Protocol"),
 			p2pManager,
+			p2pManager.GetLocalPeer(),
```
jonastheis, line 61: Should we better hand it in when actually issuing a block instead of storing it in the protocol? It is also not the peer that is issuing the block but rather the account.

pkg/model/parents.go
```diff
@@ -30,3 +30,6 @@ type Parent struct {
 	ID   iotago.BlockID
 	Type ParentsType
 }
+
+// ParentReferences is a map between parent type and block IDs.
+type ParentReferences map[ParentsType]iotago.BlockIDs
```
jonastheis, line 35: instead of having this type can't we just have 3 separate lists/maps for each parent type?

pkg/protocol/blockissuer/blockissuer.go
jonastheis, line 1: Does this distinction between block factory and issuer make sense? Seems just to complicate things

pkg/protocol/protocol.go
```diff
@@ -101,6 +109,21 @@ func (p *Protocol) Run() {
...
+		lo.PanicOnErr(p.localPeer.Database().LocalPrivateKey()),
+		p.API,
+		p.mainEngine.IsBootstrapped,
+		func() (*iotago.Commitment, iotago.SlotIndex) {
+			return p.mainEngine.Storage.Settings().LatestCommitment(), p.mainEngine.Storage.Settings().LatestFinalizedSlot()
+		},
+		p.ProcessBlock,
+		p.TipManager,
+		p.optsBlockIssuerOptions...,
+	)
```
jonastheis, line 123: I still think it would be better to also have providers for the protocol and its submodules

pkg/protocol/blockissuer/blockissuer.go
```diff
@@ -0,0 +1,93 @@
...
+			latestCommitmentFunc,
+			i.optsBlockFactoryOptions...)
+
+		i.Events.BlockConstructed.LinkTo(i.factory.Events.BlockConstructed)
+		i.Events.Error.LinkTo(i.factory.Events.Error)
+	})
+}
+
+// IssuePayload creates a new block including sequence number and tip selection, submits it to be processed and returns it.
+func (i *BlockIssuer) IssuePayload(p iotago.Payload, parentsCount ...int) (block *model.Block, err error) {
```
jonastheis, line 60: Here we should maybe have another function to wait for some events. If they for example all take a `blocks.Block` we could even make it so you pass the event you want to wait for.

jonastheis (changes requested):

components/activity/issuer.go
karimodm: I think the caller should assess if the protocol is bootstrapped and what wants to do with it. It would have convoluted the activity for instance, as we have to ignore this bootstrapped value in some situation, but at the same time you don't want to ignore it throughout the blockissuing.

pkg/model/parents.go
karimodm: It will bloat the number of arguments passed around block creation functions and their return values. Isn't this more compact?

pkg/blockissuer/blockissuer.go
```diff
@@ -0,0 +1,174 @@
...
+
+	select {
+	case <-ctx.Done():
+		return errors.Errorf("context canceled whilst waiting for event on block %s", block.ID())
+	case <-triggered:
+		return nil
+	}
+}
+
+// CreateBlock creates a new block including sequence number and tip selection and returns it.
```
jonastheis, line 81: no sequence numbers

pkg/blockissuer/blockissuer.go
```diff
@@ -0,0 +1,174 @@
...
+		references, err = i.getReferencesWithRetry(ctx, p, strongParentsCount)
+		if err != nil {
+			return nil, errors.Wrap(err, "error while trying to get references")
+		}
+	}
+
+	slotCommitment := i.protocol.MainEngineInstance().Storage.Settings().LatestCommitment()
+	lastFinalizedSlot := i.protocol.MainEngineInstance().Storage.Settings().LatestFinalizedSlot()
+
+	// TODO: do we have to ensure that the issuing time is at least the Max (latest) of the parents' issuing times?
```
jonastheis, line 104: Yes, otherwise the block might be invalid if the local time is wrong. However, maybe we can just error out and suggest that the local time is wrong?

pkg/blockissuer/blockissuer.go
```diff
@@ -0,0 +1,174 @@
...
+
+	for {
+		references = i.protocol.TipManager.Tips(parentsCount)
+		if len(references[model.StrongParentType]) > 0 {
+			return references, nil
+		}
+
+		select {
+		case <-interval.C:
+			i.events.Error.Trigger(errors.Wrap(err, "could not get references"))
```
jonastheis, line 154: Better log this instead of triggering an error?

jonastheis (approved):

pkg/blockissuer/blockissuer.go
jonastheis: Will be fixed with #61

jonastheis (approved):

PR #72, opened 2023-05-05
Implement APIs based on current status

### LINKED ISSUE #68: Implement core/dashboard APIs based on current status
Implement APIs based on what we have now:
* [x] `/info`
* [x] `/blocks` APIs
* [x] `/commitment` APIs
* [x] dashboard-metrics from hornet 
* [ ] UTXO retrieval
* [ ] Block issuing / creation

jkrvivian:
Add core APIs:
* `/info`
   * Add syncManager in protocol to track last accepted block etc.
   * Add metricsTracker in protocol to calculate BPS, confirmed BPS, confirmation rate.
* `/commitments`
* `/blockmetadata` , `/blocks/issuance`
   * POST block is not included yet

Add dashboard-metrics APIs:
* `extended info`
* `/database/sizes`

API document updates:
* Add `s` to commitment APIs, e.g., `/commitment/:` -> `/commitments/:`
* Update links in README

close #68


components/coreapi/types.go
```diff
@@ -76,62 +87,86 @@ type infoResponse struct {
...
 	BlocksPerSecond float64 `json:"blocksPerSecond"`
-	// The current rate of referenced blocks per second.
-	ReferencedBlocksPerSecond float64 `json:"referencedBlocksPerSecond"`
-	// The ratio of referenced blocks in relation to new blocks of the last confirmed milestone.
-	ReferencedRate float64 `json:"referencedRate"`
+	// The current rate of confirmed blocks per second.
+	ConfirmedBlocksPerSecond float64 `json:"confirmedBlocksPerSecond"`
+	// The ratio of confirmed blocks in relation to new blocks of the last confirmed milestone.
+	ConfirmedRate float64 `json:"confirmedRate"`
 }
```
cyberphysic4l, line 137: Should this be "accepted" or "committed" rather than "confirmed"? Also, need to change the wording in relation to milestones. I don't really understand what the equivalent of ConfirmedRate would be without milestones or what the purpose of this metric is.

pkg/protocol/syncmanager/trivialsyncmanager/syncmanager.go
cyberphysic4l, line 1: Could you use RLock() and RUnlock() for the mutexes in the getters here?

components/coreapi/types.go
cyberphysic4l: Perhaps the names of these metrics are actually fine, but more description is needed in the comment to understand precisely what the metric actually tells you.

components/coreapi/blocks.go
```diff
@@ -28,13 +29,18 @@ func blockMetadataResponseByID(c echo.Context) (*blockMetadataResponse, error) {
...
+		BlockID:            block.ID().ToHex(),
+		StrongParents:      block.Block().StrongParents.ToHex(),
+		WeakParents:        block.Block().WeakParents.ToHex(),
+		ShallowLikeParents: block.Block().ShallowLikeParents.ToHex(),
+		BlockState:         blockStatePending.String(),
 	}
 	_, exists := deps.Protocol.MainEngineInstance().Block(block.ID())
 	if !exists {
-		bmResponse.BlockError = "block not found"
+		bmResponse.BlockStateReason = "block not found"
```
alexsporn, line 43: where does this reason as a free-text come from? If we are using text here these need to be defined in the API so that every implementation matches

components/coreapi/component.go
```diff
@@ -202,16 +215,140 @@ func configure() error {
...
+			return c.Blob(http.StatusOK, httpserver.MIMEApplicationVendorIOTASerializerV1, output.Bytes())
+
+		default:
+			j, err := deps.Protocol.API().JSONEncode(output.Output())
+			if err != nil {
+				return err
+			}
+
+			return c.Blob(http.StatusOK, echo.MIMEApplicationJSON, j)
+		}
```
alexsporn, line 290: you can refactor this whole part (reading mimetype, switching over it, encoding an object and returning the response) into a function that takes an `any` argument so that you simply pass an output to, or a block, etc.

alexsporn: that way you don't need to copy-paste it so many times

components/coreapi/commitment.go
```diff
@@ -0,0 +1,19 @@
...
+	iotago "github.com/iotaledger/iota.go/v4"
+)
+
+func getCommitment(index iotago.SlotIndex) (*commitmentInfoResponse, error) {
+	commitment, err := deps.Protocol.MainEngineInstance().Storage.Permanent.Commitments().Load(index)
+	if err != nil {
+		return nil, err
+	}
+
+	return &commitmentInfoResponse{
```
alexsporn, line 25: use the iota.go serialization of the iotago.Commitment. just call `commitment.Commitment()` to get it

components/coreapi/component.go
```diff
@@ -202,16 +215,140 @@ func configure() error {
...
+		return httpserver.JSONResponse(c, http.StatusOK, resp)
+	}, checkNodeSynced())
+
+	routeGroup.GET(RouteSlotByIndexUTXOChanges, func(c echo.Context) error {
+		_, err := httpserver.ParseUint64Param(c, restapipkg.ParameterSlotIndex)
+		if err != nil {
+			return err
+		}
+
+		// get utxo changes
```
alexsporn: Add some `//TODO` so that we can find what is still missing

pkg/protocol/metricstracker/trivialmetricstracker/trivialmetricstracker.go
```diff
@@ -0,0 +1,124 @@
...
+	confirmedRate     float64
+	confirmedRateLock sync.RWMutex
+
+	isBootstrappedFunc isBootstrappedFunc
+
+	module.Module
+}
+
+// NewProvider creates a new MetricsTracker provider.
+func NewProvider() module.Provider[*engine.Engine, metricstracker.MetricsTracker] {
```
alexsporn, line 40: Why is this an engine module instead of an app component?

components/coreapi/commitment.go
jkrvivian: not sure if we're going to add more fields here, so we still add a struct here.

pkg/protocol/metricstracker/trivialmetricstracker/trivialmetricstracker.go
jkrvivian: You're right, this can be an app component.

components/coreapi/blocks.go
jkrvivian: We don't even need the codes from line 41, if there's any issue fetching a block, it should return an error in the upper section.

components/coreapi/blocks.go
```diff
@@ -28,26 +33,100 @@ func blockMetadataResponseByID(c echo.Context) (*blockMetadataResponse, error) {
...
 }
 
 func blockIssuance(_ echo.Context) (*blockIssuanceResponse, error) {
-	//nolint:nilnil // temporary nil,nil
-	return nil, nil
+	references := deps.Protocol.TipManager.Tips(iotago.BlockMaxParents)
+	slotCommitment := deps.Protocol.MainEngineInstance().Storage.Settings().LatestCommitment()
+
+	if len(references[model.StrongParentType]) == 0 {
+		return nil, errors.Errorf("could not get references")
```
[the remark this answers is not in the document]
alexsporn: Also if the node is not synced, since we should not allow block issuance while syncing

components/coreapi/utxo.go
```diff
@@ -0,0 +1,43 @@
...
+	spent, err := deps.Protocol.MainEngineInstance().Ledger.IsOutputSpent(outputID)
+	if err != nil {
+		return nil, err
+	}
+
+	return &outputMetadataResponse{
+		BlockID:            output.BlockID().ToHex(),
+		TransactionID:      outputID.TransactionID().ToHex(),
+		OutputIndex:        outputID.Index(),
+		IsSpent:            spent,
```
alexsporn, line 40: In HORNET we had some extra optional fields when the output was spent, so that a client could "follow" the funds

components/coreapi/blocks.go
```diff
@@ -28,26 +33,100 @@ func blockMetadataResponseByID(c echo.Context) (*blockMetadataResponse, error) {
...
+		}
+
+	default:
+		return nil, echo.ErrUnsupportedMediaType
+	}
+
+	mergedCtx, mergedCtxCancel := contextutils.MergeContexts(c.Request().Context(), Component.Daemon().ContextStopped())
+	defer mergedCtxCancel()
+
+	blockID, err := deps.BlockIssuer.AttachBlock(mergedCtx, iotaBlock)
```
alexsporn, line 112: Here we need to handle if the node allows clients to send unfinished blocks or not. Same as with the PoW this is an opt-in feature of the node operator, since it will consume the node operators Mana if the node issues the block instead of the user.

components/coreapi/utxo.go
jkrvivian: We also have a few extra fields in the OutputMetadata, I'll add a TODO here.

alexsporn (approved):

PR #74, opened 2023-05-08
Fix rootblocks and eviction index 

jonastheis:
This PR fixes the eviction index of root blocks and adds a test case to make sure eviction and eviction delay of rootblocks is done correctly

### DIFF
--- a/pkg/model/eviction_index.go
+++ b/pkg/model/eviction_index.go
@@ -1,12 +1,18 @@
 package model
 
-import iotago "github.com/iotaledger/iota.go/v4"
+import (
+	iotago "github.com/iotaledger/iota.go/v4"
+)
 
 type EvictionIndex struct {
 	index *iotago.SlotIndex
 }
 
-func (e EvictionIndex) ShouldEvict(newIndex iotago.SlotIndex) bool {
+func NewEvictionIndex() *EvictionIndex {
+	return &EvictionIndex{}
+}
+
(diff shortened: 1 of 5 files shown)


pkg/protocol/engine/eviction/state_test.go
```diff
@@ -0,0 +1,61 @@
+package eviction
```
alexsporn, line 1: `package eviction_test`

pkg/protocol/engine/eviction/testframework.go
```diff
@@ -0,0 +1,105 @@
+package eviction
```
alexsporn, line 1: `package eviction_test` and add `_test.go` suffix if possible

pkg/model/eviction_index.go
```diff
@@ -1,20 +1,26 @@
...
+import (
+	iotago "github.com/iotaledger/iota.go/v4"
+)
 
 type EvictionIndex struct {
 	index *iotago.SlotIndex
 }
 
-func (e EvictionIndex) ShouldEvict(newIndex iotago.SlotIndex) bool {
+func NewEvictionIndex() *EvictionIndex {
```
alexsporn, line 11: do we need an empty `New` here or should the user just call `new(model.EvictionIndex)` ?

jonastheis: convention for almost all of our data types are they have a constructor. also we could add an optional initial index as a parameter to it

alexsporn (approved):

PR #86, opened 2023-05-12
Inherit conflicts in booker

### LINKED ISSUE #82: Always fork TX on booking

jonastheis:
Inherit conflicts from parents based on their type/vote.

This PR implements this in a simplified manner: we assume that each TX is "treated" as a conflict in the mempool (track AW via conflictdag for all transactions). Therefore, we inherit conflictIDs/labels on the tangle for all transactions, which means that once a block is booked its conflicts can't change anymore (there's no forking logic). 

Closes #82


pkg/protocol/engine/booker/inmemorybooker/booker.go
```diff
@@ -102,20 +104,35 @@ func (b *Booker) isPayloadSolid(_ *blocks.Block) (isPayloadSolid bool, err error
...
 func (b *Booker) book(block *blocks.Block) error {
+	if err := b.trackWitnessWeight(block); err != nil {
+		return errors.Wrapf(err, "failed to track witness weight for block %s", block.ID())
+	}
+
+	conflictsToInherit, err := b.inheritConflicts(block)
+	if err != nil {
+		return errors.Wrapf(err, "failed to inherit conflicts for block %s", block.ID())
+	}
+	block.SetConflictIDs(conflictsToInherit)
```
alexsporn, line 115: you are inheriting both conflicts and payloadconflicts of your parents into your own conflicts. why the difference then? Or are the payload conflicts what your own payload generates? `SetPayloadConflicts` seems unused as of now.

jonastheis: `SetPayloadConflicts` needs to be called after the conflicts are determined by the mempool (ie after booking the tx). 

The difference is necessary because of the different references: 
- `Weak`: we only take the payload's conflicts
- `Like`: we take the payload's conflicts + remove everything that is conflicting with it

So storing it separately makes this quite easy.

pkg/protocol/engine/booker/inmemorybooker/booker.go
```diff
@@ -173,6 +192,39 @@ func (b *Booker) markInvalid(block *blocks.Block, err error) {
...
+			//  remove all conflicting conflicts from conflictIDsToInherit
+			for _, conflictID := range parentBlock.PayloadConflictIDs().Slice() {
+				if conflictingConflicts, exists := b.conflictDAG.ConflictingConflicts(conflictID); exists {
+					conflictIDsToInherit.DeleteAll(conflictingConflicts)
+				}
+			}
+		}
+	}
+
+	conflictIDsToInherit.AddAll(block.PayloadConflictIDs())
```
karimodm, line 222: Add a comment to explain that this is the "forking on booking", and set it through SetPayloadConflictIDs.

pkg/protocol/engine/therealledger/utxoledger/ledger.go
```diff
@@ -40,6 +40,9 @@ type Ledger struct {
 func NewProvider() module.Provider[*engine.Engine, therealledger.Ledger] {
 	return module.Provide(func(e *engine.Engine) therealledger.Ledger {
 		l := New(e.Workers.CreateGroup("Ledger"), e.Storage.Ledger(), e.API, e.Events.Error.Trigger)
+		// TODO: do we always book a block even if its transaction is not solid?
```
karimodm, line 43: Very good point. Although your own conflict is always known, even if the transaction is not solid. What you do not know are the parents of such conflict to properly create the conflict with the correct relationships in the ConflictDAG. It doesn't change the labels that this block carries though!
This should conern the component that it is actually creating the conflict in the ConflictDAG, as, by doing so, it has to also specify its parents.

pkg/protocol/engine/booker/inmemorybooker/booker.go
karimodm: You are inheriting only the conflicts of your parents + your own payload conflict. The former is actually, in turn, their parent's (inherited) conflicts and their own payload conflicts, in a recursive definition. The payload conflicts is what your own payload generates (you are always a conflict if you contain a TX in this case). It probably makes sense to still distinguish the two just to make clear what is the actual origin of the conflict: local (my optional TX payload), or remote (my parents), but certainly you have to set the payload conflict separately to be inherited separately with weak arrows.

pkg/protocol/engine/booker/inmemorybooker/booker.go
```diff
@@ -173,6 +192,39 @@ func (b *Booker) markInvalid(block *blocks.Block, err error) {
...
+
+	for _, parent := range block.ModelBlock().ParentsWithType() {
+		parentBlock, exists := b.blockCache.Block(parent.ID)
+		if !exists {
+			return nil, errors.Errorf("parent %s does not exist", parent.ID)
+		}
+
+		switch parent.Type {
+		case model.StrongParentType:
+			conflictIDsToInherit.AddAll(parentBlock.ConflictIDs())
```
karimodm, line 206: Here you should inherit the parent's block PayloadConflicts as well, and you have toset these conflicts separately: otherwise you cannot inherit only the payload conflicts separetely.

pkg/protocol/engine/booker/inmemorybooker/booker.go
```diff
@@ -173,6 +192,39 @@ func (b *Booker) markInvalid(block *blocks.Block, err error) {
...
+		case model.WeakParentType:
+			conflictIDsToInherit.AddAll(parentBlock.PayloadConflictIDs())
```
karimodm, line 208: You never actually set these.

pkg/protocol/engine/booker/inmemorybooker/booker.go
```diff
@@ -173,6 +192,39 @@ func (b *Booker) markInvalid(block *blocks.Block, err error) {
...
+		case model.ShallowLikeParentType:
+			// TODO: check whether it contains a TX, otherwise shallow like reference is invalid?
+
+			conflictIDsToInherit.AddAll(parentBlock.PayloadConflictIDs())
```
karimodm, line 212: Same here.

Richie0448:
Wow so Amazing

pkg/protocol/engine/ledger/ledger.go
```diff
@@ -1,9 +1,27 @@
...
+	"github.com/iotaledger/iota-core/pkg/protocol/engine/ledgerstate"
+	"github.com/iotaledger/iota-core/pkg/protocol/engine/mempool"
+	"github.com/iotaledger/iota-core/pkg/protocol/engine/mempool/conflictdag"
 	iotago "github.com/iotaledger/iota.go/v4"
 )
 
 type Ledger interface {
-	Output(id iotago.OutputID) (output State, err error)
+	AttachTransaction(block *blocks.Block) (transactionMetadata mempool.TransactionMetadata, containsTransaction bool)
+	Output(id iotago.IndexedUTXOReferencer) (*ledgerstate.Output, error)
```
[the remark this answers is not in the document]
piotrm50: `iotago.OutputID` provides a method to return `UTXOInput` that fulfills this interface. Now we don't need anything beyond what `OutputID` offers, but if we want to support additional reference types in the future (example I use when thinking about it is dust output, for which the Reference would need to contain some sort of proof as well). 

The type name is not the most intuitive, so maybe we could try to change that? Wdyt?

pkg/protocol/engine/booker/inmemorybooker/booker.go
```diff
@@ -75,14 +88,24 @@ func New(workers *workerpool.Group, sybilProtection sybilprotection.SybilProtect
 var _ booker.Booker = new(Booker)
 
 // Queue checks if payload is solid and then adds the block to a Booker's CausalOrder.
-func (b *Booker) Queue(block *blocks.Block) (wasQueued bool, err error) {
-	if isSolid, err := b.isPayloadSolid(block); !isSolid {
-		return false, errors.Wrap(err, "payload is not solid")
+func (b *Booker) Queue(block *blocks.Block) error {
+	if transactionMetadata, containsTransaction := b.ledger.AttachTransaction(block); containsTransaction {
```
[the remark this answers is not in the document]
piotrm50: Agreed

pkg/protocol/engine/booker/inmemorybooker/booker.go
```diff
@@ -75,14 +88,24 @@ func New(workers *workerpool.Group, sybilProtection sybilprotection.SybilProtect
...
+			// Based on the assumption that we always fork and the UTXO and Tangle paste cones are always fully known.
+			transactionMetadata.OnBooked(func() {
+				block.SetPayloadConflictIDs(transactionMetadata.ConflictIDs().Get())
+				b.bookingOrder.Queue(block)
+			})
+
+			return nil
+		}
+
+		return errors.Errorf("transaction in %s was not attached", block.ID())
```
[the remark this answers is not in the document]
piotrm50: When the internal call to `memPool.AttachTransaction` returns an error. The error is then passed to `errorHandler` callback and the method `return nil, true` - so it's a valid case.

karimodm (approved):

PR #98, opened 2023-05-18
Update attestations and CW to track committing to instead of committed into

### LINKED ISSUE #43: CW and Attestations "committing to"
* CW stays as part of the hash of the commitment ID, representing the CW of what is committed *into* a slot: this means that it acts automatically as a "filter" to prevent lying about CW without altering your commitmentID.
  * But how can we commit to a CW of the slot we are committing to? We won't have a perception of CW quite yet, so what should a block include? Should it be then the CW of the previous slot to serve as a lower weight bound for the slot we are committing? 
* Enforce a minimum slot age as a block filter/tipselection rule: preventing that weight committed *into* a slot does not converge with weight committed *to* a slot. This enabled CW to be treated as a lower bound given enough fork depth.
* Attestations root should probably go.
* CW should be used as a _lower_ bound but weight is proved against *committing to*.

jonastheis:
Closes #43


pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -0,0 +1,178 @@
...
+	if pendingAttestations != nil {
+		pendingAttestations.ForEach(func(key iotago.AccountID, value *iotago.Attestation) bool {
+			m.tracker.TrackAttestation(value)
+			return true
+		})
+	}
+
+	// Get all attestations for the valid time window of attestationSlotIndex up to index (as we just applied the pending attestations).
+	attestationSlotIndex := m.computeAttestationCommitmentOffset()
+	attestations := m.tracker.Attestations(attestationSlotIndex)
```
karimodm, line 128: `attestations` can be `nil` here and used in the range.

pkg/protocol/engine/attestation/slotattestation/tracker.go
```diff
@@ -0,0 +1,88 @@
...
+func NewTracker(cutoffIndexCallback func() iotago.SlotIndex) *Tracker {
+	return &Tracker{
+		latestAttestations:  shrinkingmap.New[iotago.AccountID, *iotago.Attestation](),
+		attestationsPerSlot: memstorage.NewIndexedStorage[iotago.SlotIndex, iotago.AccountID, *iotago.Attestation](),
+
+		cutoffIndexCallback: cutoffIndexCallback,
+	}
+}
+
+func (s *Tracker) TrackAttestation(attestation *iotago.Attestation) {
```
karimodm, line 26: Rename receivers to t?

pkg/protocol/engine/attestation/attestation.go
```diff
@@ -0,0 +1,20 @@
+package attestation
+
+import (
+	"github.com/iotaledger/hive.go/ads"
+	"github.com/iotaledger/hive.go/runtime/module"
+	"github.com/iotaledger/iota-core/pkg/protocol/engine/blocks"
+	iotago "github.com/iotaledger/iota.go/v4"
+)
+
+type Attestation interface {
```
karimodm, line 10: Should we rename this inferface? It is confusing with an actual iotago.Attestation

karimodm: `Attestations` (plural) maybe?

karimodm (commented):
Looks very good! I've just left some comments.
We should extend the tests a little to check for all the eviction / latest attestation logic coming from committe members and not members!

pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -0,0 +1,178 @@
...
+
+// Manager is the manager of slot attestations. It works in two phases:
+//  1. It stores "pending" attestations temporarily until the corresponding slot becomes committable.
+//  2. When a slot is committed:
+//     a. Apply the pending attestations of the committed slot.
+//     b. It computes the cumulative weight of the committed slot.
+//     c. Get all attestations for attestation slot = committed slot - attestationCommitmentOffset and store in bucketed storage.
+//     d. Compute the cumulative weight of the attestation slot.
+type Manager struct {
+	weightsProviderFunc func() *account.Accounts[iotago.AccountID, *iotago.AccountID]
```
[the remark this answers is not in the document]
jonastheis: Renamed it to `committeeFunc`.

pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -0,0 +1,178 @@
...
+	lastCumulativeWeight        uint64
+	attestationCommitmentOffset iotago.SlotIndex
+	commitmentMutex             sync.RWMutex
+
+	module.Module
+}
+
+func NewProvider() module.Provider[*engine.Engine, attestation.Attestation] {
+	return module.Provide(func(e *engine.Engine) attestation.Attestation {
+		m := NewManager(4, e.Storage.Prunable.Attestations, e.SybilProtection.Accounts)
```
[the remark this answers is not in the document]
jonastheis: Created a constant (eg can be useful as a default value) but made the parameter mandatory when creating a new provider. Similarly, I changed it in the notariazation. Now you always must specify these two parameter:
- attestationCommitmentOffset
- minCommittableAge

pkg/protocol/engine/attestation/slotattestation/tracker.go
```diff
@@ -0,0 +1,88 @@
...
+	var previousAttestationIndex iotago.SlotIndex
+	updated := true
+
+	updatedAttestation := s.latestAttestations.Compute(attestation.IssuerID, func(currentValue *iotago.Attestation, exists bool) *iotago.Attestation {
```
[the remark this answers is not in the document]
jonastheis: Since we need more information within the Tracker I can't easily see how. Also the logic is very simple and just few lines so I don't see a big problem with it

pkg/protocol/engine/attestation/slotattestation/manager.go
jonastheis: This is valid and will just result in 0 iterations. Don't see an issue here.

pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -0,0 +1,178 @@
...
+	})
+}
+
+func (m *Manager) Commit(index iotago.SlotIndex) (newCW uint64, attestationsMap *ads.Map[iotago.AccountID, iotago.Attestation, *iotago.AccountID, *iotago.Attestation], err error) {
+	m.commitmentMutex.Lock()
+	defer m.commitmentMutex.Unlock()
+
+	// Remove all pending attestations of slot and apply to tracker.
+	pendingAttestations := m.pendingAttestations.Evict(index)
+	if pendingAttestations != nil {
```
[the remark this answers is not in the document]
jonastheis: Yes if there's no attestation of a validator in the current slot. 

I don't think it changes anything. It obviously depends on the combination of parameters but an empty slot shouldn't have any bad consequences. If there's many slots without attestations it might result in no attestations for a slot that we're committing. But that is fine too, as the CW then simply doesn't grow and the tree will be empty. 

In fact, for the first slots that is even the case, as nodes first need to commit to genesis and only after `attestationCommitmentOffset` is reached we get attestations for the slot we're committing.

pkg/protocol/engine/attestation/attestation.go
```diff
@@ -0,0 +1,20 @@
...
+	Get(index iotago.SlotIndex) (attestations *ads.Map[iotago.AccountID, iotago.Attestation, *iotago.AccountID, *iotago.Attestation], err error)
+	AddAttestationFromBlock(block *blocks.Block)
+	Commit(index iotago.SlotIndex) (newCW uint64, attestationsMap *ads.Map[iotago.AccountID, iotago.Attestation, *iotago.AccountID, *iotago.Attestation], err error)
+
+	// TODO: integrate into snapshot
+	//  Import(reader io.ReadSeeker) (err error)
+	//  Export(writer io.WriteSeeker, targetSlot iotago.SlotIndex) (err error)
```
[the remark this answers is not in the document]
jonastheis: That's not correct: we can create a snapshot up to the latest commitment index. A node needs to be able to bootstrap and continue running from there. Since we consider slots from before the last commitment when creating a new commitment, we need that information. Specifically, I think we need to export (and persist on disk for correct restoring of a node after restart) the attestations that are in the tracker. So that we can compute the correct root and CW when creating a new commitment.

pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -0,0 +1,249 @@
...
+
+func (m *Manager) Shutdown() {
+	if err := m.writeToDisk(); err != nil {
+		panic(err)
+	}
+	m.TriggerStopped()
+}
+
+// Get returns the attestations that are included in the commitment of the given slot.
+// If attestationOffset=3 and commitment is 10, then the returned attestations are blocks from 7 to 10 that commit to at least 7.
```
piotrm50, line 108:
```suggestion
// If attestationCommitmentOffset=3 and commitment is 10, then the returned attestations are blocks from 7 to 10 that commit to at least 7.
```

pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -0,0 +1,249 @@
...
+
+		return currentValue
+	})
+
+	// The new attestation is smaller or equal the previousAttestation. There's no need to update pendingAttestations.
+	if !updated {
+		return
+	}
+
+	for i := cutoffIndex; i <= updatedAttestation.SlotCommitmentID.Index(); i++ {
```
piotrm50, line 183: This `for` basically makes sure that we add an Attestation not only to the committing slot, but all pending slots before that in case of a gap, right?

pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -0,0 +1,249 @@
...
+//  1. It stores "future" attestations temporarily until the corresponding slot becomes committable.
+//
+//  2. When a slot is being committed:
+//     a. Apply the future attestations of the committed slot to the pending attestations window.
+//     b. Determine attestations from window for current slot = committed slot - attestationCommitmentOffset and store in bucketed storage as committed attestations.
+//     c. Compute the new cumulative weight based on the newly committed attestations and the previous weight.
+//
+// Example with attestationCommitmentOffset = 2 and lastCommittedSlot = 12:
+//
+//	|      9     |     10     |     11     |     12     |     13     |     14     |     14     |
```
piotrm50, line 38:
```suggestion
//	|      9     |     10     |     11     |     12     |     13     |     14     |     15     |
```

pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -0,0 +1,249 @@
...
+//	|            |            |            |            |            |            |            |
```
piotrm50, line 39:
```suggestion
//	|      9     |     10     |     11     |     12     |     13     |     14     |     15     |
```

piotrm50 (approved):
Looks good and is pretty easy to understand. Just added some minor comments regarding code comments.
 The only thing question I have is about the constraints for `attestationCommitmentOffset`. In practice it needs to be at least equal to `minCommittableAge`, otherwise blocks will always attest a slot for which we aready have committed (empty) attestations. Is that correct?

pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -0,0 +1,249 @@
...
+	newAttestation := iotago.NewAttestation(block.Block())
+
+	// We keep only the latest attestation for each committee member.
+	m.futureAttestations.Get(block.ID().Index(), true).Compute(block.Block().IssuerID, func(currentValue *iotago.Attestation, exists bool) *iotago.Attestation {
+		if !exists {
+			return newAttestation
+		}
+
+		// Replace the attestation only if the new one is greater.
+		if newAttestation.Compare(currentValue) == 1 {
```
karimodm, line 148: You are only comparing the issuing time of the block here, not what they commit to. So in theory a newer block that commits to something older would be considered newer. We already had a chat about this, but I don't remember why we decided to remove the commitment from the comparison.

pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -0,0 +1,249 @@
...
+
+	updated := true
+
+	updatedAttestation := m.pendingAttestations.Get(cutoffIndex, true).Compute(attestation.IssuerID, func(currentValue *iotago.Attestation, exists bool) *iotago.Attestation {
+		if !exists {
+			return attestation
+		}
+
+		// Replace the stored attestation only if the new one is greater.
+		if attestation.Compare(currentValue) == 1 {
```
karimodm, line 169: Same here, see above.

pkg/protocol/engine/attestation/slotattestation/manager.go
karimodm: This ensures that the "latest" attestation for each issuer is "applied" to all indeces before the one it commits to. If the attestation, say, committed to 5, it should be a valid attestation for all the slots before it.
If the `Compare` above doesn't take into account the commitment of the attestation itself, but only its issuing time, wouldn't you override a possibly later slot attestation here?

karimodm: You could have the same issuer having an attestation committing to 5 on slots 0-4 and an attestation committing to 3 on slot 5, provided that this latest attestation has a newer issuingTime.

pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -0,0 +1,249 @@
...
+
+	// Store all attestations of cutoffIndex in bucketed storage via ads.Map / sparse merkle tree -> committed attestations.
+	tree, err := m.adsMapStorage(cutoffIndex)
+	if err != nil {
+		return 0, types.Identifier{}, errors.Wrapf(err, "failed to get attestation storage when committing slot %d", index)
+	}
+
+	// Add all attestations to the tree and calculate the new cumulative weight.
+	for _, a := range attestations {
+		// TODO: which weight are we using here? The current one? Or the one of the slot of the attestation/commitmentID?
```
karimodm, line 230: We need to come clear what committee we are using here, as you are requesting it "on demand": this should be the *TOTAL* committee always; in order words you should be always able to find the issuer's weight in the commitee; and since the weights are advanced only after a commitment, this should be fine. We should double-check that the sybilprotection vector is advanced only after any "weight-depending" committment activity has been performed!

pkg/protocol/engine/attestation/attestations.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/iotaledger/hive.go/ads"
+	"github.com/iotaledger/hive.go/ds/types"
+	"github.com/iotaledger/hive.go/runtime/module"
+	"github.com/iotaledger/iota-core/pkg/protocol/engine/blocks"
+	iotago "github.com/iotaledger/iota.go/v4"
+)
+
+type Attestations interface {
+	// Get returns the attestations that are included in the commitment of the given slot.
+	// If attestationOffset=3 and commitment is 10, then the returned attestations are blocks from 7 to 10 that commit to at least 7.
```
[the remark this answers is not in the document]
karimodm:
```suggestion
	// If attestationCommitmentOffset=3 and commitment is 10, then the returned attestations are blocks from 7 to 10 that commit to at least 7.
```

pkg/protocol/engine/eviction/state_test.go
```diff
@@ -17,7 +17,7 @@ func TestState_RootBlocks(t *testing.T) {
 	prunableStorage := prunable.New(database.Config{
 		Engine:    hivedb.EngineMapDB,
 		Directory: t.TempDir(),
-	}, errorHandler)
+	}, 0, errorHandler)
```
karimodm, line 20: We test somewhere for the rootCommitment right? It is kinda tricky and it is worth to make sure.

pkg/protocol/protocol.go
```diff
@@ -232,7 +231,11 @@ func (p *Protocol) initChainManager() {
 	}, event.WithWorkerPool(wp))
 
 	p.Events.Engine.SlotGadget.SlotFinalized.Hook(func(index iotago.SlotIndex) {
-		rootCommitment := p.MainEngineInstance().EarliestRootCommitment()
+		rootCommitment, valid := p.MainEngineInstance().EarliestRootCommitment(index)
+		fmt.Println("Slot finalized:", index, "root commitment:", rootCommitment, "valid:", valid)
```
karimodm, line 235: It would be nice to actually have a properly formatted log for this! All the logging is a separate topic altogether though.

jonastheis: I agree but I think we should come up with a general solution soon. I'll leave that print for now and we should replace it in a follow up PR.

pkg/protocol/engine/eviction/state_test.go
jonastheis: We do in the startup test: https://github.com/iotaledger/iota-core/blob/958554fb848c1f0b4c131186a43cd9a4281cad30/pkg/tests/protocol_startup_test.go#L451

pkg/protocol/engine/engine.go
```diff
@@ -384,14 +395,17 @@ func (e *Engine) readSnapshot(filePath string) (err error) {
 // EarliestRootCommitment is used to make sure that the chainManager knows the earliest possible
 // commitment that blocks we are solidifying will refer to. Failing to do so will prevent those blocks
 // from being processed as their chain will be deemed unsolid.
-func (e *Engine) EarliestRootCommitment() *model.Commitment {
-	earliestRootCommitmentID := e.EvictionState.EarliestRootCommitmentID()
+func (e *Engine) EarliestRootCommitment(lastFinalizedSlot iotago.SlotIndex) (earliestCommitment *model.Commitment, valid bool) {
```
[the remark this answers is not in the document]
jonastheis: Extended the comment in code by:
```
// lastFinalizedSlot is needed to make sure that the root commitment is not younger than the last finalized slot.
// If setting the root commitment based on the last evicted slot this basically means we won't be able to solidify another
// chain beyond a window based on eviction, which in turn is based on acceptance. In case of a partition, this behavior is
// clearly not desired.
```

jonastheis:
> question I have is about the constraints for attestationCommitmentOffset. In practice it needs to be at least equal to minCommittableAge, otherwise blocks will always attest a slot for which we aready have committed (empty) attestations. Is that correct?

Yes that is correct. When we inititalize the attestation provider we need to specify that value and it needs to be compatible with (meaning higher than) `minCommittableAge`.

pkg/protocol/engine/attestation/slotattestation/manager.go
jonastheis: > but all pending slots before that in case of a gap
Exactly. A "vote" on slot 5 indirectly "votes" for all slots `<= 5`.

> If the Compare above doesn't take into account the commitment of the attestation itself
The `compare` takes the commitment index into account.

PR #100, opened 2023-05-19
Implement block creation method that uses options 

piotrm50:
Implement block creation method that uses options and use it in tets-suite.

### DIFF
--- a/pkg/blockissuer/blockissuer.go
+++ b/pkg/blockissuer/blockissuer.go
@@ -56,6 +56,83 @@ func (i *BlockIssuer) Shutdown() {
 	i.workerPool.ShutdownComplete.Wait()
 }
 
+// CreateBlockWithOptions creates a new block with the options.
+func (i *BlockIssuer) CreateBlockWithOptions(ctx context.Context, opts ...options.Option[BlockParams]) (*iotago.Block, error) {
+	blockParams := options.Apply(&BlockParams{}, opts)
+
+	blockBuilder := builder.NewBlockBuilder()
+
+	var slotCommitment *iotago.Commitment
+	if blockParams.slotCommitment != nil {
+		slotCommitment = blockParams.slotCommitment
+	} else {
+		slotCommitment = i.protocol.MainEngineInstance().Storage.Settings().LatestCommitment().Commitment()
+	}
+	blockBuilder.SlotCommitment(slotCommitment)
+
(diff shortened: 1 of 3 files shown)


pkg/blockissuer/blockissuer.go
```diff
@@ -95,31 +180,19 @@ func (i *BlockIssuer) CreateBlock(ctx context.Context, p iotago.Payload, parents
...
-		StrongParents(references[model.StrongParentType]).
-		WeakParents(references[model.WeakParentType]).
-		ShallowLikeParents(references[model.ShallowLikeParentType]).
-		SlotCommitment(slotCommitment.Commitment()).
-		LatestFinalizedSlot(lastFinalizedSlot).
-		Payload(p).
-		Sign(i.Account.ID(), i.Account.PrivateKey()).
-		ProofOfWork(ctx, float64(i.protocol.MainEngineInstance().Storage.Settings().ProtocolParameters().MinPoWScore)).
-		Build()
+func (i *BlockIssuer) CreateBlockWithReferences(ctx context.Context, p iotago.Payload, references model.ParentReferences, strongParentsCountOpt ...int) (modelBlock *model.Block, err error) {
```
jonastheis, line 183: Why do we still need this function and `CreateBlock`? Can't we simply model this with `CreateBlockWithOptions` and rename it to `CreateBlock` ? Would be cleaner in my opinion

pkg/blockissuer/blockissuer.go
```diff
@@ -56,6 +56,91 @@ func (i *BlockIssuer) Shutdown() {
...
+	if blockParams.slotCommitment == nil {
+		blockParams.slotCommitment = i.protocol.MainEngineInstance().Storage.Settings().LatestCommitment().Commitment()
+	}
+
+	if blockParams.latestFinalizedSlot == nil {
+		latestFinalizedSlot := i.protocol.MainEngineInstance().Storage.Settings().LatestFinalizedSlot()
+		blockParams.latestFinalizedSlot = &latestFinalizedSlot
+	}
+
+	if blockParams.payload == nil {
```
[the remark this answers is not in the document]
piotrm50: Done

pkg/blockissuer/blockissuer.go
piotrm50: done

jonastheis (approved):

PR #106, opened 2023-05-24
Update Explorer.tsx

Phyloiota:
remove mention and link to "thetangle.org"

### DIFF
--- a/components/dashboard/frontend/src/app/components/Explorer.tsx
+++ b/components/dashboard/frontend/src/app/components/Explorer.tsx
@@ -46,8 +46,6 @@ export class Explorer extends React.Component<Props, any> {
                 </Row>
                 <ExplorerLiveFeed/>
                 <small>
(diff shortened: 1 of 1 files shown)


components/dashboard/frontend/src/app/components/Explorer.tsx
```diff
@@ -46,8 +46,6 @@ export class Explorer extends React.Component<Props, any> {
                 </Row>
                 <ExplorerLiveFeed/>
                 <small>
```
karimodm, line 48: We should also remove sorrounding tag

karimodm (changes requested):

components/dashboard/frontend/src/app/components/Explorer.tsx
Phyloiota: Will be fixed ina different PR by Max, so I'll close this one

Phyloiota:
Fixed in another PR

PR #107, opened 2023-05-24
Add Docker network snapshot with genesis output

piotrm50:
This PR fixes a minor problem when starting a node using Docker image.
It also extends the tool to create genesis snapshot that includes Genesis output with a given seed.


tools/docker-network/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "3.9"
 # Create a custom build that can be re-used within Docker network
 x-build-iota-core: &iota-core_build
   context: ../../../
-  dockerfile: ./iota-core/Dockerfile
+  dockerfile: ./iota-core/Dockerfile.workdir
```
alexsporn, line 5: Revert this

alexsporn (changes requested):
Revert the docker-compose.yml file and the rest lgtm

PR #111, opened 2023-05-25
Chain Switching

### LINKED ISSUE #104: Engine chain switching

alexsporn:
Depends on https://github.com/iotaledger/iota.go/pull/433/
Closes #104 

This PR implements Chain Switching and the evaluation i.e. fork-choice rule thereof. The high-level flow is the following:
1. A fork is detected, its chain solidified and it passes some basic tests, e.g. commitment of fork not too far in future of local clock. (this is already implemented within the ChainManager).
2. Attestations are requested one-by-one forwards starting from the forking point
3. When the attestations are received the commitment combined with the attestations are verified by the `CommitmentVerifier`
4. If the commitment, attestations and its cumulative weight are valid, ... TBD


pkg/storage/prunable/prunable.go
```diff
@@ -11,6 +11,8 @@ const (
 	blocksPrefix byte = iota
 	rootBlocksPrefix
 	attestationsPrefix
+	rootsPrefix
+	RootsKey
```
alexsporn, line 15: separate key from automatic prefix list

PR #113, opened 2023-05-26
Fix Docker image building with and without workdir

piotrm50:


tools/docker-network/run.sh
```diff
@@ -23,8 +23,15 @@ fi
...
 export COMPOSE_DOCKER_CLI_BUILD=1
 echo "Build iota-core"
+
 # Allow docker compose to build and cache an image
-docker compose -f $DOCKER_COMPOSE_FILE build --build-arg WITH_GO_WORK=${WITH_GO_WORK:-0}
+if [[ "$WITH_GO_WORK" -eq 1 ]]
+then
+  export DOCKER_BUILD_CONTEXT="../../../"
+  export DOCKERFILE_PATH="./iota-core/Dockerfile"
+fi
```
jonastheis, line 32: Would be a bit easier to understand if you declare all defaults before this if and then override. Then you just need to pass it in the call to `docker compose`

jonastheis (commented):
Dev environment doesn't work for me
```
./run_dev.sh

 => ERROR [build 6/8] COPY ./config_defaults.json /app/config.json                                                                              0.0s
 => ERROR [build 7/8] COPY ./peering.json /app/peering.json                                                                                     0.0s
```

jonastheis (approved):

PR #127, opened 2023-06-06
Confirmation States

jonastheis:
This PR introduces a bunch of changes around confirmation states of blocks and slots.

### Renaming acceptance/confirmation to pre-acceptance/confirmation, get rid of `ratified`
Since the user facing properties are the ratified versions of the confirmation state and the previous accepted/confirmed states are only used within the protocol, it makes more sense to get rid of the additional `ratified` term and simply call the ratified versions `acceptance/confirmation` and the precursor `pre-acceptance/confirmation`.

### Introduce `confirmation` similar to `acceptance`
There are now 4 confirmation types for blocks:

**Acceptance** (live, based on online committee):
- Pre-Acceptance: block `b` is pre-accepted if supermajority of the online committee approves block `b`
- Acceptance: block `b` is accepted if supermajority of the online committee has pre-accepted blocks that approve block `b`

**Confirmation** (safe, based on total committee):
- Pre-Confirmation: block `b` is pre-confirmed if supermajority of the total committee approves block `b`
- Confirmation: block `b` is confirmed if supermajority of the online committee has pre-confirmed blocks that approve block `b` within `optsConfirmationRatificationThreshold=2`. This means that confirmations on individual blocks need to happen fast within 2 slots. Otherwise blocks can become confirmed (indirectly) when a slot becomes finalized.

It is important to note that 
- `commitments` are created and therefore the chain advanced upon `acceptance`. 
- `slot finalization` happens upon `confirmation`

### Slot finalization
Slot finalization of slot `i` is triggered only if a supermajority of the total committee has confirmed blocks in slot `s` which committed to at least `c_i` (that is referenced via slot commitment within a block).

### Propagate WW and respective confirmation states on blocks across all edges on tangle level
Witnesses, ratifiers and (monotonic) acceptance/confirmation states are now propagated via all parents equally. 

### Cast transaction votes only on pre-acceptance instead on booking
This is done in order to adhere to the same confirmation states for TX as blocks do. A transaction becomes included when its attachment block becomes `accepted`. Similarly, we need to cast votes on UTXO/conflict level on block `pre-acceptance`. 

This example (implemented in `pkg/tests/confirmation_state_test.go`) shows the dependencies between the different flags: 
image


components/coreapi/types.go
```diff
@@ -113,13 +113,13 @@ type nodeStatus struct {
 	LastConfirmedBlockID string `json:"lastConfirmedBlockId"`
 	// The latest finalized slot.
 	FinalizedSlot iotago.SlotIndex `json:"latestFinalizedSlot"`
-	// The Accepted Tangle Time
+	// The PreAccepted Tangle Time
```
piotrm50, line 116: Are those really PreAccepted/PreConfirmed or is it an artifact of bulk renaming? In the clock I see that there is no PreAccepted time tracked.

components/metricstracker/component.go
```diff
@@ -69,7 +69,7 @@ func run() error {
 			deps.Protocol.Events.Engine.Notarization.SlotCommitted.Hook(func(_ *notarization.SlotCommittedDetails) {
 				deps.MetricsTracker.measure()
 			}, event.WithWorkerPool(Component.WorkerPool)).Unhook,
-			deps.Protocol.Events.Engine.BlockGadget.BlockConfirmed.Hook(func(b *blocks.Block) {
+			deps.Protocol.Events.Engine.BlockGadget.BlockPreConfirmed.Hook(func(b *blocks.Block) {
```
piotrm50, line 72: I think this should be `Confirmed` instead.

components/dashboard/visualizer.go
```diff
@@ -64,8 +64,8 @@ func runVisualizer(component *app.Component) {
 				// 	currentSlot.Store(int64(block.ID().Index()))
 				// }
 			}, event.WithWorkerPool(component.WorkerPool)).Unhook,
-			deps.Protocol.Events.Engine.BlockGadget.BlockAccepted.Hook(func(block *blocks.Block) {
-				sendVertex(block, block.IsAccepted())
+			deps.Protocol.Events.Engine.BlockGadget.BlockPreAccepted.Hook(func(block *blocks.Block) {
+				sendVertex(block, block.IsPreAccepted())
```
piotrm50, line 68: Use accepted instead? I think it would make more sense as that flag actually advances the tangle time.

pkg/protocol/engine/eviction/state.go
```diff
@@ -94,6 +95,7 @@ func (s *State) EarliestRootCommitmentID(lastFinalizedSlot iotago.SlotIndex) (ea
 			continue
 		}
 		err := storage.Stream(func(id iotago.BlockID, commitmentID iotago.CommitmentID) error {
+			fmt.Println("  ", id, commitmentID)
```
piotrm50, line 98:
```suggestion
```

piotrm50 (changes requested):
Looks good, but added some minor suggestions.

pkg/protocol/engine/eviction/state.go
```diff
@@ -215,15 +217,19 @@ func (s *State) LatestRootBlocks() iotago.BlockIDs {
 }
 
 // Export exports the root blocks to the given writer.
-func (s *State) Export(writer io.WriteSeeker, targetSlot iotago.SlotIndex) (err error) {
+func (s *State) Export(writer io.WriteSeeker, lowerTarget iotago.SlotIndex, targetSlot iotago.SlotIndex) (err error) {
```
alexsporn, line 220: `lowerTarget` means what?

pkg/protocol/engine/ledger/utxoledger/ledger.go
```diff
@@ -259,3 +261,12 @@ func (l *Ledger) BlockAccepted(block *blocks.Block) {
 		return
 	}
 }
+
+func (l *Ledger) blockPreAccepted(block *blocks.Block) {
+	votePower := booker.NewBlockVotePower(block.ID(), block.Block().IssuingTime)
```
alexsporn, line 266: it is weird that this is happening here but the booker contains the vote power

pkg/protocol/engine/eviction/state.go
jonastheis: Rootblocks are special when creating a snapshot because we not only need them to depend on the slot we're targeting to export (usually last committed slot) but also to derive the rootcommitment. The rootcommitment,  however, **must not** depend on the committed slot but on the finalized slot. Otherwise we could never switch a chain after committing (as the rootcommitment is our genesis and we don't solidify/switch chains below it).

alexsporn (approved):

piotrm50 (approved):

PR #171, opened 2023-07-05
Adapt API and ProtocolParameters to be future-proof

alexsporn:
- Use new API and ProtocolParameter interfaces from iota.go and avoid global state
- Fix engine initialization order and snapshot loading to always have a proper API available
- Various simplifications due to changes in the hive.go TypedStore and ads packages


pkg/storage/permanent/settings.go
```diff
@@ -1,264 +1,440 @@
...
 	defer s.mutex.Unlock()
 
-	s.settingsModel.LatestFinalizedSlot = index
+	return s.store.Set([]byte{latestFinalizedSlotKey}, index.MustBytes())
+}
 
-	if err = s.ToFile(); err != nil {
-		return errors.Wrap(err, "failed to persist latest confirmed slot")
+func (s *Settings) ProtocolParametersAndVersionsHash() ([32]byte, error) {
+	// TODO: concatenate and hash protocol parameters and protocol versions
```
piotrm50, line 239: Was this left out intentionally?

components/protocol/component.go
```diff
@@ -111,10 +111,10 @@ func provide(c *dig.Container) error {
 				poa.NewProvider(validators),
 			),
 			protocol.WithNotarizationProvider(
-				slotnotarization.NewProvider(iotago.SlotIndex(ParamsProtocol.Notarization.MinSlotCommittableAge)),
+				slotnotarization.NewProvider(),
 			),
 			protocol.WithAttestationProvider(
-				slotattestation.NewProvider(iotago.SlotIndex(ParamsProtocol.Notarization.MinSlotCommittableAge)+slotattestation.DefaultAttestationCommitmentOffset),
+				slotattestation.NewProvider(slotattestation.DefaultAttestationCommitmentOffset),
```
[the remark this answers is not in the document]
alexsporn: I had the same discussion with Andrea today, but we'll wait for Jonas for that, since he introduced it. I think it should be relatived to EvictionAge somehow

pkg/storage/permanent/settings.go
alexsporn: Yes. This will be done in a different PR.

piotrm50 (approved):

PR #172, opened 2023-07-05
Load PoA committee from a snapshot.

piotrm50:
Previously PoA committee was loaded from node config instead of a snapshot. Currently the PoA committee is initialized with the accounts saved in a snapshot for each epoch.


pkg/protocol/sybilprotection/sybilprotectionv1/performance/snapshot.go
```diff
@@ -349,10 +351,23 @@ func (t *Tracker) exportCommittees(pWriter *utils.PositionedWriter, _ iotago.Epo
...
+		epoch := iotago.EpochIndex(binary.LittleEndian.Uint64(epochBytes))
+		if epoch > epochFromTargetSlot {
+			// FIXME: EvictionAge*2 here should be MaxCommittableAge
+			if targetSlot+(apiForSlot.ProtocolParameters().EvictionAge()*2) < apiForSlot.TimeProvider().EpochEnd(epochFromTargetSlot) {
+				return true
+			}
+
+			committee, _, err := account.AccountsFromBytes(committeeBytes)
+			if err != nil {
+				innerErr = err // TODO: wrap the error
```
karimodm, line 363: This can be addressed

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
```diff
@@ -12,81 +12,105 @@ import (
...
-			ledger:          e.Ledger,
+			apiProvider:             e,
+			optsSeatManagerProvider: poa.NewProvider(),
 		}, opts,
-			func(o *Orchestrator) {
+			func(o *SybilProtection) {
+				o.seatManager = o.optsSeatManagerProvider(e)
+
 				e.HookConstructed(func() {
+					// TODO: the following fields should be initialized after the engine is constructed,
```
karimodm, line 55: Is it still the case?

pkg/core/account/accounts.go
```diff
@@ -162,6 +171,13 @@ func (a *Accounts) readFromReadSeeker(reader io.ReadSeeker) (n int, err error) {
 		a.setWithoutLocking(accountID, pool)
 	}
 
+	var reused bool
+	if err = binary.Read(reader, binary.LittleEndian, &reused); err != nil {
+		return n, errors.Wrap(err, "unable to read reused flag")
```
muXxer, line 176: Can you please use `ierrors` here instead?

pkg/protocol/engine/engine.go
```diff
@@ -307,6 +303,8 @@ func (e *Engine) ImportContents(reader io.ReadSeeker) (err error) {
 		return errors.Wrap(err, "failed to import commitments")
 	} else if err = e.Ledger.Import(reader); err != nil {
 		return errors.Wrap(err, "failed to import ledger")
+	} else if err := e.SybilProtection.Import(reader); err != nil {
+		return errors.Wrap(err, "failed to import epoch gadget")
```
muXxer, line 307: Can you please use `ierrors` here instead?

pkg/protocol/engine/engine.go
```diff
@@ -328,6 +326,8 @@ func (e *Engine) Export(writer io.WriteSeeker, targetSlot iotago.SlotIndex) (err
 		return errors.Wrap(err, "failed to export commitments")
 	} else if err = e.Ledger.Export(writer, targetSlot); err != nil {
 		return errors.Wrap(err, "failed to export ledger")
+	} else if err := e.SybilProtection.Export(writer, targetSlot); err != nil {
+		return errors.Wrap(err, "failed to export epoch gadget")
```
muXxer, line 330: Can you please use `ierrors` here instead?

muXxer (changes requested):

pkg/core/account/accounts.go
muXxer: Only for the added line, there rest should be replaced by my following PR anyway.

pkg/protocol/engine/engine.go
muXxer: Only for the added line, there rest should be replaced by my following PR anyway.

karimodm (approved):

PR #179, opened 2023-07-07
Improve locking and fix potential deadlocks.

piotrm50:
This PR replaces mutexes from Go's `sync` package with hive.go's `syncutil` package that allows to easily use other implementation of mutex for debugging purposes. 

It also fixes occurrences of double RLocks, fixes some occurrences of acquiring locks in different order (goroutine 1 locks A and then B, but goroutine 2 locks B and then A). Not all occurrences were possible to fix, and for those an additional explanation is provided on why it will not cause a deadlock in practice.


pkg/storage/permanent/settings.go
```diff
@@ -105,6 +109,31 @@ func (s *Settings) APIForVersion(version iotago.Version) iotago.API {
...
+	case 3:
+		a = iotago.V3API(protocolParams)
+		s.apiMutex.Lock()
+		s.apiByVersion[3] = a
+		s.apiMutex.Unlock()
+
+		return a
+	default:
+		panic(ierrors.Errorf("unsupported protocol version %d", protocolParams.Version()))
+	}
```
muXxer, line 134: Maybe this could be refactored into a function `apiFromProtocolParameters` so you don't need to duplicate the code.

muXxer:
```golang

func (s *Settings) apiForVersion(version iotago.Version) iotago.API {
	if api, exists := s.apiByVersion[version]; exists {
		return api
	}
	
	api := s.apiFromProtocolParameters(version)

	s.apiMutex.Lock()
	s.apiByVersion[version] = api
	s.apiMutex.Unlock()

	return api
}

func (s *Settings) APIForVersion(version iotago.Version) iotago.API {
	s.apiMutex.RLock()
	if api, exists := s.apiByVersion[version]; exists {
		s.apiMutex.RUnlock()
		return api
	}
	s.apiMutex.RUnlock()

	api := s.apiFromProtocolParameters(version)

	s.apiMutex.Lock()
	s.apiByVersion[version] = api
	s.apiMutex.Unlock()

	return api
}

func (s *Settings) apiFromProtocolParameters(version iotago.Version) iotago.API {
	protocolParams := s.protocolParameters(version)
	if protocolParams == nil {
		panic(fmt.Errorf("protocol parameters for version %d not found", version))
	}

	switch protocolParams.Version() {
	case 3:
		return iotago.V3API(protocolParams)
	default:
		panic(ierrors.Errorf("unsupported protocol version %d", protocolParams.Version()))
	}
}
```

muXxer:
The rest looks good to me, you can also go ahead an merge it if you want to ignore my comment.

muXxer (approved):

muXxer (approved):

PR #181, opened 2023-07-10
Core REST-API

muXxer:
This PR adapts the API to the latest Core REST-API TIP.


components/restapi/core/accounts.go
```diff
@@ -0,0 +1,203 @@
...
+		BlockIssuanceCredits: account.Credits.Value,
+	}, nil
+}
+
+func congestionForAccountID(c echo.Context) (*apimodels.CongestionResponse, error) {
+	accountID, err := httpserver.ParseAccountIDParam(c, restapipkg.ParameterAccountID)
+	if err != nil {
+		return nil, err
+	}
+	mca := deps.Protocol.LatestAPI().ProtocolParameters().EvictionAge()
```
alexsporn, line 45: This should use `CurrentAPI()` instead of `LatestAPI()`

pkg/retainer/retainer.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"github.com/iotaledger/hive.go/runtime/module"
+	"github.com/iotaledger/iota-core/pkg/model"
+	iotago "github.com/iotaledger/iota.go/v4"
+	"github.com/iotaledger/iota.go/v4/nodeclient/apimodels"
+)
+
+// Retainer keeps and resolves all the information needed in the API and INX.
+type Retainer interface {
```
alexsporn, line 11: This should not be an engine module imho.
It should be an optional node component since it is not needed for normal node operation.

components/restapi/core/component.go
```diff
@@ -85,31 +87,29 @@ const (
 	// GET returns the output IDs of all UTXO changes.
 	RouteCommitmentByIndexUTXOChanges = "/commitments/by-index/:" + restapipkg.ParameterSlotIndex + "/utxo-changes"
 
-	// RouteAccountsByAcciuntID is the route for getting an account by its accountID.
-	// GET returns the account details.
-	RouteAccountsByAcciuntID = "/accounts/:" + restapipkg.ParameterAccountID
+	// RouteBlockIssuanceCredits is the route for getting block issuance credits balance for an account.
+	// GET returns the block issuance credits balance.
+	RouteBlockIssuanceCredits = "/accounts/:" + restapipkg.ParameterAccountID
```
alexsporn, line 92: Why do we have 2 separate routes? I thought we agreed on adding both BIC and the congestion stuff to one single endpoint, since a wallet would query both anyway

components/restapi/core/component.go
```diff
@@ -85,31 +87,29 @@ const (
...
+	// RouteStakingAccount is the route for getting an account by its accountID.
+	// GET returns the account details.
+	RouteStakingAccount = "/staking/:" + restapipkg.ParameterAccountID
 
-	// RouteControlDatabasePrune is the control route to manually prune the database.
-	// POST prunes the database.
-	RouteControlDatabasePrune = "/control/database/prune"
+	// RouteRewards is the route for getting the rewards for staking of delegation based on sstaking account or delegaion output.
+	// GET returns the rewards.
+	RouteRewards = "/rewards/:" + restapipkg.ParameterAccountID
```
alexsporn, line 108: why is this an accountID and not an outputID?

pkg/retainer/retainer/retainer.go
```diff
@@ -0,0 +1,294 @@
...
+		})
+
+		r.TriggerInitialized()
+
+		return r
+	})
+}
+
+func (r *Retainer) Shutdown() {
+	r.retainerFunc = nil
```
alexsporn, line 152: this is dangerous because if any of the on* still gets called it will crash with a nil pointer exception

pkg/retainer/retainer/retainer.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+func (r *Retainer) Block(blockID iotago.BlockID) (*model.Block, error) {
```
alexsporn, line 155: Why is this different from `func (e *Engine) Block(id iotago.BlockID) (*model.Block, bool)`

alexsporn (dismissed):

pkg/retainer/retainer/retainer.go
daria305: In this form it's the same. 
Only one use case I see, could be to update it later with additional error info, like not only `not found`,  but `orphaned` or 'failed due to error...' cause we have this info in the retainer.

components/restapi/core/component.go
daria305: I added separate BIC route again, because while implementing I realized we do not provide any endpoint to query BIC for a chosen slot. 
In BlockIssuuance enpoint we provide all details related to the issuance of the block, so the BIC should be provided for `slotIndex-MCA` only, cause this makes sense in the context of issuance, you check if balance is >0. 
So simple request about out latest balance or balance for specific slot  would be too confusing, and hidden with this blockissuance endpoint bc this `-MCA`.

components/restapi/core/component.go
daria305: that's mistake, we use later outputID, will correct that

pkg/retainer/retainer.go
daria305: what about core-api responses? like for block metadata, without it we are not able to provide correct statuses. 
Also placing it inside engine seemed correct for chain switching, as data in retainer only makes sense on one chain.

components/restapi/core/accounts.go
```diff
@@ -0,0 +1,203 @@
...
+
+	return &apimodels.CongestionResponse{
+		SlotIndex:            slotIndex,
+		Ready:                false, // TODO: update after scheduler is implemented
+		ReferenceManaCost:    0,     // TODO: update after RMC is implemented
+		BlockIssuanceCredits: account.Credits.Value,
+	}, nil
+}
+
+func staking() (*apimodels.AccountStakingListResponse, error) {
```
muXxer, line 66: I thought we need an endpoint that returns all possible stakes, instead of the active ones? Or is this exactly that?
Does activeValidators here mean that they were active in the last activity window?

components/restapi/core/component.go
```diff
@@ -196,64 +196,64 @@ func configure() error {
...
 		if err != nil {
 			return err
 		}
 
 		return httpserver.JSONResponse(c, http.StatusOK, resp)
-	}, checkNodeSynced())
+	})
 
 	routeGroup.GET(RouteCommitmentByIndex, func(c echo.Context) error {
 		indexUint64, err := httpserver.ParseUint64Param(c, restapipkg.ParameterSlotIndex)
```
muXxer, line 231: We can use the ParseSlot here

components/restapi/core/component.go
```diff
@@ -196,64 +196,64 @@ func configure() error {
...
 			return err
 		}
 
-		return httpserver.JSONResponse(c, http.StatusOK, resp)
-	}, checkNodeSynced())
+		return responseByHeader(c, resp)
+	})
 
 	routeGroup.GET(RouteCommitmentByIndexUTXOChanges, func(c echo.Context) error {
 		index, err := httpserver.ParseUint64Param(c, restapipkg.ParameterSlotIndex)
```
muXxer, line 245: We can use the ParseSlot here

pkg/retainer/retainer/retainer.go
```diff
@@ -0,0 +1,299 @@
...
+	latestCommittedSlotFunc LatestCommittedSlotFunc
+	finalizedSlotFunc       FinalizedSlotFunc
+	errorHandler            func(error)
+
+	workers *workerpool.Group
+
+	module.Module
+}
+
+// the retainer should store the "confirmed" flag of blocks between they got committed and finalized.
```
muXxer, line 41: we should remove that section or rewrite it a bit cleaner :D

pkg/retainer/retainer/retainer.go
```diff
@@ -0,0 +1,299 @@
...
+	if err != nil {
+		return nil, ierrors.Wrapf(err, "failed to get block status for %s", blockID.ToHex())
+	}
+
+	hasTx, txStatus, err := r.transactionStatus(blockID, block)
+	if err != nil {
+		return nil, ierrors.Wrapf(err, "failed to get transaction status for %s", blockID.ToHex())
+	}
+
+	// TODO: fill in blockReason, TxReason.
```
muXxer, line 184: we have these reasons now?

pkg/retainer/retainer/retainer.go
```diff
@@ -0,0 +1,299 @@
...
+	return &retainer.BlockMetadata{
+		BlockStatus:       blockStatus,
+		HasTx:             hasTx,
+		TransactionStatus: txStatus,
+	}, nil
+}
+
+func (r *Retainer) RetainBlockFailure(blockID iotago.BlockID, failureCode apimodels.BlockFailureReason) {
+	retainerStore := r.retainerFunc(blockID.Index())
+	_ = retainerStore.StoreBlockFailure(blockID, failureCode)
```
muXxer, line 194: should we take care of the errors maybe?

pkg/retainer/retainer/retainer.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+func (r *Retainer) RetainTransactionFailure(transactionID iotago.TransactionID, slotIndex iotago.SlotIndex, failureCode apimodels.TransactionFailureReason) {
+	retainerStore := r.retainerFunc(slotIndex)
+	_ = retainerStore.StoreTransactionFailure(transactionID, failureCode)
```
muXxer, line 199: should we take care of the errors maybe?

pkg/retainer/retainer/retainer.go
```diff
@@ -0,0 +1,299 @@
...
+		// check if confirmed
+		if confirmed, err := r.retainerFunc(blockID.Index()).WasTransactionConfirmed(blockID); err != nil {
+			return true, apimodels.TransactionStateFailed, ierrors.Wrap(err, "failed to get transaction confirmed state in retainer.")
+		} else if confirmed {
+			return true, apimodels.TransactionStateConfirmed, nil
+		}
+	}
+
+	if pending, err := r.retainerFunc(blockID.Index()).WasTransactionPending(blockID); err != nil {
+		return true, apimodels.TransactionStateFailed, ierrors.Wrap(err, "failed to get transaction pending state in retainer.")
```
muXxer, line 262: Mhh I wonder if we should return the failed state here, because it isn't the transaction that actually failed.

pkg/retainer/retainer/retainer.go
```diff
@@ -0,0 +1,299 @@
...
+	} else if pending {
+		return true, apimodels.TransactionStatePending, nil
+	}
+
+	return true, apimodels.TransactionStateFailed, ierrors.Errorf("failed to get transaction state in retainer.")
```
muXxer, line 267: We never return the real reason for a possible transaction failure anywhere. How can we do this?

pkg/storage/prunable/retainer.go
```diff
@@ -0,0 +1,199 @@
...
+	iotago "github.com/iotaledger/iota.go/v4"
+	"github.com/iotaledger/iota.go/v4/nodeclient/apimodels"
+)
+
+const (
+	blockOrphanedPrefix byte = iota
+	blockConfirmedPrefix
+	transactionPendingPrefix
+	transactionConfirmedPrefix
+	blockFailurePrefix
```
muXxer, line 18: Maybe all block prefixes should be in order.

pkg/storage/prunable/retainer.go
```diff
@@ -0,0 +1,199 @@
...
+	transactionFailurePrefix
+)
+
+type Retainer struct {
+	slot                      iotago.SlotIndex
+	blockOrphanedStore        *kvstore.TypedStore[iotago.BlockID, types.Empty]
+	blockConfirmedStore       *kvstore.TypedStore[iotago.BlockID, types.Empty]
+	transactionPendingStore   *kvstore.TypedStore[iotago.BlockID, types.Empty]
+	transactionConfirmedStore *kvstore.TypedStore[iotago.BlockID, types.Empty]
+	blockFailureStore         *kvstore.TypedStore[iotago.BlockID, storable.SerializableInt64]
```
muXxer, line 28: Same here, first blocks, then transactions. It's a bit nicer to read.

pkg/storage/prunable/retainer.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func (r *Retainer) Store(blockID iotago.BlockID, hasTx bool) error {
+	err := r.blockOrphanedStore.Set(blockID, types.Void)
+	if err != nil {
+		return err
+	}
+
+	if hasTx {
+		err2 := r.transactionPendingStore.Set(blockID, types.Void)
```
muXxer, line 93: That looks wrong. Why err2, and then return err which is nil here?

pkg/storage/prunable/retainer.go
```diff
@@ -0,0 +1,199 @@
...
+func (r *Retainer) DeleteTransactionConfirmed(prevID iotago.BlockID) error {
+	err := r.transactionConfirmedStore.Delete(prevID)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *Retainer) StoreBlockFailure(blockID iotago.BlockID, failureType apimodels.BlockFailureReason) error {
```
muXxer, line 183: I would call it failureReason everywhere as well.

pkg/retainer/retainer.go
alexsporn: Then this is a main component of the protocol and nothing "optional" and we should treat is as such

components/restapi/core/accounts.go
```diff
@@ -0,0 +1,203 @@
...
+	}, nil
+}
+
+func congestionForAccountID(c echo.Context) (*apimodels.CongestionResponse, error) {
+	accountID, err := httpserver.ParseAccountIDParam(c, restapipkg.ParameterAccountID)
+	if err != nil {
+		return nil, err
+	}
+	mca := deps.Protocol.CurrentAPI().ProtocolParameters().EvictionAge()
+	slotIndex := deps.Protocol.CurrentAPI().TimeProvider().SlotFromTime(time.Now())
```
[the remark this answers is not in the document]
alexsporn: But this is actually wrong if the node is not synced

components/restapi/core/accounts.go
```diff
@@ -0,0 +1,203 @@
...
+		return nil, err
+	}
+	latestCommittedSlot := deps.Protocol.SyncManager.LatestCommittedSlot()
+
+	accountData, exists, err := deps.Protocol.MainEngineInstance().Ledger.Account(accountID, latestCommittedSlot)
+	if err != nil {
+		return nil, err
+	}
+	if !exists {
+		return nil, ierrors.Errorf("account not found: %s for latest committedSlot %d", accountID.ToHex(), latestCommittedSlot)
```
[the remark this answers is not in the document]
alexsporn: yes, inside the rest api any not-found should wrap the http not found error, else the automatic 404 status code won't work.

components/restapi/core/accounts.go
```diff
@@ -0,0 +1,203 @@
...
+
+func selectedCommittee(c echo.Context) *apimodels.CommitteeResponse {
+	timeProvider := deps.Protocol.APIForSlot(deps.Protocol.SyncManager.LatestCommittedSlot()).TimeProvider()
+
+	var slotIndex iotago.SlotIndex
+
+	epochIndex, err := httpserver.ParseEpochQueryParam(c, restapipkg.ParameterEpochIndex)
+	if err != nil {
+		// by default we return current epoch
+		slotIndex = timeProvider.SlotFromTime(time.Now())
```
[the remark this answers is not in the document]
alexsporn: This is a problem if the node is not synced

components/restapi/core/accounts.go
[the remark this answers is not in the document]
muXxer: yeah, I was wondering the same. We should use the latest commited slot maybe? Idk

pkg/retainer/retainer.go
muXxer: So we can call it resolved, because its not optional.

components/restapi/core/component.go
muXxer: I would say we don't allow to specify the slotIndex for the congestion endpoint at all, and we remove the BIC endpoint because it might be confusing to show a positive BIC balance but the account could still have a negative BIC balance at `latestSlot-MCA`.

pkg/retainer/retainer/retainer.go
muXxer: We will remove it.

components/restapi/core/accounts.go
```diff
@@ -0,0 +1,203 @@
...
+		BlockIssuanceCredits: account.Credits.Value,
+	}, nil
+}
+
+func congestionForAccountID(c echo.Context) (*apimodels.CongestionResponse, error) {
+	accountID, err := httpserver.ParseAccountIDParam(c, restapipkg.ParameterAccountID)
+	if err != nil {
+		return nil, err
+	}
+	mca := deps.Protocol.CurrentAPI().ProtocolParameters().EvictionAge()
```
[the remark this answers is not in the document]
muXxer: changed it to `evictionAge`.

components/restapi/core/component.go
muXxer: We removed the `RouteBlockIssuanceCredits` endpoint.

PR #182, opened 2023-07-10
Speedup Docker builds

karimodm:


muXxer (approved):
I tried it locally on my machine, it still works but I don't see that much improvement :sweat_smile:

PR #185, opened 2023-07-10
Protocol upgrades

### LINKED ISSUE #78: Protocol Upgrades signalling and detection
We should try to come up with the first example of an upgrade we have to perform and start supporting that.

Givens:
1. We have a component that will return us the outcome of the the validators signalling

Assume two different scenarios:
1. A hard fork: for instance to deprecate the existing block type in favor of another.
2. A soft fork: change of paramaters that still retain backwards compatibility, decreasing storage deposit requirement, increasing scheduler throughput.

jonastheis:
Closes #78


pkg/core/account/seated_accounts.go
```diff
@@ -5,10 +5,23 @@ import (
 	"sort"
 
 	"github.com/iotaledger/hive.go/ds/shrinkingmap"
+	"github.com/iotaledger/hive.go/ierrors"
 	iotago "github.com/iotaledger/iota.go/v4"
 )
 
-type SeatIndex int
+type SeatIndex byte
```
alexsporn, line 12: better use uint8

pkg/protocol/engine/upgrade/orchestrator.go
```diff
@@ -0,0 +1,329 @@
...
+	upgradeSignalsFunc func(slot iotago.SlotIndex) *prunable.UpgradeSignals,
+	apiProvider api.Provider,
+	setProtocolParametersEpochMappingFunc func(iotago.Version, iotago.EpochIndex) error,
+	protocolParametersAndVersionsHashFunc func(iotago.SlotIndex) (iotago.Identifier, error),
+	epochForVersionFunc func(iotago.Version) (iotago.EpochIndex, bool),
+	seatManager seatmanager.SeatManager, opts ...options.Option[Orchestrator]) *Orchestrator {
+	return options.Apply(&Orchestrator{
+		errorHandler:            errorHandler,
+		latestSignals:           memstorage.NewIndexedStorage[iotago.SlotIndex, account.SeatIndex, *prunable.SignaledBlock](),
+		permanentUpgradeSignals: kvstore.NewTypedStore[iotago.EpochIndex, VersionAndHash](permanentUpgradeSignal, iotago.EpochIndex.Bytes, iotago.EpochIndexFromBytes, VersionAndHash.Bytes, VersionAndHashFromBytes),
```
alexsporn, line 104: you can remove the generic types here since it can infer it properly from the assignment

pkg/storage/prunable/prunable.go
```diff
@@ -85,6 +86,16 @@ func (p *Prunable) PerformanceFactors(slot iotago.SlotIndex) *PerformanceFactors
 	return NewPerformanceFactors(slot, store)
 }
 
+func (p *Prunable) UpgradeSignals(slot iotago.SlotIndex) *UpgradeSignals {
+	// TODO: make sure that the minimum pruning delay for this is at least 1 epoch, otherwise we won't be able to calculate the reward pools
```
alexsporn, line 90: wrong todo comment

pkg/protocol/engine/upgrade/orchestrator.go
```diff
@@ -0,0 +1,329 @@
...
+	protocolParametersAndVersionsHashFunc func(iotago.SlotIndex) (iotago.Identifier, error)
+	epochForVersionFunc                   func(iotago.Version) (iotago.EpochIndex, bool)
+
+	apiProvider api.Provider
+	seatManager seatmanager.SeatManager
+
+	module.Module
+}
+
+func NewProvider(opts ...options.Option[Orchestrator]) module.Provider[*engine.Engine, *Orchestrator] {
```
alexsporn, line 72: I am missing the way to provide the new set of version+protocolparams from the config

alexsporn (approved):

PR #187, opened 2023-07-12
Adapt to changes of version type in iota.go and make version deserialization safe

jonastheis:
Depends on https://github.com/iotaledger/iota.go/pull/453. After it is merged `go.mod` should be updated to latest commit.

### DIFF
--- a/pkg/protocol/commitment_verifier.go
+++ b/pkg/protocol/commitment_verifier.go
@@ -5,6 +5,7 @@ import (
 	"github.com/iotaledger/hive.go/ds/set"
 	"github.com/iotaledger/hive.go/ierrors"
 	"github.com/iotaledger/hive.go/kvstore/mapdb"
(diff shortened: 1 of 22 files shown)


pkg/protocol/commitment_verifier.go
```diff
@@ -27,11 +28,16 @@ func (c *CommitmentVerifier) verifyCommitment(prevCommitment, commitment *model.
 		iotago.Identifier.Bytes,
 		iotago.IdentifierFromBytes,
 		func(attestation *iotago.Attestation) ([]byte, error) {
-			return c.engine.APIForVersion(attestation.ProtocolVersion).Encode(attestation)
+			return lo.PanicOnErr(c.engine.APIForVersion(attestation.ProtocolVersion)).Encode(attestation)
```
alexsporn, line 31: maybe return error here

alexsporn (approved):

PR #192, opened 2023-07-17
Evil spammer & explorer working & deadlock fixes

karimodm:


karimodm:
Not going to fix all linting warnings on evil tooling

components/protocol/component.go
```diff
@@ -123,71 +117,75 @@ func provide(c *dig.Container) error {
...
-		Component.LogErrorf("Error in Protocol: %s", err)
-	})
+	// deps.Protocol.Events.Error.Hook(func(err error) {
+	// 	Component.LogErrorf("Error in Protocol: %s", err)
+	// })
 
-	deps.Protocol.Events.Network.Error.Hook(func(err error, id network.PeerID) {
-		Component.LogErrorf("NetworkError: %s Source: %s", err.Error(), id)
-	})
+	// deps.Protocol.Events.Network.Error.Hook(func(err error, id network.PeerID) {
```
alexsporn, line 124: revert these changes

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/iotaledger/iota-core
 go 1.20
 
 require (
+	github.com/AlecAivazis/survey/v2 v2.3.7
```
alexsporn, line 6: make a separate go.mod for the evil-wallet and evil-spammer, please do not pollute the main go.mod

pkg/protocol/protocol.go
```diff
@@ -330,6 +332,12 @@ func (p *Protocol) APIForEpoch(epoch iotago.EpochIndex) iotago.API {
 	return p.MainEngineInstance().APIForEpoch(epoch)
 }
 
+func (p *Protocol) HookLogging() {
```
alexsporn, line 335: remove this

pkg/protocol/engine/ledger/ledger.go
```diff
@@ -28,6 +28,7 @@ type Ledger interface {
 	AddUnspentOutput(unspentOutput *utxoledger.Output) error
 
 	ConflictDAG() conflictdag.ConflictDAG[iotago.TransactionID, iotago.OutputID, BlockVoteRank]
+	MemPool() mempool.MemPool[BlockVoteRank]
```
alexsporn, line 31: why is this exposed now?

pkg/protocol/engine/ledger/ledger/ledger.go
```diff
@@ -225,7 +228,24 @@ func (l *Ledger) Output(outputID iotago.OutputID) (*utxoledger.Output, error) {
 	case *ExecutionOutput:
 		txWithMetadata, exists := l.memPool.TransactionMetadata(outputID.TransactionID())
 		if !exists {
-			return nil, err
+			var output *utxoledger.Output
+			stateRequest := l.resolveState(outputID.UTXOInput())
```
alexsporn, line 232: why are you resolving here? this should come from the mempool. If it is for the API, look at the other function I added `OutputOrSpent`

alexsporn (changes requested):
Look at the comments above

pkg/protocol/engine/ledger/ledger.go
karimodm: We need to obtain the MemPool from the dashboard and tipselection to be able to load transaction metadata.

pkg/protocol/engine/ledger/ledger/ledger.go
karimodm: If the mempool happens to still hold this "intermediate" representation of an executed output you might run into a race condition where the transaction metadata is not available anymore in the mempool to create an utxoledger.Output. If you run into this race then you try to resolve the output against the storage, and at some point it must be there, as the race resolves: hence the promise.

alexsporn (approved):

PR #194, opened 2023-07-18
Added initial INX support

alexsporn:
Requires
- https://github.com/iotaledger/iota.go/pull/461


pkg/storage/permanent/settings.go
```diff
@@ -186,21 +213,36 @@ func (s *Settings) LatestCommitment() *model.Commitment {
 
 	if err != nil {
 		if ierrors.Is(err, kvstore.ErrKeyNotFound) {
-			return model.NewEmptyCommitment(s.LatestAPI())
+			return model.NewEmptyCommitment(s.apiProvider.LatestAPI())
```
jonastheis, line 216: We need to use `CurrentAPI()` everywhere

pkg/storage/permanent/settings.go
jonastheis, line 1: I think using typed KV stores with the prefixed realm would make this a bit easier to read and less verbose

jonastheis (changes requested):

pkg/storage/permanent/settings.go
alexsporn: Yes, now this can work, before we would have a cyclic dependency between `CurrentAPI` and `LatestCommitment`

pkg/storage/permanent/settings.go
alexsporn: I'll do a separate PR with some cleanups

jonastheis (approved):

PR #197, opened 2023-07-20
Ansible code to deploy wire-guarded Feature network

karimodm:


components/protocol/params.go
```diff
@@ -27,8 +27,8 @@ type ParametersDatabase struct {
 	Engine           string `default:"rocksdb" usage:"the used database engine (rocksdb/mapdb)"`
 	Path             string `default:"testnet/database" usage:"the path to the database folder"`
 	MaxOpenDBs       int    `default:"10" usage:"maximum number of open database instances"`
-	PruningThreshold uint64 `default:"360" usage:"how many confirmed slots should be retained"`
-	DBGranularity    int64  `default:"1" usage:"how many slots should be contained in a single DB instance"`
+	PruningThreshold uint64 `default:"16384" usage:"how many confirmed slots should be retained"`
```
muXxer, line 30: I would keep at least 30 days tbh, if not even more.
With hornet we have 30GB as a default DB size now if we prune by size.

But for our tests now I guess 2 days is ok.

config_defaults.json
```diff
@@ -105,8 +105,8 @@
     "bindAddress": "0.0.0.0:8081",
     "basicAuth": {
       "enabled": false,
-      "username": "core",
-      "password": "core"
+      "username": "iota-core",
```
muXxer, line 108: you have to change these values in the `params.go` of the dashboard component, and then run `tools/gendoc` instead of modifying the file manually.

muXxer (approved):
Feel free to address my comments, otherwise everything else looks fine so far.

components/protocol/params.go
karimodm: I would like this to be the "minimum" for a reasonable network: hence the choice; I didn't run any number, I've set this purely by components' minimal need.

PR #200, opened 2023-07-21
Persist newest slot with rootblocks to prevent costly IO

karimodm:


pkg/storage/permanent/permanent.go
```diff
@@ -100,6 +102,14 @@ func (p *Permanent) Accounts(optRealm ...byte) kvstore.KVStore {
 	return lo.PanicOnErr(p.accounts.WithExtendedRealm(optRealm))
 }
 
+func (p *Permanent) LatestNonEmptySlot(optRealm ...byte) kvstore.KVStore {
+	if len(optRealm) == 0 {
+		return p.accounts
```
muXxer, line 107: copy paste error I guess?
```suggestion
		return p.latestNonEmptySlot
```

pkg/protocol/engine/eviction/state.go
```diff
@@ -217,13 +226,15 @@ func (s *State) LatestRootBlocks() iotago.BlockIDs {
 // The lowerTarget is usually going to be the last finalized slot because Rootblocks are special when creating a snapshot.
 // They not only are needed as a Tangle root on the slot we're targeting to export (usually last committed slot) but also to derive the rootcommitment.
 // The rootcommitment, however, must not depend on the committed slot but on the finalized slot. Otherwise, we could never switch a chain after committing (as the rootcommitment is our genesis and we don't solidify/switch chains below it).
-func (s *State) Export(writer io.WriteSeeker, lowerTarget iotago.SlotIndex, targetSlot iotago.SlotIndex) (err error) {
+func (s *State) Export(writer io.WriteSeeker, lowerTarget, targetSlot iotago.SlotIndex) (err error) {
```
muXxer, line 229: I don't like that particular change :P

muXxer (changes requested):

pkg/storage/permanent/permanent.go
karimodm: Yep!

PR #208, opened 2023-07-25
Validate attestation issuer public key to match accountID

### LINKED ISSUE #144: Verify that Attestations signature's pubkey matches accountID signed
...and support key rotations across epochs and committee rotations.

jonastheis:
This PR introduces validation rules so that the public key used to sign an attestation/block matches the accountID (whether it is registered as an issuer key for the given account). There are some complications when evaluating this without having all information of the chain (we only have the ledger state at the forking point): 

Specifically, following has to be considered carefully
```
Verify that cumulative weight of commitment matches (as an upper bound) with calculated weight from attestations.
This is necessary due to the following edge case that can happen in the window of forking point and the current state of the other chain:
		1. A public key is added to an account.
	    We do not count a seat for the issuer for this slot and the computed CW will be lower than the CW in
		   the commitment. This is fine, since this is a rare occasion and a heavier chain will become heavier anyway, eventually.
		   It will simply take a bit longer to accumulate enough CW so that the chain-switch rule kicks in.
	
	2. A public key is removed from an account.
	   We count the seat for the issuer for this slot even though we shouldn't have. According to the protocol, a valid
	   chain with such a block can never exist because the block itself (here provided as an attestation) would be invalid.
	   However, we do not know about this yet since we do not have the latest ledger state.
	   In a rare and elaborate scenario, an attacker might have acquired such removed (private) keys and could
	   fabricate attestations and thus a theoretically heavier chain (solely when looking on the chain backed by attestations)
	   than it actually is. Nodes might consider to switch to this chain, even though it is invalid which will be discovered
	   before the candidate chain/engine is activated (it will never get heavier than the current chain).
```

Fixes #144


pkg/protocol/commitment_verifier.go
```diff
@@ -55,21 +65,37 @@ func (c *CommitmentVerifier) verifyCommitment(prevCommitment, commitment *model.
...
-	// 3. Verify cumulative weight of commitment matches with calculated weight from attestations.
-	if prevCommitment.CumulativeWeight()+seatCount != commitment.CumulativeWeight() {
-		return nil, ierrors.Errorf("invalid cumulative weight for commitment %s", commitment.ID())
+	// 3. Verify that cumulative weight of commitment matches (as an upper bound) with calculated weight from attestations.
+	// This is necessary due to the following edge case that can happen in the window of forking point and the current state of the other chain:
+	//	1. A public key is added to an account.
+	//     We do not count a seat for the issuer for this slot and the computed CW will be lower than the CW in
+	//	   the commitment. This is fine, since this is a rare occasion and a heavier chain will become heavier anyway, eventually.
+	//	   It will simply take a bit longer to accumulate enough CW so that the chain-switch rule kicks in.
+	//	TODO: what happens in an extreme case where all change their keys?
```
karimodm, line 83: I don't believe this situation can be solved: how can you verify data on-chain if you only trust the forking point, in which, all the keys are now discontinued?
You cannot verify the chain up-front: your only chance is to actually sync the chain, regardless of its CW and, once synced, check its candidate CW against your main CW.

pkg/protocol/engine/accounts/accountsledger/manager.go
```diff
@@ -185,6 +185,36 @@ func (m *Manager) Account(accountID iotago.AccountID, targetIndex iotago.SlotInd
 	return loadedAccount, true, nil
 }
 
+// PastAccounts loads the past accounts' data at a specific slot index.
+func (m *Manager) PastAccounts(accountIDs iotago.AccountIDs, targetIndex iotago.SlotIndex) map[iotago.AccountID]*accounts.AccountData {
```
karimodm, line 189: It is a bit ugly to expose this even if it is only used in this context. Could we relax the check on the `Account` method instead? It is nice that we enforce a strict limit there, but why do we do it anyway?

pkg/protocol/protocol.go
```diff
@@ -289,6 +289,9 @@ func (p *Protocol) ProcessBlock(block *model.Block, src network.PeerID) error {
 			candidateEngine.ProcessBlockFromPeer(block, src)
 			if candidateEngine.IsBootstrapped() &&
 				candidateEngine.Storage.Settings().LatestCommitment().CumulativeWeight() > mainEngine.Storage.Settings().LatestCommitment().CumulativeWeight() {
+				// TODO: when activating the candidate engine should we make sure that the candidate engine produces the same commitments as the chain we decided to switch to?
```
karimodm, line 292: Seems like a reasonable sanity check, although I am not sure if it is really necessary, as we only spawn a candidate engine when we proved that a heavier chain exists, and once a heavier chain is synced (even though not the same one) you should anyway prefer this candidate engine over your main as its CW is higher!

pkg/protocol/commitment_verifier.go
```diff
@@ -78,9 +104,27 @@ func (c *CommitmentVerifier) verifyAttestations(attestations []*iotago.Attestati
...
-		//  First, this can be based on the latest commonly known ledger state.
-		//  Later, this needs to include a proof of added/removed public keys.
+		// 1. Make sure the public key used to sign is valid for the given issuerID.
+		//    We ignore attestations that don't have a valid public key for the given issuerID.
+		//    1. The attestation might be fake.
+		//    2. The issuer might have added a new public key in the meantime, but we don't know about it yet
+		//       since we only have the ledger state at the forking point.
+		accountData, exists := c.validatorAccountsAtFork[att.IssuerID]
+
+		// We always need to have the accountData for a validator.
```
karimodm, line 114: ... unless the attestation provided belongs to a different committee

pkg/protocol/engine/accounts/accountsledger/manager.go
jonastheis: I don't know why this check is in the `Account` method in first place. Also the method signature seems to be a bit strange to me with `account, bool, err` as the return type.

pkg/protocol/commitment_verifier.go
jonastheis: I agree, we should have a test of engine switching over epoch boundaries where a committee changes. However, this shouldn't be part of this PR. I've created #210 to keep track of this

pkg/protocol/engine/accounts/accountsledger/manager.go
jonastheis: After talking to @cyberphysic4l I think the separate method (for explicitly getting `PastAccounts`) should stay as the `Account` method has a safeguard to accidentally triggering too many disk accesses in `rollbackAccountTo`. 

> the intended use of the account manager is to load the account info when checking an arriving block. The arriving block can't be older than `latestCommittedSlot`  which means it's commitment can't be older than `latestCommittedSlot-MaxCommittableAge`

pkg/protocol/commitment_verifier.go
jonastheis: I added a note that the only solution to prevent this is to add changes (provable) to the commitments

pkg/protocol/protocol.go
jonastheis: Addressed this in `onForkDetected`

jonastheis: Addressed this in `onForkDetected`

pkg/protocol/commitment_verifier.go
```diff
@@ -7,21 +7,31 @@ import (
...
+		engine:                  mainEngine,
+		forkingPoint:            forkingPoint,
+		cumulativeWeight:        forkingPoint.CumulativeWeight(),
+		validatorAccountsAtFork: mainEngine.Ledger.PastAccounts(committeeAtForkingPoint, forkingPoint.Index()),
+		// TODO: what happens if the committee rotated after the fork?
 	}
 }
 
-func (c *CommitmentVerifier) verifyCommitment(prevCommitment, commitment *model.Commitment, attestations []*iotago.Attestation, merkleProof *merklehasher.Proof[iotago.Identifier]) (iotago.BlockIDs, error) {
+func (c *CommitmentVerifier) verifyCommitment(commitment *model.Commitment, attestations []*iotago.Attestation, merkleProof *merklehasher.Proof[iotago.Identifier]) (iotago.BlockIDs, uint64, error) {
```
alexsporn, line 34: do you want to give the return parameters some names for readability? (but do not use those vars ;-) )

pkg/protocol/protocol.go
```diff
@@ -189,7 +189,8 @@ func (p *Protocol) shutdown() {
 	p.activeEngineMutex.RLock()
 	p.mainEngine.Shutdown()
 	if p.candidateEngine != nil {
-		p.candidateEngine.Shutdown()
+		// TODO: do we want to delete the candidate engine's data as well?
```
alexsporn, line 192: remove TODO, this happens automatically on node restart (see enginemanager `LoadActiveEngine`)

pkg/protocol/protocol_fork.go
```diff
@@ -82,57 +83,96 @@ func (p *Protocol) onForkDetected(fork *chainmanager.Fork) {
...
+	// Clean up events when we switch to the candidate engine.
+	detachMainEngineSwitched = p.Events.MainEngineSwitched.Hook(func(_ *engine.Engine) {
 		detachRequestBlocks()
 		detachProcessCommitment()
-	}, event.WithMaxTriggerCount(1))
+	}, event.WithMaxTriggerCount(1)).Unhook
+
+	// Clean up candidate engine if we never switch to it.
+	go func() {
+		timer := time.NewTimer(10 * time.Minute)
```
alexsporn, line 140: I never see this timer getting cancelled if we switch to the candidate engine properly

pkg/protocol/commitment_verifier.go
```diff
@@ -91,8 +91,8 @@ func (c *CommitmentVerifier) verifyCommitment(commitment *model.Commitment, atte
 	//    than it actually is. Nodes might consider to switch to this chain, even though it is invalid which will be discovered
 	//    before the candidate chain/engine is activated (it will never get heavier than the current chain).
 	c.cumulativeWeight += seatCount
-	if c.cumulativeWeight <= commitment.CumulativeWeight() {
-		return nil, 0, ierrors.Errorf("invalid cumulative weight for commitment %s", commitment.ID())
+	if c.cumulativeWeight > commitment.CumulativeWeight() {
```
karimodm, line 96: The comment above is not true for this comparison anymore
> 	// 3. Verify that cumulative weight of commitment matches (as an upper bound) with calculated weight from attestations.
I would explain why you check the cumulative weight of the commitment as a lower bound now.

pkg/protocol/protocol_fork.go
```diff
@@ -82,50 +83,88 @@ func (p *Protocol) onForkDetected(fork *chainmanager.Fork) {
...
+		defer timeutil.CleanupTimer(timer)
+
+		select {
+		case <-timer.C:
+			p.ErrorHandler()(ierrors.Errorf("timeout waiting for candidate engine %s to sync", candidateEngineInstance.Name()))
+		case <-p.context.Done():
+			// TODO: what exactly do we do with a candidate engine when shutting down a node?
```
karimodm: It should get cleaned up on the next start.

pkg/protocol/protocol.go
```diff
@@ -284,14 +285,11 @@ func (p *Protocol) ProcessBlock(block *model.Block, src network.PeerID) error {
...
-			if candidateEngine.IsBootstrapped() &&
-				candidateEngine.Storage.Settings().LatestCommitment().CumulativeWeight() > mainEngine.Storage.Settings().LatestCommitment().CumulativeWeight() {
-				// TODO: when activating the candidate engine should we make sure that the candidate engine produces the same commitments as the chain we decided to switch to?
-				//   -> if they are different we could shut down this engine.
-
+	if candidateEngineInstance := p.CandidateEngineInstance(); candidateEngineInstance != nil {
+		if candidateChain := candidateEngineInstance.ChainID(); chainCommitment.Chain().ForkingPoint.ID() == candidateChain || candidateEngineInstance.BlockRequester.HasTicker(block.ID()) {
+			candidateEngineInstance.ProcessBlockFromPeer(block, src)
+			if candidateEngineInstance.IsBootstrapped() &&
+				candidateEngineInstance.Storage.Settings().LatestCommitment().CumulativeWeight() > mainEngine.Storage.Settings().LatestCommitment().CumulativeWeight() {
```
karimodm, line 291: How do you enforce here that the candidateEngine is actually synced on the chain you meant to sync in the first place?
Why shouldn't I be able to prove to you that a heavier chain exists, and then make your candidateEngine sync on a very shallow fork (1 commitment deep) that is heavier than your main engine? You will switch here!

PR #209, opened 2023-07-25
Adjust block issuer key type.

piotrm50:
This PR adjusts block issuer key type to be in line with recent iota.go changes in https://github.com/iotaledger/iota.go/pull/462


muXxer (approved):

muXxer:
Man, that was hard to fix the go modules :D

PR #213, opened 2023-07-26
Fix bugs in DRR scheduler. 

piotrm50:


pkg/protocol/engine/congestioncontrol/scheduler/drr/issuerqueue.go
```diff
@@ -123,6 +123,10 @@ func (q *IssuerQueue) Front() *blocks.Block {
 
 // PopFront removes the first ready block from the queue.
 func (q *IssuerQueue) PopFront() *blocks.Block {
+	if q.inbox.Len() == 0 {
```
muXxer, line 126: Isn't that check a bit racy with the `heap.Pop` ?

pkg/blockfactory/blockissuer.go
```diff
@@ -43,14 +43,19 @@ type BlockIssuer struct {
...
 	return options.Apply(&BlockIssuer{
-		events:     NewEvents(),
-		workerPool: p.Workers.CreatePool("BlockIssuer"),
-		protocol:   p,
-		Account:    account,
+		events:                      NewEvents(),
+		workerPool:                  p.Workers.CreatePool("BlockIssuer"),
+		protocol:                    p,
+		Account:                     account,
+		optsIncompleteBlockAccepted: false,
```
cyberphysic4l, line 55: What is `optsIncompleteBlockAccepted` supposed to indicate?

pkg/protocol/engine/congestioncontrol/scheduler/drr/issuerqueue.go
```diff
@@ -135,8 +139,27 @@ func (q *IssuerQueue) PopFront() *blocks.Block {
 }
 
 func (q *IssuerQueue) RemoveTail() *blocks.Block {
+	var oldestSubmittedBlock *blocks.Block
```
muXxer, line 142: I didn't look in detail how the IssuerQueue works... but shouldn't we introduce some mutexes before we loop pver submitted and pop the tail etc? Or is the logic safe against such data races?

pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
```diff
@@ -337,15 +328,66 @@ func (s *Scheduler) selectIssuer(start *IssuerQueue) (int64, *IssuerQueue) {
...
+func (s *Scheduler) removeIssuer(q *IssuerQueue) {
+	q.submitted.ForEach(func(id iotago.BlockID, block *blocks.Block) bool {
+		block.SetDropped()
+		s.events.BlockDropped.Trigger(block)
+
+		return true
+	})
+
+	for q.inbox.Len() > 0 {
+		block := q.PopFront()
```
muXxer, line 340: Could also be racy, what if the inbox had an element during the length check, but not during PopFront?

pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
```diff
@@ -337,15 +328,66 @@ func (s *Scheduler) selectIssuer(start *IssuerQueue) (int64, *IssuerQueue) {
...
+			if err != nil {
+				err = ierrors.Wrapf(err, "could not get deficit for issuer %s", accountID)
+				return 0
+			}
+
+			// load with max deficit if the issuer has Mana but has been removed from the deficits map
+			return s.optsMaxDeficit
+		}
+
+		if int64(currentValue)+delta < 0 {
```
muXxer, line 378: Shouldn't we use the safemath functions?

cyberphysic4l (approved):

pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
piotrm50: Yeah! But I think applying safemath belongs to a separate PR that changes this in all the places across codebase. We're not using that anywhere yet, right?

muXxer: Sure, can you add a TODO there so we don't miss it?

pkg/protocol/engine/congestioncontrol/scheduler/drr/issuerqueue.go
piotrm50: Access to the buffer (`IssuerQueue`) is protected by `s.bufferMutex` in the scheduler - that guarantees that only one goroutine can call scheduler methods that work with the buffer at a time. The same applies to your other comments.

pkg/blockfactory/blockissuer.go
piotrm50: It defines whether the node allows to fill in incomplete block and issue it for user.
 
 It's one of the node parameters. I'll add a comment in the block issuer as well to be clear.

pkg/protocol/engine/congestioncontrol/scheduler/drr/issuerqueue.go
muXxer: Protected from outside by bufferMutex

pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
muXxer: Protected from outside by bufferMutex.

muXxer (approved):

PR #173, opened 2023-07-05
Congestion Control: Accounts Filter only

cyberphysic4l:
This adds a new module - the accounts filter.

The accounts filter takes care of filter checks which rely on account information from the account manager which in turns relies on commitments. The checks include BIC, signature verification (including that the key used to sign the block is actually registered to the account), burned Mana check (which relies on RMC and hence commitments).

The accounts filter also deals with "future blocks" whose commitments are not yet available to a node. Future blocks were dealt with in the BlockDAG, but they are no longer required there as the accounts filter sits before the BlockDAG.

PR #221, opened 2023-07-31
Min and max committable age adjustments 

jonastheis:
This PR makes a bunch of related changes around :
- get rid of the constant `slotattestation.DefaultAttestationCommitmentOffset=4` by utilizing `MinCommittableAge` and `MaxCommittableAge` as the window bounds for attestations.
- introduce a better way to work with slot-based tests (ie tests that fabricate/simulate slot commitments and epochs) to express them easier as well as make them more stable. Specifically, following tests are reworked
  - upgrade signaling test
  - engine switching test
  - startup / snapshot loading test
- add unsolid commitment blocks at protocol level and get rid of future blocks at engine level
- simplify `EarliestRootCommitment` by utilizing `MaxCommittableAge`
- add `getCommitment` to blockissuer to make sure only valid blocks according to min/max committable age are produced
- add an event `LatestCommitmentUpdated` to get rid of the race condition of `SlotCommitted` and reading `LastCommittedSlot`
- fix some other race condition and bugs regarding engine/chain switching, shutdown, snapshot and node startup
- properly handle blocks that are already stored to disk on node restart (we shortcut missing block requests to check whether a block in a certain slot range is available on disk)
- Add `SplitIntoPartitions` to testsuite to create a partition in a running network and be able to merge it back again 
- Change dataflow: blocks with unsolid commitment are buffered in the protocol. all other blocks are forwarded to their respective engine (even if the engine itself has not yet that far committed) to enable solidification requests from the BlockDAG. Otherwise, it might happen, that we got stuck because the blocks that would lead to a new commitment could not be requested. This change also requires that the commitmentfilter is now after the BlockSolid event
- Fix bug and simplify import/export of attestations

Fixes: #216 

Requires: https://github.com/iotaledger/hive.go/pull/539, https://github.com/iotaledger/iota-core/pull/173, https://github.com/iotaledger/iota.go/pull/444


pkg/protocol/engine/engine.go
```diff
@@ -188,15 +193,18 @@ func New(
...
 				e.EvictionState.PopulateFromStorage(e.Storage.Settings().LatestCommitment().Index())
+
 				if err := e.Attestations.RestoreFromDisk(); err != nil {
 					panic(ierrors.Wrap(err, "failed to restore attestations from disk"))
 				}
 				if err := e.UpgradeOrchestrator.RestoreFromDisk(e.Storage.Settings().LatestCommitment().Index()); err != nil {
 					panic(ierrors.Wrap(err, "failed to restore upgrade orchestrator from disk"))
 				}
+
+				e.startupMaxBlockSlot = e.Storage.Settings().LatestCommitment().Index() + e.CurrentAPI().ProtocolParameters().MaxCommittableAge()
```
karimodm, line 205: Maybe add a comment here that this represent your "caching" window for blocks, below which blocks cannot be found in cache and should be found in storage instead. I would also rename this field to something more in line with what it actually does.

pkg/protocol/engine/engine.go
```diff
@@ -480,17 +504,22 @@ func (e *Engine) setupPruning() {
 // If setting the root commitment based on the last evicted slot this basically means we won't be able to solidify another
 // chain beyond a window based on eviction, which in turn is based on acceptance. In case of a partition, this behavior is
 // clearly not desired.
-func (e *Engine) EarliestRootCommitment(lastFinalizedSlot iotago.SlotIndex) (earliestCommitment *model.Commitment, valid bool) {
-	earliestRootCommitmentID, valid := e.EvictionState.EarliestRootCommitmentID(lastFinalizedSlot)
-	if !valid {
-		return nil, false
+func (e *Engine) EarliestRootCommitment(lastFinalizedSlot iotago.SlotIndex) (earliestCommitment *model.Commitment) {
+	maxCommittableAge := e.APIForSlot(lastFinalizedSlot).ProtocolParameters().MaxCommittableAge()
```
karimodm, line 508: Way simpler rather than scanning all rootblocks xD

pkg/protocol/protocol.go
```diff
@@ -275,7 +289,13 @@ func (p *Protocol) ProcessBlock(block *model.Block, src network.PeerID) error {
 	}
 
 	chainCommitment := p.ChainManager.LoadCommitmentOrRequestMissing(block.ProtocolBlock().SlotCommitmentID)
+	// If the commitment is not solid (it is not known), we store the block in a small buffer and process it once we
```
karimodm, line 292:
```suggestion
	// If the commitment is not solid (its chain is not known), we store the block in a small buffer and process it once we
```

pkg/protocol/protocol_fork.go
```diff
@@ -196,7 +199,7 @@ func (p *Protocol) processFork(fork *chainmanager.Fork) (anchorBlockIDs iotago.B
 	ch := make(chan *commitmentVerificationResult)
 	defer close(ch)
 
-	commitmentVerifier := NewCommitmentVerifier(p.MainEngineInstance(), fork.ForkingPoint)
+	commitmentVerifier := NewCommitmentVerifier(p.MainEngineInstance(), fork.MainChain.Commitment(fork.ForkingPoint.Index()-1).Commitment())
```
karimodm, line 202: Isn't the ForkingPoint the last common commitment between chains already? Why you pass the -1?

karimodm: Also, the `forkingPoint` that you save in the `CommitmentVerifier` is never used: only its CW is saved during initialization.

karimodm (changes requested):

pkg/blockfactory/blockissuer.go
```diff
@@ -163,24 +167,53 @@ func (i *BlockIssuer) CreateValidationBlock(ctx context.Context, opts ...options
 	return modelBlock, nil
 }
 
+func (i *BlockIssuer) getCommitment(blockSlot iotago.SlotIndex) (*iotago.Commitment, error) {
+	protoParams := i.protocol.CurrentAPI().ProtocolParameters()
+	commitment := i.protocol.MainEngineInstance().Storage.Settings().LatestCommitment().Commitment()
+
+	if blockSlot > commitment.Index+protoParams.MaxCommittableAge() {
```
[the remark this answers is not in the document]
jonastheis: Exactly. For normal users that is the desired behavior. Either you need to backdate or you need to force empty commitments.

pkg/blockfactory/blockissuer.go
```diff
@@ -163,24 +167,53 @@ func (i *BlockIssuer) CreateValidationBlock(ctx context.Context, opts ...options
...
+	}
+
+	return commitment, nil
+}
+
 // CreateBlock creates a new block with the options.
 func (i *BlockIssuer) CreateBlock(ctx context.Context, opts ...options.Option[BlockParams]) (*model.Block, error) {
 	blockParams := options.Apply(&BlockParams{}, opts)
 
+	if blockParams.IssuingTime == nil {
```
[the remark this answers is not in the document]
jonastheis: All these commitment and blockissuer related things should be addressed in #274. Here I just had to implement something to even make it work as we didn't have a way to produce valid blocks. 

For general cleanup of the blockissuer I created #279. All of this should be addressed in detail there imo

pkg/protocol/protocol_fork.go
jonastheis: Forking point is the first different commitment. We can maybe just pass the weight of the previous commitment. It is needed to correctly compute the cumulative weight.

pkg/protocol/engine/engine.go
jonastheis: What name would you suggest?

jonastheis: Added a comment and renamed to `startupAvailableBlocksWindow`

pkg/protocol/engine/blockdag/inmemoryblockdag/blockdag.go
```diff
@@ -101,6 +122,12 @@ func (b *BlockDAG) Attach(data *model.Block) (block *blocks.Block, wasAttached b
 	if block, wasAttached, err = b.attach(data); wasAttached {
 		b.events.BlockAttached.Trigger(block)
 
+		if block.SlotCommitmentID().Index() > b.latestCommitmentFunc().Commitment().Index {
```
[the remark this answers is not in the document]
jonastheis: I adjusted it to `AddWithFunc` (evaluate the condition while holding the lock) or `GetAndEvict` atomically. I also added an extensive comment on why this buffer is necessary and what its effects are.

pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -69,13 +67,14 @@ type Manager struct {
...
-func NewProvider(attestationCommitmentOffset iotago.SlotIndex) module.Provider[*engine.Engine, attestation.Attestations] {
+func NewProvider() module.Provider[*engine.Engine, attestation.Attestations] {
 	return module.Provide(func(e *engine.Engine) attestation.Attestations {
 		latestCommitment := e.Storage.Settings().LatestCommitment()
+
 		return NewManager(
 			latestCommitment.Index(),
 			latestCommitment.CumulativeWeight(),
-			attestationCommitmentOffset,
+			e.CurrentAPI().ProtocolParameters().MaxCommittableAge(),
```
alexsporn, line 77: don't hardcode this, instead pass the provider and fetch it where you use it.

pkg/testsuite/mock/node.go
```diff
@@ -131,35 +137,35 @@ func (n *Node) hookEvents() {
...
-func (n *Node) HookLogging() {
+func (n *Node) hookLogging(failOnBlockFiltered bool) {
 	events := n.Protocol.Events
 
-	n.attachEngineLogs(n.Protocol.MainEngineInstance())
+	n.attachEngineLogs(failOnBlockFiltered, n.Protocol.MainEngineInstance())
 
 	events.Network.BlockReceived.Hook(func(block *model.Block, source identity.ID) {
-		fmt.Printf("%s > Network.BlockReceived: from %s %s - %d\n", n.Name, source, block.ID(), block.ID().Index())
+		fmt.Printf("%s > network.BlockReceived: from %s %s - %d\n", n.Name, source, block.ID(), block.ID().Index())
```
alexsporn, line 146: why lowercased network? IDE rename strikes again?

alexsporn (changes requested):
Check the APIProvider comment above

piotrm50 (approved):

pkg/protocol/engine/engine.go
karimodm: Maybe `cachedBlocksWindow`?

pkg/protocol/engine/blockdag/inmemoryblockdag/blockdag.go
```diff
@@ -101,6 +121,20 @@ func (b *BlockDAG) Attach(data *model.Block) (block *blocks.Block, wasAttached b
...
 		b.events.BlockAttached.Trigger(block)
 
+		// We add blocks that commit to a commitment we haven't committed ourselves yet to this limited size buffer and
+		// only let them become solid once we committed said slot ourselves (to the same commitment).
+		// This is necessary in order to make sure that all necessary state is available after a block is solid (specifically
+		// the state of the referenced commitment for the commitment filter). All the while, we need to make sure that
+		// the requesting of missing blocks (done in b.attach) continues so that we can advance our state and eventually
+		// commit the slot.
+		// This limited size buffer has a nice side effect: In normal behavior (e.g. no attack of a neighbor that sends you
+		// unsolidifiable blocks in your committed slots) it will prevent the node from storing too many blocks in memory.
```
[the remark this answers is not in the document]
jonastheis: This limits the ability for a node to run out of memory without warpsync. Which is obviously a good thing as we will start to through away blocks that are committing to not yet committed slots. Since the buffer is an LRU cache, we will ultimately request the more recent blocks and should be able to eventually solidify.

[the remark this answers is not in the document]
karimodm: We have already concluded that this should be a compatible mechanism. In theory, when warpsync is there, there is not reason to keep the protocol-level buffer of blocks: as you will solidify the commitment chain, drop the corresponding block and delegate the solidification to warpsync instead of caching observed unsolid blocks directly.
We although decided to also keep the protocol buffer in place for now, as, without it, it will complicate syntetic tests that don't rely on a constant activity: in fact, if you throw away the block that triggered the initial commitment chain solidification, you will need to observe another one building on the (now solid) commit chain to trigger a warpsync against the matched engine's chain.

pkg/testsuite/mock/node.go
jonastheis: Seems like it

pkg/protocol/engine/engine.go
jonastheis: I don't think they are cached. They are accepted blocks from the previous "run" and only available if you restart your node. So I think what I have now `startupAvailableBlocksWindow` fits better

pkg/blockfactory/blockissuer.go
```diff
@@ -163,24 +167,53 @@ func (i *BlockIssuer) CreateValidationBlock(ctx context.Context, opts ...options
...
+	if blockSlot > commitment.Index+protoParams.MaxCommittableAge() {
+		return nil, ierrors.Errorf("can't issue block: block slot %d is too far in the future, latest commitment is %d", blockSlot, commitment.Index)
+	}
+
+	if blockSlot < commitment.Index+protoParams.MinCommittableAge() {
+		if blockSlot < protoParams.MinCommittableAge() || commitment.Index < protoParams.MinCommittableAge() {
+			return commitment, nil
+		}
+
+		commitmentSlot := commitment.Index - protoParams.MinCommittableAge()
```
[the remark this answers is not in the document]
jonastheis: I don't see how if `MinCommittableAge < MaxCommittableAge` which should be given sensible protocol parameters.

pkg/blockfactory/blockissuer.go
```diff
@@ -163,24 +167,53 @@ func (i *BlockIssuer) CreateValidationBlock(ctx context.Context, opts ...options
...
+func (i *BlockIssuer) getCommitment(blockSlot iotago.SlotIndex) (*iotago.Commitment, error) {
+	protoParams := i.protocol.CurrentAPI().ProtocolParameters()
+	commitment := i.protocol.MainEngineInstance().Storage.Settings().LatestCommitment().Commitment()
+
+	if blockSlot > commitment.Index+protoParams.MaxCommittableAge() {
+		return nil, ierrors.Errorf("can't issue block: block slot %d is too far in the future, latest commitment is %d", blockSlot, commitment.Index)
+	}
+
+	if blockSlot < commitment.Index+protoParams.MinCommittableAge() {
+		if blockSlot < protoParams.MinCommittableAge() || commitment.Index < protoParams.MinCommittableAge() {
```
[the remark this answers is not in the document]
jonastheis: As `blockSlot` is provided by the user it can be any slot. We only provide facility to create a _syntactically valid_ block here. It doesn't mean that it's a going to be _useful_ block. Therefore, we make sure that if `blockSlot` is smaller than `commitment.Index` that we either return said commitment (in case it's too close to 0) or retrieve a correct one for this `blockSlot` later on

alexsporn (approved):

karimodm (approved):

PR #246, opened 2023-08-02
Adaptive RMC

### LINKED ISSUE #235: Implement adaptive RMC
Implement the additive increase additive decrease RMC adaptation algorithm. 

Add to the mana manager or as a new component.

cyberphysic4l:
This PR introduces the RMC manager for calculating the new RMC upon slot commitment.

This also adds the RMC check to the accounts filter and adds accounts filter tests (which I forgot to add in the accounts filter PR...).

close #235


pkg/model/block.go
```diff
@@ -131,3 +131,12 @@ func (blk *Block) String() string {
 
 	return out.String()
 }
+
+func (blk *Block) ManaCost(rmc iotago.Mana) (iotago.Mana, error) {
```
alexsporn, line 135: I feel like this should be in iota.go

pkg/protocol/engine/commitmentfilter/accountsfilter/commitmentfilter.go
```diff
@@ -145,6 +149,29 @@ func (c *CommitmentFilter) evaluateBlock(block *model.Block) {
...
+	}
+	manaCost, err := block.ManaCost(rmc)
+	if err != nil {
+		c.events.BlockFiltered.Trigger(&commitmentfilter.BlockFilteredEvent{
+			Block:  block,
+			Reason: ierrors.Wrapf(err, "could not calculate Mana cost for block"),
+		})
+	}
+	basicBlock, isBasic := block.BasicBlock()
+	if isBasic && manaCost < basicBlock.BurnedMana {
```
alexsporn, line 168: you can inline the variables in the if since they are not used outside of it

pkg/protocol/engine/congestioncontrol/rmc/rmc.go
```diff
@@ -0,0 +1,151 @@
...
+
+func NewManager(commitmentLoader func(iotago.SlotIndex) (*model.Commitment, error)) *Manager {
+	return &Manager{
+		slotWork:         shrinkingmap.New[iotago.SlotIndex, iotago.WorkScore](),
+		rmc:              shrinkingmap.New[iotago.SlotIndex, iotago.Mana](),
+		commitmentLoader: commitmentLoader,
+	}
+}
+
+func (m *Manager) SetRMCParameters(rmcMin iotago.Mana, rmcIncrease iotago.Mana, rmcDecrease iotago.Mana, increaseThreshold iotago.WorkScore, decreaseThreshold iotago.WorkScore) {
```
alexsporn, line 44: You should use an APIProvider and get the correct RMC values for the current slot/epoch else this won't work with a protocol upgrade if the values change

pkg/protocol/engine/ledger/ledger/ledger.go
```diff
@@ -74,8 +76,20 @@ func NewProvider() module.Provider[*engine.Engine, ledger.Ledger] {
...
-			l.accountsLedger.SetCommitmentEvictionAge(iotagoAPI.ProtocolParameters().MaxCommittableAge())
-			l.accountsLedger.SetLatestCommittedSlot(e.Storage.Settings().LatestCommitment().Index())
+			maxCommittableAge := iotagoAPI.ProtocolParameters().MaxCommittableAge()
+			latestCommittedSlot := e.Storage.Settings().LatestCommitment().Index()
+			l.accountsLedger.SetCommitmentEvictionAge(maxCommittableAge)
+			l.accountsLedger.SetLatestCommittedSlot(latestCommittedSlot)
+			l.rmcManager = rmc.NewManager(e.Storage.Commitments().Load)
+			l.rmcManager.SetCommitmentEvictionAge(maxCommittableAge)
+			l.rmcManager.SetLatestCommittedSlot(latestCommittedSlot)
+			l.rmcManager.SetRMCParameters(
```
alexsporn, line 86: See comment above, pass the apiProvider instead

pkg/protocol/engine/ledger/ledger/ledger.go
```diff
@@ -74,8 +76,20 @@ func NewProvider() module.Provider[*engine.Engine, ledger.Ledger] {
 			// TODO: how do we want to handle changing API here?
 			iotagoAPI := l.apiProvider.CurrentAPI()
 			l.manaManager = mana.NewManager(iotagoAPI.ManaDecayProvider(), iotagoAPI.ProtocolParameters().RentStructure(), l.resolveAccountOutput)
-			l.accountsLedger.SetCommitmentEvictionAge(iotagoAPI.ProtocolParameters().MaxCommittableAge())
-			l.accountsLedger.SetLatestCommittedSlot(e.Storage.Settings().LatestCommitment().Index())
+			maxCommittableAge := iotagoAPI.ProtocolParameters().MaxCommittableAge()
+			latestCommittedSlot := e.Storage.Settings().LatestCommitment().Index()
+			l.accountsLedger.SetCommitmentEvictionAge(maxCommittableAge)
```
alexsporn, line 81: This should also not hardcode the value but fetch it from the active protocol parameters using an APIProvider

pkg/model/block.go
cyberphysic4l: I agree, I have added a todo and linked an issue to do this properly when we add scheduler params because it is tricky at the moment anyway and I have added a simplification for now

pkg/protocol/engine/congestioncontrol/rmc/rmc.go
```diff
@@ -0,0 +1,143 @@
...
+	// load the slotWork for the current slot
+	currentSlotWork, exists := m.slotWork.Get(index)
+	if !exists {
+		currentSlotWork = 0
+	}
+	// calculate the new RMC
+	var newRMC iotago.Mana
+	rmcParameters := m.apiProvider.APIForSlot(index).ProtocolParameters().RMCParameters()
+	if currentSlotWork < rmcParameters.DecreaseThreshold {
+		// TODO: use safemath here
```
alexsporn, line 86: are you going to fix this TODO?

pkg/protocol/engine/ledger/ledger/ledger.go
```diff
@@ -74,8 +76,10 @@ func NewProvider() module.Provider[*engine.Engine, ledger.Ledger] {
 			// TODO: how do we want to handle changing API here?
 			iotagoAPI := l.apiProvider.CurrentAPI()
 			l.manaManager = mana.NewManager(iotagoAPI.ManaDecayProvider(), iotagoAPI.ProtocolParameters().RentStructure(), l.resolveAccountOutput)
```
alexsporn, line 78: the manaManager should probably also take the provider directly and not the current api

pkg/protocol/engine/congestioncontrol/rmc/rmc.go
```diff
@@ -0,0 +1,146 @@
...
+	m.slotWork.Delete(index)
+
+	// evict old RMC from current slot - maxCommittableAge
+	maxCommittableAge := m.apiProvider.APIForSlot(index).ProtocolParameters().MaxCommittableAge()
+	if index > maxCommittableAge {
+		m.rmc.Delete(index - maxCommittableAge)
+	}
+
+	// update latestCommittedIndex
+	if index <= m.latestCommittedSlot {
```
karimodm, line 113: Shouldn't this check happen at the beginning of the method?

pkg/protocol/engine/congestioncontrol/rmc/rmc.go
```diff
@@ -0,0 +1,146 @@
...
+		return 0, ierrors.Errorf("cannot get RMC for slot %d: not committed yet", slot)
+	}
+	// this should never happen when checking the RMC for a slot that is not committed yet
+
+	if slot+m.apiProvider.CurrentAPI().ProtocolParameters().MaxCommittableAge() < m.latestCommittedSlot {
+		return 0, ierrors.Errorf("cannot get RMC for slot %d: already evicted", slot)
+	}
+
+	rmc, exists := m.rmc.Get(slot)
+	if !exists {
```
karimodm, line 135: Here you allow to load RMC for slots older than maxCommittableAge, should this ever happen?
Also you would never evict them again as you only evict index - maxCommittableAge in `CommitSlot`

alexsporn (approved):

PR #249, opened 2023-08-03
Implement validator block issuance component.

piotrm50:
Component replaces the previous Activity component. 
The new one is aware if the BlockIssuer's account is a validator and:
* does not issue anything if the account is not a registered validator
* issues ValidatorBlock at regular intervals in an epoch in which the account is part of the committee
* issues Blocks at regular intervals to register the account as a candidate to be a validator in an epoch in which the account is not part of the committee.


components/activity/params.go
```diff
@@ -9,7 +9,7 @@ import (
 // ParametersActivity contains the definition of the configuration parameters used by the Activity component.
 type ParametersActivity struct {
 	// Enabled defines whether the Activity component is enabled.
-	Enabled bool `default:"true" usage:"whether the Activity component is enabled"`
+	Enabled bool `default:"false" usage:"whether the Activity component is enabled"`
```
alexsporn, line 12: Do we still need the activity component then? I think you can remove it

piotrm50: Removed.

components/validator/issuer.go
```diff
@@ -0,0 +1,66 @@
...
+				Tag: []byte("CANDIDATE BLOCK"),
+			}),
+		)
+		if err != nil {
+			Component.LogWarnf("error creating candidate validator block: %s", err.Error())
+
+			return
+		}
+
+		nextBroadcast = blockIssuingTime.Add(ParamsValidator.CommitteeBroadcastInterval)
```
karimodm, line 55: This should be CandidateBroadcastInterval

components/validator/issuer.go
```diff
@@ -0,0 +1,66 @@
...
+			}),
+		)
+		if err != nil {
+			Component.LogWarnf("error creating committee validator block: %s", err.Error())
+
+			return
+		}
+	} else {
+		var err error
+		modelBlock, err = deps.BlockIssuer.CreateBlock(ctx,
```
karimodm, line 43: We should be issuing validator blocks, so we can track performance just the same, but with a lower emission frequency.

components/validator/issuer.go
karimodm: A comment would be nice here to explain that you are overriding a variable that doesn't get captured by the defer above!

karimodm (changes requested):
We still need the activity component, no?
If you have stake in your node account you are either a validator and emit validator blocks, or you are trying to become a validator emitting candidate validator blocks.
But if I have no stake, but I have an account with BIC I should still be able to issue blocks: hence the activity component. We still want some activity in the network in form of regular blocks not coming from staking accounts.

piotrm50:
> We still need the activity component, no? If you have stake in your node account you are either a validator and emit validator blocks, or you are trying to become a validator emitting candidate validator blocks. But if I have no stake, but I have an account with BIC I should still be able to issue blocks: hence the activity component. We still want some activity in the network in form of regular blocks not coming from staking accounts.

But do we need an activity plugin then that is built-in the node? I think it makes more sense to simply use a block spammer in this case, no?

components/validator/issuer.go
piotrm50: Done.

components/validator/issuer.go
piotrm50: Done. The only thing is that when implementing tip selection for validator blocks, we should keep track of validator blocks from current committee members.

karimodm (approved):

PR #254, opened 2023-08-03
Resolve all inputs in Mempool

### LINKED ISSUE #159: Refactor VM execution closure in Mempool
...so we can wait for resolving commitment input and al.

Blocked by #158 and #139

karimodm:
This PR also offers a refactor of Inputs into an interface container called `State`. An UTXO or a Context state complies instead with extended interfaces called `OutputState` and `ContextState` respectively.
Simarly, when a state managed to get resolved, its metadata state is wrapped in a `OutputStateMetadata` and `ContextStateMetadata` interfaces respectively.
The Mempool now is able to "wait" for the resolution of context inputs, wherever they might be according to the above mentioned interfaces, and passes the resolved contexts to the VM for cross-validation and transition execution. Currently, the only context input that can be passed resolved is the CommitmentInput (called `timeReference` in the VM context); in fact, BIC and Reward inputs rely on other inputs for their resolution and intrinsic validation: input commitment for Accounts, and UTXO input with staking or delegation feature for the Reward inputs.

Depends on https://github.com/iotaledger/iota.go/pull/474
 
Resolves #159


pkg/protocol/engine/ledger/ledger/ledger.go
```diff
@@ -644,22 +649,38 @@ func (l *Ledger) resolveState(stateRef iotago.Input) *promise.Promise[mempool.St
...
+		}
+
+		// the commitment is not yet available
+		if loadedCommitment == nil {
+			return p.Reject(ierrors.Wrapf(iotago.ErrCommitmentInputInvalid, "failed to load commitment at index %d: %w", concreteStateRef.CommitmentID.Index(), mempool.ErrStateNotFound))
+		}
+
+		return p.Resolve(loadedCommitment)
+	case iotago.InputBlockIssuanceCredit, iotago.InputReward:
+		// these are always resolved as they depend on the commitment or UTXO inputs
```
piotrm50, line 682: Shouldn't we check that the commitment in the input is correct?

pkg/protocol/engine/mempool/v1/mempool.go
```diff
@@ -192,17 +206,29 @@ func (m *MemPool[VoteRank]) solidifyInputs(transaction *TransactionMetadata) {
...
+			switch state.Type() {
+			case iotago.InputUTXO:
+				outputStateMetadata := state.(*OutputStateMetadata)
+
+				transaction.publishInput(index, outputStateMetadata)
+
+				if created {
+					m.setupState(outputStateMetadata)
+				}
+			case iotago.InputCommitment, iotago.InputBlockIssuanceCredit, iotago.InputReward:
```
piotrm50, line 223: We should setup an eviction mechanism for ContextInputs.

pkg/protocol/engine/mempool/v1/mempool.go
```diff
@@ -115,20 +115,34 @@ func (m *MemPool[VoteRank]) TransactionMetadata(id iotago.TransactionID) (transa
...
+	})
 	stateRequest.OnError(func(requestErr error) { err = ierrors.Errorf("failed to request state: %w", requestErr) })
 	stateRequest.WaitComplete()
 
 	return state, err
 }
 
+func (m *MemPool[VoteRank]) PublishCommitmentState(commitment *iotago.Commitment) {
+	stateRequest, _ := m.cachedStateRequests.GetOrCreate(commitment.StateID(), lo.NoVariadic(promise.New[mempool.State]))
+	stateRequest.Resolve(commitment)
```
piotrm50, line 143: We should not preemtively create promises for commitments. Here we should only resolve existing promises.

If we already have a commitment and a transaction has that on the input side, it should by lazily loaded in `solidifyInputs` logic.

That will make it easier to manage eviction of those.

karimodm: Makes sense, fixed here to only resolve if it was previously requested.

karimodm: InputBlockIssuanceCredit

pkg/protocol/engine/ledger/ledger/vm.go
```diff
@@ -144,11 +135,15 @@ func (l *Ledger) executeStardustVM(_ context.Context, stateTransition mempool.Tr
 func (l *Ledger) loadCommitment(inputCommitmentID iotago.CommitmentID) (*iotago.Commitment, error) {
```
piotrm50, line 135: Since this method is only used in the ledger.go, it should be moved there.

pkg/protocol/engine/mempool/v1/mempool.go
```diff
@@ -192,17 +209,39 @@ func (m *MemPool[VoteRank]) solidifyInputs(transaction *TransactionMetadata) {
...
+				contextStateMetadata := state.(*ContextStateMetadata)
+
+				transaction.publishCommitmentInput(contextStateMetadata)
+
+				if created {
+					contextStateMetadata.onAllSpendersRemoved(func() { m.cachedStateRequests.Delete(contextStateMetadata.StateID()) })
+				}
+			case iotago.InputBlockIssuanceCredit, iotago.InputReward:
+			default:
+				// TODO: how to handle error?
```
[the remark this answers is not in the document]
karimodm: I still think this should be a panic though, because it means we successfully resolved a unknown Input type.

piotrm50 (approved):

Issue #50, opened 2023-04-20
Implement core REST API


jonastheis:
Many of the APIs are implemented in #72 

```[tasklist]
### TODOs
- [x] control APIs
- [x] peers APIs
- [x] blockReason, TxState, TxReason in BlockMetadata
- [x] CommitmentIDSpent, TransactionIDSpent in OutputMetadata
- [x] pruningSlot, baseToken in `/info`
- [x] accounts retrieval
- [x] review Swagger file
- [x] Implement health check
- [x] Finish congestion api (scheduler)
- [x] rewards api: epoch should be returned by the reward calculations
- [x] pagination for stakers endpoint
- [ ] Finish congestion: RMC
- [ ] extra fields to add to CommitmentResponse: RMC
- [ ] Finish staking api (supported protocol version of each validator)
- [ ] finish  `checkUpcomingUnsupportedProtocolVersion`
- [ ] In `responseByHeader`, encode object with its api version
- [ ] add performance factor to committee response
- [ ] does event API needs any updates
```

PR #275, opened 2023-08-07
Update core API endpoints

daria305:
Issue #50  

Implement:
 - add client lib functions and tests
 - update staking response with pagination, it contains registered validators not only active
 - rewards endpoint returns epoch range that was used
 - other missing endpoints and fixes
 - implement RMC endpoint

What's left:
 - protocol versioning, probably should be tracked within performance tracker, as we need latest versions of registered validator, not only committee members
 - evalevaluate  event API


jkrvivian (changes requested):

components/restapi/core/accounts.go
```diff
@@ -31,33 +37,43 @@ func congestionForAccountID(c echo.Context) (*apimodels.CongestionResponse, erro
...
+	pageSize, _ := httpserver.ParseUint32QueryParam(c, restapipkg.QueryParameterPageSize)
+	requestedSlotIndex, cursorIndex, _ := httpserver.ParseCursorQueryParam(c, restapipkg.QueryParameterCursor)
 	latestCommittedSlot := deps.Protocol.SyncManager.LatestCommitment().Index()
-	nextEpoch := deps.Protocol.APIForSlot(latestCommittedSlot).TimeProvider().EpochFromSlot(latestCommittedSlot) + 1
 
-	activeValidators, err := deps.Protocol.MainEngineInstance().SybilProtection.EligibleValidators(nextEpoch)
-	if err != nil {
-		return nil, err
+	if cursorIndex == 0 {
+		requestedSlotIndex = latestCommittedSlot
```
jkrvivian, line 52: Why is `requestedSlotIndex` set to `latestCommittedSlot` if `cursorIndex` == 0?
Or how would the request be if requesting the first page of a specific slot index?

daria305: the cursor index was supposed to be 0 for the first page, that's why I set the requestAtSlotIndex to latestCommittedSlot (that represent current slot and snapshot responses based on it)
But cursor is now string, I could just check for err to see if it was provided. I change also last page response.

components/restapi/core/accounts.go
```diff
@@ -21,49 +28,73 @@ func congestionForAccountID(c echo.Context) (*apimodels.CongestionResponse, erro
...
-	activeValidators, err := deps.Protocol.MainEngineInstance().SybilProtection.EligibleValidators(nextEpoch)
-	if err != nil {
-		return nil, err
+	slotRange := uint32(requestedSlotIndex / RequestsMemoryCacheGranularity)
+	registeredValidators, exists := deps.Protocol.MainEngineInstance().Retainer.RegisteredValidatorsCache(slotRange)
+	if !exists {
+		var err error
+		registeredValidators, err = deps.Protocol.MainEngineInstance().SybilProtection.OrderedRegisteredValidatorsList(nextEpoch)
+		if err != nil {
+			return nil, err
```
karimodm, line 77: Wrap the error

pkg/protocol/sybilprotection/sybilprotection.go
```diff
@@ -8,13 +8,16 @@ import (
...
 )
 
 type SybilProtection interface {
 	TrackValidationBlock(block *blocks.Block)
 	EligibleValidators(epoch iotago.EpochIndex) (accounts.AccountsData, error)
-	ValidatorReward(validatorID iotago.AccountID, stakeAmount iotago.BaseToken, epochStart, epochEnd iotago.EpochIndex) (validatorReward iotago.Mana, err error)
-	DelegatorReward(validatorID iotago.AccountID, delegatedAmount iotago.BaseToken, epochStart, epochEnd iotago.EpochIndex) (delegatorsReward iotago.Mana, err error)
+	OrderedRegisteredValidatorsList(epoch iotago.EpochIndex) ([]*apimodels.ValidatorResponse, error)
+	IsActive(validatorID iotago.AccountID, epoch iotago.EpochIndex) bool
+	ValidatorReward(validatorID iotago.AccountID, stakeAmount iotago.BaseToken, epochStart, epochEnd iotago.EpochIndex) (validatorReward iotago.Mana, usedStart, usedEnd iotago.EpochIndex, err error)
```
karimodm, line 19: I think the returned epoch values should have more descriptive names, and you should provide a comment description of what they mean.

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
```diff
@@ -311,6 +315,52 @@ func (o *SybilProtection) EligibleValidators(epoch iotago.EpochIndex) (accounts.
 	return validators, nil
 }
 
+// IsActive returns true if the given validator is currently active.
+func (o *SybilProtection) IsActive(validatorID iotago.AccountID, epoch iotago.EpochIndex) bool {
```
karimodm, line 319: This method should be called IsActiveCandidate, no?

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
```diff
@@ -311,6 +315,52 @@ func (o *SybilProtection) EligibleValidators(epoch iotago.EpochIndex) (accounts.
...
+	activeCandidates := o.performanceTracker.EligibleValidatorCandidates(epoch)
+	return activeCandidates.Has(validatorID)
+}
+
+// OrderedRegisteredValidatorsList returns the currently known list of registered validator candidates for the given epoch.
+func (o *SybilProtection) OrderedRegisteredValidatorsList(epoch iotago.EpochIndex) ([]*apimodels.ValidatorResponse, error) {
```
karimodm, line 325: Should be called OrderedRegisteredCandidateValidatorsList

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
```diff
@@ -311,6 +315,52 @@ func (o *SybilProtection) EligibleValidators(epoch iotago.EpochIndex) (accounts.
...
+	validatorResp := make([]*apimodels.ValidatorResponse, 0, candidates.Size())
+	if err := candidates.ForEach(func(candidate iotago.AccountID) error {
+		accountData, exists, err := o.ledger.Account(candidate, o.lastCommittedSlot)
+		if err != nil {
+			return err
+		}
+		if !exists {
+			return ierrors.Errorf("account of committee candidate does not exist: %s", candidate)
+		}
+		if accountData.StakeEndEpoch < epoch {
```
karimodm, line 338: Is the StakeEndEpoch inclusive or exclusive? in other words: is the StakeEndEpoch the last epoch in which your stake is active or is the first epoch such stake doesn't apply anymore?

karimodm (changes requested):

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
daria305: From TIP: Feature whose `End Epoch` is the current one or has passed, it is no longer considered for validator selection and the. So the cond was incorrect, changed to `<=`

jkrvivian (approved):

karimodm (approved):

PR #276, opened 2023-08-08
Properly filter for validation blocks in the blockgadget, attestations, upgrade orchestrator and sybilprotection

alexsporn:
I did not separate the events but renamed the functions expecting validation blocks and added checks to verify the passed block is a validation block

### DIFF
--- a/pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go
+++ b/pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go
@@ -56,15 +56,17 @@ func (t *Tracker) RegisterCommittee(epoch iotago.EpochIndex, committee *account.
 	return t.committeeStore.Set(epoch, committee)
 }
 
-func (t *Tracker) BlockAccepted(block *blocks.Block) {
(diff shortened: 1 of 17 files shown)


pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go
```diff
@@ -56,23 +56,28 @@ func (t *Tracker) RegisterCommittee(epoch iotago.EpochIndex, committee *account.
...
-
 	performanceFactors := t.performanceFactorsFunc(block.ID().Index())
 	pf, err := performanceFactors.Load(block.ProtocolBlock().IssuerID)
 	if err != nil {
 		// TODO replace panic with errors in the future, like triggering an error event
 		panic(ierrors.Errorf("failed to load performance factor for account %s", block.ProtocolBlock().IssuerID))
 	}
 
+	// TODO: store highest supported version per validator?
+	_ = validatorBlock.HighestSupportedVersion
```
jonastheis, line 79: If we want to store this we should store the version and hash together.

jonastheis (approved):

PR #280, opened 2023-08-09
Feat: WarpSync

### LINKED ISSUE #193: Warpsyncing
```[tasklist]
### Tasks
- [ ] `pkg/protocol/engine/engine.go`, in `setupBlockRequester`, only request missing blocks when not in warpsync range.
```

hmoog:
Resolves #193 

This PR changes a few key things on how block are processed and solidified:

- A separate BlockDispatcher component has been created to dispatch blocks to their respective engines and hook up all the network events associated to it. This functionality has been mostly stripped from the main engine, to make the dataflow more self-contained.
- A new CommittedSlotAPI has been introduced to obtain basic components of a slot commitment: blocks, roots, etc.
- The warpsync component works in the following fashion:
  - If a block is received from the network layer, and its commitment is unsolid it is parked in a protocol-level buffer for unsolid commitments and its chain solidified.
  - If a block is within [lastCommittedSlot, lastCommittedSlot+2] of either the main or the candidate engine, and it is on the same chain, the block is processed directly from the Engine, and solidified future to past.
  - If a block exceeds lastCommittedSlot+2:
    - It gets dropped.
    - All the block IDs belonging to such committment get requested via a warpsync request.
    - The IDs received are directly injected as missing blocks in the BlockDAG, and therefore requested through the solidifier.
    - A bunch of slots are requested in parallel, from past to future, starting from lastCommittedSlot up to `warpSyncWindowSize`. Please note that `warpSyncWindowSize` needs to be at least minCommittableAge + 1, as otherwise the engine won't possibly advance the clock enough to make the earliest slot of the window committable.
    - Each time a commitment is produced (published) the step above is repeated, until receiving only blocks belonging to the regular solidification window.


pkg/protocol/block_dispatcher.go
```diff
@@ -0,0 +1,345 @@
...
+
+// initEngineMonitoring initializes the automatic monitoring of the engine instances.
+func (b *BlockDispatcher) initEngineMonitoring() {
+	b.monitorLatestEngineCommitment(b.protocol.mainEngine)
+
+	b.protocol.engineManager.OnEngineCreated(b.monitorLatestEngineCommitment)
+
+	b.protocol.Events.ChainManager.CommitmentPublished.Hook(func(chainCommitment *chainmanager.ChainCommitment) {
+		chainCommitment.SolidEvent().OnTrigger(func() {
+			b.runTask(func() {
```
jonastheis, line 108: Is this `b.runTask` everywhere really necessary?

pkg/protocol/block_dispatcher.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+// initEngineMonitoring initializes the automatic monitoring of the engine instances.
+func (b *BlockDispatcher) initEngineMonitoring() {
+	b.monitorLatestEngineCommitment(b.protocol.mainEngine)
+
+	b.protocol.engineManager.OnEngineCreated(b.monitorLatestEngineCommitment)
+
+	b.protocol.Events.ChainManager.CommitmentPublished.Hook(func(chainCommitment *chainmanager.ChainCommitment) {
+		chainCommitment.SolidEvent().OnTrigger(func() {
```
jonastheis, line 107: Can we add a comment about the flow here? This seems to be a key action why this whole thing works: once a commitment becomes solid in a chain, we know which engine it belongs to. Which in turn will allow us to advance the engine (and thus its created commitments/committed slots).

pkg/protocol/block_dispatcher.go
```diff
@@ -0,0 +1,345 @@
...
+	maxCommittableAge := engine.APIForSlot(slotCommitmentID.Index()).ProtocolParameters().MaxCommittableAge()
+
+	return block.ID().Index() > latestCommitmentIndex+maxCommittableAge
+}
+
+// warpSync triggers warp sync from the latest committed slot up to the warpsync window.
+func (b *BlockDispatcher) warpSync(e *engine.Engine, chainCommitment *chainmanager.ChainCommitment) {
+	if e == nil || chainCommitment == nil {
+		return
+	}
```
jonastheis, line 255: Weird design. Shouldn't it be the responsibility of the caller to make sure this is not nil?

pkg/protocol/block_dispatcher.go
```diff
@@ -0,0 +1,345 @@
...
+
+	chain := chainCommitment.Chain()
+	maxCommittableAge := e.APIForSlot(chainCommitment.Commitment().Index()).ProtocolParameters().MaxCommittableAge()
+	warpSyncWindowSize := lo.Cond(maxCommittableAge > b.optWarpSyncWindowSize, maxCommittableAge, b.optWarpSyncWindowSize)
```
jonastheis, line 259: isn't that just `lo.Max`?

pkg/protocol/block_dispatcher.go
```diff
@@ -0,0 +1,345 @@
...
+
+	b.protocol.Events.Engine.SlotGadget.SlotFinalized.Hook(b.evict)
+}
+
+// initNetworkConnection initializes the network connection of the BlockDispatcher instance.
+func (b *BlockDispatcher) initNetworkConnection() {
+	b.protocol.Events.Engine.BlockRequester.Tick.Hook(func(blockID iotago.BlockID) {
+		b.runTask(func() {
+			b.protocol.networkProtocol.RequestBlock(blockID)
+		}, b.dispatchWorkers)
```
jonastheis, line 136: Why using `b.dispatchWorkers`? Oh I see, seems that otherwise there's a task conention or so because `processWarpSyncResponse` triggers `BlockMissing` which in turn starts the ticker and then all these tasks would be queue until 
```
for _, blockID := range blockIDs {
	targetEngine.BlockDAG.GetOrRequestBlock(blockID)
}
```

is done

pkg/protocol/block_dispatcher.go
```diff
@@ -0,0 +1,345 @@
...
+	})
+
+	b.pendingWarpSyncRequests.Events.Tick.Hook(func(id iotago.CommitmentID) {
+		b.runTask(func() {
+			b.protocol.networkProtocol.SendWarpSyncRequest(id)
+		}, b.dispatchWorkers)
```
jonastheis, line 142: Why using b.dispatchWorkers?

pkg/protocol/block_dispatcher.go
```diff
@@ -0,0 +1,345 @@
...
+// initNetworkConnection initializes the network connection of the BlockDispatcher instance.
+func (b *BlockDispatcher) initNetworkConnection() {
+	b.protocol.Events.Engine.BlockRequester.Tick.Hook(func(blockID iotago.BlockID) {
+		b.runTask(func() {
+			b.protocol.networkProtocol.RequestBlock(blockID)
+		}, b.dispatchWorkers)
+	})
+
+	b.pendingWarpSyncRequests.Events.Tick.Hook(func(id iotago.CommitmentID) {
+		b.runTask(func() {
```
jonastheis, line 140: Why can't we use the workerpool functionality of the event?

pkg/protocol/block_dispatcher.go
```diff
@@ -0,0 +1,345 @@
...
+		return ierrors.Errorf("failed to verify merkle proof for %s", commitmentID)
+	}
+
+	b.pendingWarpSyncRequests.StopTicker(commitmentID)
+
+	b.processedWarpSyncRequests.Add(commitmentID)
+
+	targetEngine.Events.BlockDAG.BlockSolid.Hook(func(block *blocks.Block) {
+		block.ID()
+	})
```
jonastheis, line 226: ?

pkg/protocol/block_dispatcher.go
```diff
@@ -0,0 +1,345 @@
...
+
+		return ierrors.Errorf("failed to dispatch block %s: slot commitment %s is not solid", block.ID(), slotCommitment.ID())
+	}
+
+	matchingEngineFound := false
+	for _, engine := range []*engine.Engine{b.protocol.MainEngineInstance(), b.protocol.CandidateEngineInstance()} {
+		if engine != nil && (engine.ChainID() == slotCommitment.Chain().ForkingPoint.ID() || engine.BlockRequester.HasTicker(block.ID())) {
+			if !b.inWarpSyncRange(engine, block) {
+				engine.ProcessBlockFromPeer(block, src)
+			}
```
jonastheis, line 87: Why don't we forward this block to the engine? This basically means if the block is too far ahead we will simply discard it? If it is to prevent future blocks to bloat the memory, then it would be very nice to just put a comment here that explains exactly that.

pkg/protocol/block_dispatcher.go
```diff
@@ -0,0 +1,345 @@
...
+// This is the case if the block is more than a warp sync threshold ahead of the latest commitment while also committing
+// to a new slot that can be warp synced.
+func (b *BlockDispatcher) inWarpSyncRange(engine *engine.Engine, block *model.Block) bool {
+	if engine.BlockRequester.HasTicker(block.ID()) {
+		return false
+	}
+
+	slotCommitmentID := block.ProtocolBlock().SlotCommitmentID
+	latestCommitmentIndex := engine.Storage.Settings().LatestCommitment().Index()
+	maxCommittableAge := engine.APIForSlot(slotCommitmentID.Index()).ProtocolParameters().MaxCommittableAge()
```
jonastheis, line 246: Why don't we use the max of `warpSyncWindowSize := lo.Cond(maxCommittableAge > b.optWarpSyncWindowSize, maxCommittableAge, b.optWarpSyncWindowSize)` these two here?

jonastheis (changes requested):

pkg/protocol/chainmanager/manager.go
```diff
@@ -374,9 +374,11 @@ func (m *Manager) registerCommitment(commitment *model.Commitment) (isNew bool,
 	chainCommitment, created := m.getOrCreateCommitment(commitment.ID())
 
 	if !chainCommitment.PublishCommitment(commitment) {
-		return false, chainCommitment.IsSolid(), false, chainCommitment
+		return false, chainCommitment.SolidEvent().WasTriggered(), false, chainCommitment
 	}
 
+	m.Events.CommitmentPublished.Trigger(chainCommitment)
```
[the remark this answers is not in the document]
hmoog: This event notifies outside consumers about every commitment that the chainmanager sees and manages.

MissingCommitmentReceived is only emitted after a previously requested commitment was received and it is used to stop the requester (after starting it with a CommitmentMissing event).

So no, I don't think you can merge these events as these are two completely different use cases (get notified about each commitment, get notified about the fate of requests).

IMHO CommitmentMissing and MissingCommitmentReceived shouldn't even be exposed and simply be private, to dispatch stuff to the internal requester but we currently have some metrics that keep track of this requested stuff but the events definitely need to stay separate.

pkg/protocol/protocol.go
```diff
@@ -185,6 +184,8 @@ func (p *Protocol) shutdown() {
 		p.networkProtocol.Shutdown()
 	}
 
+	p.TriggerStopped()
```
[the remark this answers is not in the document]
hmoog: We needed to turn the protocol into a module similar to the engine so we can also hook up to the corresponding init and constructed events.

The additional Stop event we have in the Events is kind of unnecessary now, so yes I agree we should get rid of the manual one and simple use the one from the Module framework.

pkg/protocol/block_dispatcher.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+
+	return nil
+}
+
+// monitorLatestEngineCommitment monitors the latest commitment of the given engine instance and triggers a warp sync if
+// necessary.
+func (b *BlockDispatcher) monitorLatestEngineCommitment(engineInstance *engine.Engine) {
+	engineInstance.HookStopped(engineInstance.Events.Notarization.LatestCommitmentUpdated.Hook(func(commitment *model.Commitment) {
+		if chainCommitment, exists := b.protocol.ChainManager.Commitment(commitment.ID()); exists {
```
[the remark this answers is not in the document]
hmoog: Imagine you are syncing a new node.

You see a few new commitments, and then you solidify them. Once you have a solid commitment chain you start to warpsync the first n epochs.

Every time you commit the engine, you need to advance your warpsync window, to now request the next epoch so you start to slowly catch up. This is what happens here - as soon as we commit we adjust our warpsync window and queue the new slots that should be requested.

The case that you described where you are close to the tip of the chain and you i.e. see a commitment slightly earlier than you commit yourself should actually not cause a warpsync because there was a cut off condition in "warpSync" that prevented this from happening when you were close to the tip of the chain.

It was also called "warpSyncIfNecessary". Andrea seems to have removed this line of code and I am waiting for an answer for why he did that as you are right that we would currently warpsync epochs that we are about to commit ourselves.

pkg/protocol/block_dispatcher.go
hmoog: Because its not a heavy task that should be separated from the rest of the flow through a dedicated worker.

It's like sending a block request.

pkg/protocol/block_dispatcher.go
hmoog: Because we only want to hand over these tasks to the workerpool if it wasnt shut down yet, so we need to make the queuing conditional.

In general, I think we should eventually move away from providing the workerpool as an option in the event and always queue manually.

```
myEvent.Hook(func() {
    wp.Submit(func() {
        // dosth ...
    }),
)
```

isn't orders of magnitude more verbose as

```
myEvent.Hook(func() {
   // dosth ...
}, event.WithWorkerPool(wp))
```
 and it would allow us to massively simplify the events moving forward, which also makes them faster as we no longer need to allocate additional closures inside the event.

pkg/protocol/block_dispatcher.go
hmoog: Oh, this is something that was supposed to track if a slot was fully solidified as we didnt see the engine advance after we fed all the blocks inside, but we stopped to implement this function and then found the problem to be in the engine.

This was simply forgotten to be removed.

pkg/protocol/block_dispatcher.go
hmoog: I renamed the method to "inSyncWindow" and added a corresponding comment in the method description.

pkg/protocol/block_dispatcher.go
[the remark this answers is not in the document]
hmoog: After talking to Andrea, we re-added a cut off condition for the case where we are close to the tip of the chain to prevent us from warp syncing slots that we are about to commit ourselves.

pkg/protocol/protocol_network.go
```diff
@@ -10,30 +10,20 @@ import (
 )
 
 func (p *Protocol) runNetworkProtocol() {
-	p.networkProtocol = core.NewProtocol(p.dispatcher, p.Workers.CreatePool("NetworkProtocol"), p) // Use max amount of workers for networking
+	p.networkProtocol = core.NewProtocol(p.networkDispatcher, p.Workers.CreatePool("NetworkProtocol"), p) // Use max amount of workers for networking
 	p.Events.Network.LinkTo(p.networkProtocol.Events)
 
 	// TODO: the distinction between different worker pools as it is right now does not make much sense:
```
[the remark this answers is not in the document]
hmoog: I didnt write this TODO so I am not sure what is meant with that but at least part of the workers are somewhat cleaned up in the dispatcher now.

pkg/protocol/block_dispatcher.go
hmoog: fixed

pkg/protocol/block_dispatcher.go
hmoog: fixed

jonastheis (approved):

PR #283, opened 2023-08-14
Scheduler params and work score

### LINKED ISSUE #264: Add scheduler params and activate workscore
### Scheduler parameters
The following need to be protocol parameters as they would certainly cause major inconsistencies for any node that used the wrong value.

- `SchedulingRate`
- `MinMana`

These other scheduler params need to be in reasonable range but not exact because the outcome of the scheduler depends on local conditions anyway and is not exact.

- `MaxBufferSize` - needs to be in a reasonable range to prevent dropping unneccesarily in light congestion, but not exact.
- `MaxBlockWork `- this needs to be exact as other params depend on it.
- `MaxDeficit` - needs to be at least MaxBlockWork and should be several times this value to prevent problems when traffic is bursty.
- `TokenBucketSize` - needs to be at least MaxBlockWork and should be several times this value to prevent problems when traffic is bursty.
- `AcceptedBlockScheduleThreshold` - this is just an optimisation for nodes that fall out of sync a little.

### Activate Workscore
When we add `MaxBlockWork`, we should also integrate the real WorkScore function into the scheduler and RMC.

There are a number of places throughout the code with this issue marked in the TODOs where we should appropriately add workscore.

### LINKED ISSUE #281: Only burn RMC, not entire BurnedMana field
When we deduct burned mana from blocks upon committing slots in the accounts ledger, we currently do so by deductin the amount of mana specified in the BurnedMana field of the block.

Instead, we should only burn the RMC for the block to prevent accidentally burning too much, as the extra burn provides no benefit or prioritisation at this point.

cyberphysic4l:
Add scheduler params to protocol parameters and integrate WorkScore to the drr scheduler.

closes #264 
closes #281


pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
```diff
@@ -325,27 +362,31 @@ func (s *Scheduler) selectIssuer(start *IssuerQueue) (int64, *IssuerQueue) {
...
 			}
 
-			r := (remainingDeficit + int64(quantum) - 1) / int64(quantum) // round up division result
+			denom, err := safemath.SafeAdd(remainingDeficit, quantum-1)
+			if err != nil {
+				denom = s.maxDeficit()
+			}
+			r, err := safemath.SafeDiv(denom, quantum)
+			if err != nil {
+				panic("quantum = 0")
```
piotrm50, line 387: add error to the panic message

pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
```diff
@@ -382,64 +424,68 @@ func (s *Scheduler) removeIssuer(q *IssuerQueue, err error) {
...
-	s.deficits.Compute(accountID, func(currentValue uint64, exists bool) uint64 {
-		if !exists {
-			_, err = s.manaRetrieveFunc(accountID)
-			// manaRetrieveFunc return error if mana is less than MinMana or the Mana can not be retrieved for the account.
-			if err != nil {
-				err = ierrors.Wrapf(err, "could not get deficit for issuer %s", accountID)
-				return 0
-			}
+// addExistingIssuer adds an existing issuer to the scheduler with Max deficit. This is for an optimisation where we remove max deficit issuers which have been inactive for some time.
+// func (s *Scheduler) addExistingIssuer(accountID iotago.AccountID) {
```
piotrm50, line 440: Is this gonna be used?

pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
```diff
@@ -491,40 +538,12 @@ func (s *Scheduler) updateChildrenWithoutLocking(block *blocks.Block) {
...
 }
 
-// region Options ////////////////////////////////////////////////////////////////////////////////////////////////////
-
-func WithAcceptedBlockScheduleThreshold(acceptedBlockScheduleThreshold time.Duration) options.Option[Scheduler] {
-	return func(s *Scheduler) {
-		s.optsAcceptedBlockScheduleThreshold = acceptedBlockScheduleThreshold
-	}
+func (s *Scheduler) maxDeficit() Deficit {
+	return Deficit(1000000000)
```
piotrm50, line 542: Revert this to the original non-debugging value.

pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
cyberphysic4l: Not yet, so I suppose this should be removed.

piotrm50 (approved):
lgtm

PR #288, opened 2023-08-15
Epoch-based storage and pruning

jkrvivian:
This PR implements the pruning mechanisms and makes some key changes to the database structure:
* Bucket size is changed from slot-based to epoch-based, thus pruned by epoch.
* To meet the requirements of having different periods of keeping data around, 2 types of prunable storage are added (and old ones replaced, code reused):
   *   `slotStore`: Data in slotStore is accessed by `SlotIndex`, and the pruning threshold follows the default pruning threshold, which is configurable from outside
   * `epochStore`, `epochKVStore`: Data in epochStore is accessed with `EpochIndex`, each individual store has its own pruning threshold, e.g., `decidedUpgradeSignals` are kept for 7 epochs and `rewards` are kept for a year.
   * The API of both storages is updated so that an error is returned if a slot/epoch is already pruned (instead of `nil` as it was before).
 * Introduce `PruningIndex`, 
    * The new type is introduced to keep track of `lastPrunedEpoch` for each stores and stores it to kvstore, so when the node restores data from disk, the `lastPrunedEpoch` is always correct. Internally, there's also an in-memory `lastPrunedEpoch` to avoid frequently retrieving it from disk.

### Pruning
* The protocol forces to keep at least 1 full epoch before the epoch of last finalized slot, other thresholds are applied on top of it (=`latestPrunableEpoch`). Meaning that if the current finalized slot is in epoch 10, then the `latestPrunableEpoch=8`.
* Every time the nodes prunes, it also prunes the spent UTXO ledger up to the same epoch.
* 3 pruning methods are introduced to the API:
   * `PruneByEpochIndex`: user can prune data up until the given epoch
   * `PruneByDepth`: users can prune data until `latestPrunableEpoch - depth`
   * `PruneBySize`: users can prune data until the database is lower than the given `maxSize (in bytes) * reductionPercentage` 
* The node also tries to prune as soon as it can by hooking to `SlotFinalized` event,  with both `PruneByDepth` (pruning up until `optsPruningDelay` and `PruneBySize` (pruning until `maxSize (in bytes) * reductionPercentage`) if enabled, respectively.

Depends on https://github.com/iotaledger/iota.go/pull/480


pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go
```diff
@@ -99,28 +94,26 @@ func (t *Tracker) ApplyEpoch(epoch iotago.EpochIndex, committee *account.Account
...
 		panic(ierrors.Wrapf(err, "failed to store pool stats for epoch %d", epoch))
 	}
 
-	rewardsTree := ads.NewMap(t.rewardsStorage(epoch),
-		iotago.Identifier.Bytes,
-		iotago.IdentifierFromBytes,
-		(*PoolRewards).Bytes,
-		PoolRewardsFromBytes,
-	)
+	rewardsTree, err := t.rewardsMap(epoch)
```
daria305, line 107: maybe keep using one naming convention and change tree->map

pkg/protocol/sybilprotection/sybilprotectionv1/performance/snapshot.go
```diff
@@ -271,7 +285,7 @@ func (t *Tracker) exportPoolRewards(pWriter *utils.PositionedWriter, targetEpoch
 			return ierrors.Wrapf(err, "unable to write account count for epoch index %d", epoch)
 		}
 
-		if err := rewardsTree.Stream(func(key iotago.AccountID, value *PoolRewards) error {
+		if err := rewardsTree.Stream(func(key iotago.AccountID, value *model.PoolRewards) error {
```
daria305, line 288: the same her: tree->map?

pkg/protocol/syncmanager/trivialsyncmanager/syncmanager.go
```diff
@@ -96,20 +96,20 @@ func (s *SyncManager) SyncStatus() *syncmanager.SyncStatus {
...
+	defer s.lastPrunedEpochLock.RUnlock()
 
 	return &syncmanager.SyncStatus{
 		NodeSynced:             s.isBootstrappedFunc(),
 		LastAcceptedBlockSlot:  s.lastAcceptedBlockSlot,
 		LastConfirmedBlockSlot: s.lastConfirmedBlockSlot,
 		LatestCommitment:       s.latestCommitment,
 		LatestFinalizedSlot:    s.latestFinalizedSlot,
-		LatestPrunedSlot:       s.lastPrunedSlot,
+		LatestPrunedEpoch:      s.lastPrunedEpoch,
```
daria305, line 112: heh we should align and use everywhere `latest`

jkrvivian: I prefer `Last`.  :grin:

daria305: yeah, it's shorter but isn't last the actual "last" means: no further actions, and "latest": the most up to date?

daria305 (approved):

pkg/storage/prunable/epochstore/epoch_kv.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil, ierrors.Wrapf(database.ErrEpochPruned, "epoch %d is too old", epoch)
+	}
+
+	return lo.PanicOnErr(e.kv.WithExtendedRealm(epoch.MustBytes())), nil
+}
+
+func (e *EpochKVStore) Prune(epoch iotago.EpochIndex, defaultPruningDelay iotago.EpochIndex) error {
+	// The epoch we're trying to prune already takes into account the defaultPruningDelay.
+	// Therefore, we don't need to do anything if it is greater equal e.pruningDelay and take the difference otherwise.
+	var pruningDelay iotago.EpochIndex
```
piotrm50, line 55: I don't really understand this logic - could you explain a little more?

pkg/storage/prunable/manager.go
```diff
@@ -8,41 +8,39 @@ import (
...
-	optsGranularity int64
-	optsMaxOpenDBs  int
+	optsMaxOpenDBs int
 }
 
-func NewManager(dbConfig database.Config, errorHandler func(error), opts ...options.Option[Manager]) *Manager {
-	return options.Apply(&Manager{
-		optsGranularity: 10,
-		optsMaxOpenDBs:  10,
+func NewSlotManager(dbConfig database.Config, errorHandler func(error), opts ...options.Option[SlotManager]) *SlotManager {
```
piotrm50, line 34: Why is it called `SlotManager` if it deals with `EpochIndex`?

piotrm50 (commented):
Looks good overall, added some questions in the comments. 
And my usual question is: does it run in local Docker network?

pkg/storage/prunable/manager.go
jonastheis: I renamed it to `BucketManager`

pkg/storage/prunable/epochstore/epoch_kv.go
jonastheis: This is indeed a bit tricky and took us a while to figure out properly. 

Let's assume you could prune epoch 10 (which means you have at least finalized a slot within epoch 12). Now, instead of directly pruning epoch 10, you want to keep always 5 epochs = `defaultPruningDelay`. That means you will prune until epoch 5. 
Because all of the epoch stores have their own pruning delay you need to take the max out of both. For example, for the `decidedUpgradeSignals` it's 7. So you want to prune _that_ storage until epoch `10-7=3`. The problem is, that the epoch you'll prune by default (meaning you drop the buckets and delete the spent ledger) already "contains" the `defaultPruningDelay` of `10-5=5`. That's why you can't simply do `5-7=-2->0 and don't prune` but instead need to take the difference.

piotrm50 (approved):

PR #289, opened 2023-08-16
Cleanup and refactor block issuing

piotrm50:
Separate options for basic blocks and validator blocks.
Each block-issuing component manages its own issuer key and account - this enables more flexibility as one node can use different identity to issue blocks from API and different keys to issue validator blocks etc.


karimodm (approved):
Looks good!

PR #293, opened 2023-08-18
Feat: Reactive protocol rework

### LINKED ISSUE #447: nilpointer in the testsuite network
When running unit tests with `-race` flag, the following test occurred: 

```
2023-10-19T11:51:25.6780842Z node7 > Network.BlockReceived: from 12D3KooWG54cq6Xy8g6t83mBXtgu59FnarLXkQ8iySkkTVwxGSD3 BlockID(node2-activity.9:22) - 22
2023-10-19T11:51:25.6783677Z panic: runtime error: invalid memory address or nil pointer dereference
2023-10-19T11:51:25.6785715Z [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x112c7a3]
2023-10-19T11:51:25.6787080Z 
2023-10-19T11:51:25.6787399Z goroutine 248556 [running]:
2023-10-19T11:51:25.6788993Z github.com/iotaledger/iota-core/pkg/testsuite/mock.(*Endpoint).Send.func1()
2023-10-19T11:51:25.6792232Z 	/actions-runner/_work/iota-core/iota-core/pkg/testsuite/mock/network.go:156 +0x143
2023-10-19T11:51:25.6795385Z created by github.com/iotaledger/iota-core/pkg/testsuite/mock.(*Endpoint).Send in goroutine 229230
2023-10-19T11:51:25.6798144Z 	/actions-runner/_work/iota-core/iota-core/pkg/testsuite/mock/network.go:152 +0x63a
2023-10-19T11:51:25.6800244Z FAIL	github.com/iotaledger/iota-core/pkg/tests	182.130s
2023-10-19T11:51:25.6801404Z FAIL
2023-10-19T11:51:25.6828322Z ##[error]Process completed with exit code 1.
2023-10-19T11:51:25.7200099Z Post job cleanup.
```

The following code looks suspicious:

Image

hmoog:
Closes #447


tools/evil-spammer/evil-spammer
jonastheis, line 1: This has been moved to https://github.com/iotaledger/evil-tools/. Just delete everything in `tools/evil-spammer`.

pkg/testsuite/storage_settings.go
```diff
@@ -123,8 +125,11 @@ func (t *TestSuite) AssertChainID(expectedChainID iotago.CommitmentID, nodes ...
 
 	for _, node := range nodes {
 		t.Eventually(func() error {
-			actualChainID := node.Protocol.MainEngineInstance().ChainID()
-			if expectedChainID != node.Protocol.MainEngineInstance().ChainID() {
+			actualChainID := node.Protocol.Chains.Main.Get().ForkingPoint.Get().ID()
+
+			if expectedChainID != actualChainID {
+				fmt.Println(expectedChainID, actualChainID)
```
jonastheis: remove print

pkg/testsuite/mock/network.go
```diff
@@ -153,8 +156,10 @@ func (e *Endpoint) Send(packet proto.Message, to ...peer.ID) {
 			e.network.dispatchersMutex.RLock()
 			defer e.network.dispatchersMutex.RUnlock()
 
-			if err := dispatcher.handler(e.id, packet); err != nil {
-				fmt.Println(e.id, "ERROR: ", err)
+			if dispatcher.handler != nil {
```
jonastheis, line 159: When can this be nil? Would be good to add a comment here. And in case this is an error, also print something or so. So we can at least see it.

pkg/testsuite/mock/node.go
```diff
@@ -165,78 +182,12 @@ func (n *Node) hookEvents() {
...
-
-	events.Network.Error.Hook(func(err error, id peer.ID) {
-		fmt.Printf("%s > Network.Error: from %s %s\n", n.Name, id, err)
-	})
-
-	events.Error.Hook(func(err error) {
-		fmt.Printf("%s > Protocol.Error: %s\n", n.Name, err.Error())
+	n.Protocol.Chains.WithElements(func(chain *protocol.Chain) (teardown func()) {
+		return chain.Engine.OnUpdate(func(_ *engine.Engine, newEngine *engine.Engine) {
+			if newEngine != nil {
```
jonastheis, line 220: When could `newEngine` be nil?

pkg/protocol/chains.go
```diff
@@ -0,0 +1,180 @@
...
+// Chains is a subcomponent of the protocol that exposes the chains that are managed by the protocol and that implements
+// the chain switching logic.
+type Chains struct {
+	// Set contains all non-evicted chains that are managed by the protocol.
+	reactive.Set[*Chain]
+
+	// Main contains the main chain.
+	Main reactive.Variable[*Chain]
+
+	// HeaviestClaimedCandidate contains the candidate chain with the heaviest claimed weight.
```
jonastheis: These comments are somewhat useless: you are describing it with the same words. It would be better to describe it differently: 
```suggestion
	// HeaviestClaimedCandidate contains the candidate chain with the heaviest claimed weight according to its latest commitment. The weight has neither been checked via attestations nor verified by downloading all data.
```

pkg/protocol/chains.go
```diff
@@ -0,0 +1,180 @@
...
+	HeaviestClaimedCandidate reactive.Variable[*Chain]
+
+	// HeaviestAttestedCandidate contains the candidate chain with the heaviest attested weight.
```
jonastheis:
```suggestion
	// HeaviestAttestedCandidate contains the candidate chain with the heaviest weight as checked by attestations. The chain has not been instantiated into an engine yet.
```

pkg/protocol/chains.go
```diff
@@ -0,0 +1,180 @@
...
+	HeaviestAttestedCandidate reactive.Variable[*Chain]
+
+	// HeaviestVerifiedCandidate contains the candidate chain with the heaviest verified weight.
```
jonastheis:
```suggestion
	// HeaviestVerifiedCandidate contains the candidate chain with the heaviest verified weight, meaning the chain has been instantiated into an engine and the commitments have been produced by the engine itself.
```

pkg/testsuite/mock/node.go
```diff
@@ -148,13 +157,21 @@ func (n *Node) Initialize(failOnBlockFiltered bool, opts ...options.Option[proto
...
-
-	events.ChainManager.ForkDetected.Hook(func(fork *chainmanager.Fork) { n.forkDetectedCount.Add(1) })
+	n.Protocol.Chains.HeaviestAttestedCandidate.OnUpdate(func(_ *protocol.Chain, heaviestAttestedCandidate *protocol.Chain) {
+		if heaviestAttestedCandidate != nil {
+			n.forkDetectedCount.Add(1)
 
-	events.CandidateEngineActivated.Hook(func(e *engine.Engine) { n.candidateEngineActivatedCount.Add(1) })
+			heaviestAttestedCandidate.Engine.OnUpdate(func(prevEngine *engine.Engine, newEngine *engine.Engine) {
+				n.candidateEngineActivatedCount.Add(1)
+			})
```
jonastheis, line 172: Just for my understanding: `heaviestAttestedCandidate.Engine` is only set when we start syncing this candidate chain? Before it is nil?

pkg/protocol/protocol_attestations.go
```diff
@@ -0,0 +1,260 @@
...
+	// workerPool contains the worker pool that is used to process attestation requests and responses asynchronously.
+	workerPool *workerpool.WorkerPool
+
+	// ticker contains the ticker that is used to send attestation requests.
+	ticker *eventticker.EventTicker[iotago.SlotIndex, iotago.CommitmentID]
+
+	// commitmentVerifiers contains the commitment verifiers that are used to verify received attestations.
+	commitmentVerifiers *shrinkingmap.ShrinkingMap[iotago.CommitmentID, *CommitmentVerifier]
+
+	// Logger embeds a logger that can be used to log messages emitted by this chain.
```
jonastheis: wrong comment

pkg/protocol/protocol_blocks.go
```diff
@@ -0,0 +1,162 @@
...
+		})
+
+		protocol.Chains.WithElements(func(chain *Chain) func() {
+			return chain.Engine.WithNonEmptyValue(func(engineInstance *engine.Engine) (shutdown func()) {
+				return engineInstance.Events.BlockRequester.Tick.Hook(b.SendRequest).Unhook
+			})
+		})
+
+		protocol.Chains.Main.Get().Engine.OnUpdateWithContext(func(_ *engine.Engine, engine *engine.Engine, unsubscribeOnEngineChange func(subscriptionFactory func() (unsubscribe func()))) {
+			if engine != nil {
```
jonastheis: How could `engine` be nil here?

pkg/protocol/protocol_blocks.go
```diff
@@ -0,0 +1,162 @@
...
+				unsubscribeOnEngineChange(func() (unsubscribe func()) {
+					return lo.Batch(
+						engine.Events.Scheduler.BlockScheduled.Hook(func(block *blocks.Block) { b.SendResponse(block.ModelBlock()) }).Unhook,
+						engine.Events.Scheduler.BlockSkipped.Hook(func(block *blocks.Block) { b.SendResponse(block.ModelBlock()) }).Unhook,
```
jonastheis: Maybe @cyberphysic4l can shed some light here: do we gossip skipped blocks? I mean it should not hurt but aren't blocks only skipped when accepted? That means it should be basically useless to gossip it

pkg/protocol/protocol.go
```diff
@@ -2,316 +2,216 @@ package protocol
...
-	p.runNetworkProtocol()
-
-	p.TriggerInitialized()
+	// AttestationsProtocol contains the subcomponent that is responsible for handling attestation requests and
+	// responses.
+	AttestationsProtocol *AttestationsProtocol
 
-	<-p.context.Done()
+	// WarpSyncProtocol contains the subcomponent that is responsible for handling warp sync requests and responses.
+	WarpSyncProtocol *WarpSyncProtocol
```
jonastheis: Why are these public? As far as I can see they don't need to be. I think it would be better to keep the API minimal

pkg/protocol/protocol_commitments.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	})
+}
+
+// ProcessRequest processes the given commitment request.
+func (c *CommitmentsProtocol) ProcessRequest(commitmentID iotago.CommitmentID, from peer.ID) {
+	c.workerPool.Submit(func() {
+		commitment, err := c.protocol.Commitments.Get(commitmentID)
+		if err != nil {
+			logLevel := lo.Cond(ierrors.Is(err, ErrorCommitmentNotFound), log.LevelTrace, log.LevelError)
```
jonastheis, line 97: Seems odd to have different log levels but the same message. Just make it an if and adjust the message accordingly. Or since it is not a critical error, maybe it should always be a trace

pkg/protocol/protocol_commitments.go
```diff
@@ -0,0 +1,104 @@
...
+		} else if published {
+			c.LogTrace("received response", "commitment", commitment.LogName(), "fromPeer", from)
+		}
+	})
+}
+
+// ProcessRequest processes the given commitment request.
+func (c *CommitmentsProtocol) ProcessRequest(commitmentID iotago.CommitmentID, from peer.ID) {
+	c.workerPool.Submit(func() {
+		commitment, err := c.protocol.Commitments.Get(commitmentID)
```
jonastheis, line 95: We load commitments here from the cache right? What does that mean if someone wants to solidify a chain back to genesis? Does that even make sense for someone to do?

pkg/protocol/protocol_warp_sync.go
```diff
@@ -0,0 +1,321 @@
...
+// WarpSyncProtocol is a subcomponent of the protocol that is responsible for handling warp sync requests and responses.
+type WarpSyncProtocol struct {
+	// protocol contains a reference to the Protocol instance that this component belongs to.
+	protocol *Protocol
+
+	// workerPool contains the worker pool that is used to process warp sync requests and responses asynchronously.
+	workerPool *workerpool.WorkerPool
+
+	// ticker contains the ticker that is used to send warp sync requests.
+	ticker *eventticker.EventTicker[iotago.SlotIndex, iotago.CommitmentID]
```
jonastheis, line 31: Why no options for this ticker?

pkg/protocol/protocol_warp_sync.go
```diff
@@ -0,0 +1,321 @@
...
+// newWarpSyncProtocol creates a new warp sync protocol instance for the given protocol.
+func newWarpSyncProtocol(protocol *Protocol) *WarpSyncProtocol {
+	c := &WarpSyncProtocol{
+		Logger:     lo.Return1(protocol.Logger.NewChildLogger("WarpSync")),
+		protocol:   protocol,
+		workerPool: protocol.Workers.CreatePool("WarpSync", workerpool.WithWorkerCount(1)),
+		ticker:     eventticker.New[iotago.SlotIndex, iotago.CommitmentID](),
+	}
+
+	c.ticker.Events.Tick.Hook(c.SendRequest)
```
jonastheis, line 46: This is not really related to the changes of this PR but just comes to mind when reviewing the usage of the ticker: the ticker API is bad, where the interaction happens via the `Tick` event. Thus, you always hook to that event. I think it would be much better design (don't need to do it now) to make the ticker receive a function with the type that gets executed on each tick.

pkg/protocol/protocol_warp_sync.go
```diff
@@ -0,0 +1,321 @@
...
+	return c
+}
+
+// SendRequest sends a warp sync request for the given commitment ID to all peers.
+func (w *WarpSyncProtocol) SendRequest(commitmentID iotago.CommitmentID) {
+	w.workerPool.Submit(func() {
+		if commitment, err := w.protocol.Commitments.Get(commitmentID, false); err == nil {
+			w.protocol.Network.SendWarpSyncRequest(commitmentID)
+
+			w.LogDebug("request", "commitment", commitment.LogName())
```
jonastheis, line 78: I don't really get this style of logging: you do not capture every code path and hence couldn't tell whether this was actually called or not. For now this is better than what we had before (namely no logging). But ultimately we should come up with some form of structured logging and then apply some sort of decorator that logs time entered, time exited, and function names etc

pkg/protocol/protocol_warp_sync.go
```diff
@@ -0,0 +1,321 @@
...
+			w.LogDebug("failed to get target engine for response", "commitment", commitment.LogName())
+
+			return
+		}
+
+		commitment.BlocksToWarpSync.Compute(func(blocksToWarpSync ds.Set[iotago.BlockID]) ds.Set[iotago.BlockID] {
+			if blocksToWarpSync != nil || !commitment.WarpSyncBlocks.Get() {
+				w.LogTrace("response for already synced commitment", "commitment", commitment.LogName(), "fromPeer", from)
+
+				return blocksToWarpSync
```
jonastheis, line 130: Is `blocksToWarpSync` nil here? Returning the parameter in an error case makes this very hard to reason about. Would be better to return nil if possible

pkg/protocol/protocol_warp_sync.go
```diff
@@ -0,0 +1,321 @@
...
+				w.LogError("failed to verify mutations proof", "commitment", commitment.LogName(), "transactionIDs", transactionIDs, "proof", mutationProof, "fromPeer", from)
+
+				return blocksToWarpSync
+			}
+
+			w.ticker.StopTicker(commitmentID)
+
+			targetEngine.Workers.WaitChildren()
+
+			if !chain.WarpSyncMode.Get() {
```
jonastheis, line 164: Why check this again here? Few lines above this is already checked.

pkg/protocol/protocol_warp_sync.go
```diff
@@ -0,0 +1,321 @@
...
+				w.LogTrace("response for chain without warp-sync", "chain", chain.LogName(), "fromPeer", from)
+
+				return blocksToWarpSync
+			}
+
+			// make sure the engine is clean and requires a warp-sync before we start processing the blocks
+			if targetEngine.Workers.WaitChildren(); targetEngine.Storage.Settings().LatestCommitment().ID().Slot() > commitmentID.Slot() {
```
jonastheis: `targetEngine.Workers.WaitChildren()` is also executed a few lines above

jonastheis: Shouldn't this be `targetEngine.Storage.Settings().LatestCommitment().ID().Slot() >= commitmentID.Slot()`? We don't want to warpsync if we have a commitment that is greater or equal the one that we currently looking at no?

pkg/protocol/protocol_warp_sync.go
```diff
@@ -0,0 +1,321 @@
...
+					w.protocol.LogError("commitment does not match", "expectedCommitmentID", commitmentID, "producedCommitmentID", producedCommitment.ID())
+
+					return
+				}
+			}
+
+			if totalBlocks == 0 {
+				forceCommitmentFunc()
+
+				return blocksToWarpSync
```
jonastheis: I _really_ don't like using this `blocksToWarpSync` as a default return parameter. Imo here it should be an empty set.

pkg/protocol/protocol_attestations.go
```diff
@@ -0,0 +1,260 @@
...
+	log.Logger
+}
+
+// newAttestationsProtocol creates a new attestation protocol instance for the given protocol.
+func newAttestationsProtocol(protocol *Protocol) *AttestationsProtocol {
+	a := &AttestationsProtocol{
+		Logger:              lo.Return1(protocol.Logger.NewChildLogger("Attestations")),
+		protocol:            protocol,
+		workerPool:          protocol.Workers.CreatePool("Attestations"),
+		ticker:              eventticker.New[iotago.SlotIndex, iotago.CommitmentID](),
```
jonastheis: Why no options for this ticker?

pkg/protocol/chains.go
```diff
@@ -0,0 +1,180 @@
...
+	HeaviestVerifiedCandidate reactive.Variable[*Chain]
```
jonastheis, line 32: Maybe rename to `HeaviestProducedCandidate` or something to be more in line with `Chain.LatestProducedCommitment`?

pkg/protocol/protocol_attestations.go
```diff
@@ -0,0 +1,260 @@
...
+		commitmentVerifiers: shrinkingmap.New[iotago.CommitmentID, *CommitmentVerifier](),
+	}
+
+	a.ticker.Events.Tick.Hook(a.sendRequest)
+
+	protocol.Constructed.OnTrigger(func() {
+		protocol.Chains.WithElements(func(chain *Chain) (shutdown func()) {
+			return chain.RequestAttestations.WithNonEmptyValue(func(requestAttestations bool) (shutdown func()) {
+				return a.setupCommitmentVerifier(chain)
```
jonastheis, line 52: When exactly is this teardown func here called? When is the commitment verifier for the forking point deleted out of the map?

pkg/protocol/protocol_attestations.go
```diff
@@ -0,0 +1,260 @@
...
+			a.LogDebug("failed to publish commitment when processing attestations", "commitmentID", commitmentModel.ID(), "peer", from, "error", err)
+
+			return
+		}
+
+		if commitment.AttestedWeight.Compute(func(currentWeight uint64) uint64 {
+			if !commitment.RequestAttestations.Get() {
+				a.LogTrace("received attestations for previously attested commitment", "commitment", commitment.LogName())
+
+				return currentWeight
```
jonastheis, line 88: Why return `currentWeight` in the error cases? Wouldn't it be better to atomically set `IsAttested` and not even start `AttestedWeight.Compute` if it's already attested?

pkg/protocol/commitment_verifier.go
```diff
@@ -32,68 +36,38 @@ func NewCommitmentVerifier(mainEngine *engine.Engine, lastCommonCommitmentBefore
...
-	tree := ads.NewMap[iotago.Identifier](mapdb.NewMapDB(),
-		iotago.Identifier.Bytes,
-		iotago.IdentifierFromBytes,
-		iotago.AccountID.Bytes,
-		iotago.AccountIDFromBytes,
-		(*iotago.Attestation).Bytes,
-		iotago.AttestationFromBytes(c.engine),
-	)
+// verifyCommitment verifies the given commitment and returns the blockIDs and cumulative weight of the attestations.
+func (c *CommitmentVerifier) verifyCommitment(commitment *Commitment, attestations []*iotago.Attestation, merkleProof *merklehasher.Proof[iotago.Identifier]) (blockIDsFromAttestations iotago.BlockIDs, cumulativeWeight uint64, err error) {
```
jonastheis, line 58: Please add the comments back in this function. They are critical to understand the behavior especially in certain edge cases and serve as documentation within the code. I don't understand why you removed them in first place...

pkg/protocol/commitment_verifier.go
```diff
@@ -32,68 +36,38 @@ func NewCommitmentVerifier(mainEngine *engine.Engine, lastCommonCommitmentBefore
...
+	tree := ads.NewMap[iotago.Identifier](mapdb.NewMapDB(), iotago.Identifier.Bytes, iotago.IdentifierFromBytes, iotago.AccountID.Bytes, iotago.AccountIDFromBytes, (*iotago.Attestation).Bytes, iotago.AttestationFromBytes(c.engine))
```
jonastheis: This is less readable than having it with new lines. I know it saves some lines of code but what does that help if you can't read or modify the shit

pkg/protocol/commitment_verifier.go
```diff
@@ -32,68 +36,38 @@ func NewCommitmentVerifier(mainEngine *engine.Engine, lastCommonCommitmentBefore
...
 
 	for _, att := range attestations {
-		if err := tree.Set(att.Header.IssuerID, att); err != nil {
+		if setErr := tree.Set(att.Header.IssuerID, att); setErr != nil {
```
jonastheis: you don't actually use this error for the error message

pkg/protocol/protocol.go
```diff
@@ -2,316 +2,216 @@ package protocol
...
+		p.waitMainEngineInitialized()
+	})
 }
 
-func (p *Protocol) initChainManager() {
-	p.ChainManager = chainmanager.NewManager(p, p.HandleError, p.optsChainManagerOptions...)
-	p.Events.ChainManager.LinkTo(p.ChainManager.Events)
+// IssueBlock issues a block to the node.
+func (p *Protocol) IssueBlock(block *model.Block) error {
+	p.Network.Events.BlockReceived.Trigger(block, "self")
```
jonastheis, line 98: Why not simply calling `p.BlocksProtocol.ProcessResponse`?

pkg/protocol/options.go
```diff
@@ -1,170 +1,289 @@
...
+type Options struct {
+	// BaseDirectory is the directory where the protocol will store its data.
+	BaseDirectory string
+
+	// SnapshotPath is the path to the snapshot file that should be used to initialize the protocol.
+	SnapshotPath string
+
+	// ChainSwitchingThreshold is the threshold that defines how far away a heavier chain needs to be from its forking
+	// point to be considered for switching.
+	ChainSwitchingThreshold iotago.SlotIndex
```
jonastheis, line 56: Should this be a protocol parameter?

pkg/protocol/engine/engine.go
```diff
@@ -556,12 +532,94 @@ func (e *Engine) ErrorHandler(componentName string) func(error) {
...
+	logger, stopLogging := logger.NewEntityLogger("Engine")
+	reactiveModule = module.NewReactiveModule(logger)
+
+	e.RootCommitment.LogUpdates(reactiveModule, log.LevelTrace, "RootCommitment")
+	e.LatestCommitment.LogUpdates(reactiveModule, log.LevelTrace, "LatestCommitment")
+
+	reactiveModule.Shutdown.OnTrigger(func() {
+		reactiveModule.LogDebug("shutting down")
+
+		stopLogging()
```
jonastheis: shouldn't that be called at the very end? otherwise you won't be able to get logs while shutting down. 

and there's even a line `reactiveModule.LogDebug("stopped")` further below

pkg/protocol/engine/engine.go
```diff
@@ -556,12 +532,94 @@ func (e *Engine) ErrorHandler(componentName string) func(error) {
 	}
 }
 
-func (e *Engine) RLock() {
-	e.accessMutex.RLock()
+func (e *Engine) initRootCommitment() {
```
jonastheis, line 515: the function `EarliestRootCommitment` can be removed as this functionality is covered in here. Please also take the comment and put it here

pkg/protocol/engine/engine.go
```diff
@@ -556,12 +532,94 @@ func (e *Engine) ErrorHandler(componentName string) func(error) {
...
+			targetSlot := protocolParams.GenesisSlot()
+			if lastFinalizedSlot > targetSlot+maxCommittableAge {
+				targetSlot = lastFinalizedSlot - maxCommittableAge
+			}
+
+			if rootCommitment != nil && targetSlot == rootCommitment.Slot() {
+				return rootCommitment
+			}
+
+			return lo.Return1(e.Storage.Commitments().Load(targetSlot))
```
jonastheis: Why are you ignoring the error here? In principle this should never happen, but we should still at least log the error

pkg/protocol/engine/engine.go
```diff
@@ -556,12 +532,94 @@ func (e *Engine) ErrorHandler(componentName string) func(error) {
...
+		})
+
+		e.Shutdown.OnTrigger(unsubscribe)
+	})
+}
+
+func (e *Engine) initLatestCommitment() {
+	updateLatestCommitment := func(latestCommitment *model.Commitment) {
+		e.LatestCommitment.Compute(func(currentLatestComponent *model.Commitment) *model.Commitment {
+			return lo.Cond(currentLatestComponent == nil || currentLatestComponent.Slot() < latestCommitment.Slot(), latestCommitment, currentLatestComponent)
```
jonastheis: `currentLatestComponent` -> `currentLatestCommitment`?

pkg/testsuite/mock/network.go
hmoog: It is not an error. We set the handler to nil upon calling "UnregisterProtocol" which is currently called when shutting down.

In theory, nothing should arrive at the network level post shutdown as we should kill all the background processes and wait for all the workers to finish their jobs but since some background workers (i.e. the scheduler) are using goroutines, we can not reliably wait for these things and in practice it can happen that the network sees an attempt to send something post shutdown (this mostly happened in artificial tests that just called defer Shutdown on the testframework, which means that we dont even have a real synchronized shutdown logic).

In general, since the mocked network is somewhat independent of the implementations that build on top of this, I think it makes sense to gracefully handle the case of an unregistered protocol and simply handle it as a noop and prevent nil-pointer panics from happening.

I am not sure if it makes sense to add any logging here as we are talking about a case that happens post shutdown and I am not sure if it makes sense to print things after the test already finished (defer).

pkg/testsuite/mock/node.go
hmoog: When the chain shuts down the engine (for example if it is no longer the heaviest candidate or it gets evicted / shutdown otherwise) but even though eviction is currently disabled, the main way to address the reactive pattern is to have a well defined behavior for each state transition / mini-rule and if you are able to handle an edge-case gracefully then you should.

We can ultimately reason about some possible edge-case related checks to be unnecessary in the greater scheme of things but I think it makes sense to first have a system that tries to be on the safe side, by having possibly unnecessary nil checks to prevent unintended panics.

But generally speaking, we should have a discussion about our eviction strategy. Do we want to reset all variables to their initial empty state upon shutdown (in which case things can get nil) or do we want to keep the state "as is" to for example be able to keep it inspectable even after shutdown and rely on the fact, that once we get rid of all references to an entity, the garbage collection will take care of the rest.

In the latter case, we could make some additional assumpttions about things never getting nil after they have first been set and we can also simplify some of our shutdown related logic.

pkg/protocol/engines.go
```diff
@@ -0,0 +1,261 @@
...
+		e.Initialized.Trigger()
+	})
+
+	e.Constructed.Trigger()
+
+	return e
+}
+
+// LoadMainEngine loads the main engine from disk or creates a new one if no engine exists.
+func (e *Engines) LoadMainEngine(snapshotPath string) (*engine.Engine, error) {
```
jonastheis: Some of the methods are public and others are private. Why?

pkg/protocol/engines.go
```diff
@@ -0,0 +1,261 @@
...
+	return e.protocol.Chains.Main.WithNonEmptyValue(func(mainChain *Chain) (shutdown func()) {
+		return e.Main.DeriveValueFrom(reactive.NewDerivedVariable(func(currentMainEngine *engine.Engine, newMainEngine *engine.Engine) *engine.Engine {
+			return lo.Cond(newMainEngine == nil, currentMainEngine, newMainEngine)
+		}, mainChain.Engine))
+	})
+}
+
+// syncMainEngineInfoFile syncs the engine info file with the main engine.
+func (e *Engines) syncMainEngineInfoFile() (shutdown func()) {
+	return e.Main.OnUpdate(func(_ *engine.Engine, mainEngine *engine.Engine) {
```
jonastheis, line 233: Why not `e.Main.WithNonEmptyValue` ?

pkg/protocol/engines.go
```diff
@@ -0,0 +1,261 @@
...
+			}
+		}
+	})
+}
+
+// injectEngineInstances injects engine instances into the chains (when requested).
+func (e *Engines) injectEngineInstances() (shutdown func()) {
+	return e.protocol.Chains.WithElements(func(chain *Chain) (shutdown func()) {
+		return chain.StartEngine.OnUpdate(func(_ bool, startEngine bool) {
+			e.worker.Submit(func() {
```
jonastheis, line 246: This can't lead to race conditions because there is only a single worker in `e.worker` right?

pkg/protocol/engines.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	// load new engine if no previous engine exists.
+	e.Main.Compute(func(mainEngine *engine.Engine) *engine.Engine {
+		if mainEngine != nil {
+			return mainEngine
+		}
+
+		return e.loadEngineInstanceFromSnapshot(lo.PanicOnErr(uuid.NewUUID()).String(), snapshotPath)
+	})
```
jonastheis: Why isn't this entire block part of `e.Main.Compute`? Above you use set and then compute and even check the value again if it is nil. Why is that?

pkg/protocol/engines.go
```diff
@@ -0,0 +1,261 @@
...
+	return e.directory.Path(engineInfoFile)
+}
+
+// loadEngineInstanceFromSnapshot loads an engine instance from a snapshot.
+func (e *Engines) loadEngineInstanceFromSnapshot(engineAlias string, snapshotPath string) *engine.Engine {
+	errorHandler := func(err error) {
+		e.protocol.LogError("engine error", "err", err, "name", engineAlias[0:8])
+	}
+
+	e.protocol.Options.EngineOptions = append(e.protocol.Options.EngineOptions, engine.WithSnapshotPath(snapshotPath))
```
jonastheis: This should be done only once in the constructor as you modify a property that is reused also somewhere else

pkg/protocol/engines.go
```diff
@@ -0,0 +1,261 @@
...
+
+	return e.loadEngineInstanceWithStorage(engineAlias, storage.Create(e.directory.Path(engineAlias), DatabaseVersion, errorHandler, e.protocol.Options.StorageOptions...))
+}
+
+// loadEngineInstanceWithStorage loads an engine instance with the given storage.
+func (e *Engines) loadEngineInstanceWithStorage(engineAlias string, storage *storage.Storage) *engine.Engine {
```
jonastheis: please split up in multiple lines. it's unreadable and hard to edit if the line is 50m long and you need to scroll to the right

pkg/protocol/protocol.go
```diff
@@ -2,316 +2,216 @@ package protocol
...
+		p.Workers.Shutdown()
+	}
 }
 
-func (p *Protocol) APIForVersion(version iotago.Version) (iotago.API, error) {
-	return p.MainEngineInstance().APIForVersion(version)
+// initEviction initializes the eviction of old data when the engine advances and returns a function that shuts it down.
+func (p *Protocol) initEviction() (shutdown func()) {
+	return p.Commitments.Root.OnUpdate(func(_ *Commitment, rootCommitment *Commitment) {
+		// TODO: DECIDE ON DATA AVAILABILITY TIMESPAN / EVICTION STRATEGY
```
jonastheis, line 175: This eviction is only for the protocol and its components like commitments and chains etc?

pkg/protocol/chainmanager/manager.go
```diff
@@ -1,524 +0,0 @@
...
-
-	return m.switchMainChainToCommitment(commitment)
-}
-
-func (m *Manager) processCommitment(commitment *model.Commitment) (wasForked bool, isSolid bool, chainCommitment *ChainCommitment, err error) {
-	// Verify the commitment's version corresponds to the protocol version for the slot.
-	apiForSlot := m.apiProvider.APIForSlot(commitment.Slot())
-	if apiForSlot.Version() != commitment.Commitment().ProtocolVersion {
-		return false, false, nil, ierrors.Errorf("")
-	}
```
jonastheis, line 258: Are we doing this check somewhere?

pkg/protocol/commitment.go
```diff
@@ -0,0 +1,285 @@
...
+		// mark commitments that are marked as root as verified
+		c.IsVerified.InheritFrom(c.IsRoot),
+
+		// mark commitments that are marked as verified as attested
+		c.IsAttested.InheritFrom(c.IsVerified),
+
+		c.Parent.WithNonEmptyValue(func(parent *Commitment) func() {
+			// the weight can be fixed as a one time operation (as it only relies on static information from the parent
+			// commitment)
+			c.Weight.Set(c.CumulativeWeight() - parent.CumulativeWeight())
```
jonastheis: What happens if someone sends me a "malicious" commitment which has a lower weight than its parent? this would underflow no?

pkg/protocol/commitment.go
```diff
@@ -0,0 +1,285 @@
...
+		return child
+	})
+
+	return func() {
+		c.MainChild.Compute(func(mainChild *Commitment) *Commitment {
+			if !c.Children.Delete(child) || child != mainChild {
+				return mainChild
+			}
+
+			return lo.Return1(c.Children.Any())
```
jonastheis, line 218: Why can we just take any child here? Shouldn't it be the heaviest?

pkg/protocol/commitment.go
```diff
@@ -0,0 +1,285 @@
...
+// engine, is no longer requesting warp sync, and we are above the latest verified Commitment.
+func (c *Commitment) deriveReplayDroppedBlocks(chain *Chain) func() {
+	return c.ReplayDroppedBlocks.DeriveValueFrom(reactive.NewDerivedVariable3(func(_ bool, engineInstance *engine.Engine, warpSyncing bool, isAboveLatestVerifiedCommitment bool) bool {
+		return engineInstance != nil && !warpSyncing && isAboveLatestVerifiedCommitment
+	}, chain.Engine, chain.WarpSyncMode, c.IsAboveLatestVerifiedCommitment))
+}
+
+// forceChain forces the Chain of this Commitment to the given Chain by promoting it to the main child of its parent if
+// the parent is on the target Chain.
+func (c *Commitment) forceChain(targetChain *Chain) {
```
jonastheis, line 295: Should we have a log in here? Might be interesting to know when a chain was forced upon a commitment

pkg/protocol/commitment.go
```diff
@@ -0,0 +1,285 @@
...
+	if currentChain := c.Chain.Get(); currentChain != targetChain {
+		if parent := c.Parent.Get(); parent.Chain.Get() == targetChain {
+			parent.MainChild.Set(c)
```
jonastheis, line 298: Just for my understanding: once we set the `MainChild` it will trigger its updates and via the setup in `deriveChain` we will evict the `targetChain`?

pkg/protocol/commitment.go
```diff
@@ -0,0 +1,285 @@
...
+// commitments produced by the nodes in the network.
+type Commitment struct {
+	// Commitment contains the underlying model.Commitment that is represented by this Commitment.
+	*model.Commitment
+
+	// Parent contains the Commitment that is referenced as a parent in this Commitment.
+	Parent reactive.Variable[*Commitment]
+
+	// Children contains the Commitments that reference this Commitment as a parent.
+	Children reactive.Set[*Commitment]
```
jonastheis, line 25: How does this behave with eviction? Can the GC clean this up? Doesn't it keeps pointers in both directions?

pkg/protocol/commitments.go
```diff
@@ -0,0 +1,244 @@
...
+			return
+		}
+
+		newRootCommitment.IsRoot.Set(true)
+		if published {
+			newRootCommitment.Chain.Set(mainChain)
+		}
+
+		// TODO: USE SET HERE (debug eviction issues)
+		mainChain.ForkingPoint.DefaultTo(newRootCommitment)
```
jonastheis, line 118: Why do we keep moving the `ForkingPoint` forward?

pkg/protocol/commitments.go
```diff
@@ -0,0 +1,244 @@
...
+			publishedCommitment, _, err := c.publishCommitmentModel(modelToPublish)
+			if err != nil {
+				c.LogError("failed to publish commitment from engine", "engine", engine.LogName(), "commitment", modelToPublish, "err", err)
+
+				return
+			}
+
+			// mark it as produced by ourselves and force it to be on the right chain (in case our chain produced a
+			// different commitment than the one we erroneously expected it to be - we always trust our engine most).
+			publishedCommitment.AttestedWeight.Set(publishedCommitment.Weight.Get())
```
jonastheis, line 157: Would be good to add another comment here why we can set `Weight == AttestedWeight`. As it doesn't need to be the same

pkg/protocol/commitments.go
```diff
@@ -0,0 +1,244 @@
...
+		// TODO: USE SET HERE (debug eviction issues)
+		mainChain.ForkingPoint.DefaultTo(newRootCommitment)
+
+		c.Root.Set(newRootCommitment)
+	})
+}
+
+// publishEngineCommitments publishes the commitments of the given engine to its chain.
+func (c *Commitments) publishEngineCommitments(chain *Chain, engine *engine.Engine) (shutdown func()) {
+	latestPublishedSlot := chain.LastCommonSlot()
```
jonastheis, line 126: Seems odd to capture this variable here

pkg/protocol/chains.go
```diff
@@ -0,0 +1,180 @@
...
+// deriveLatestSeenSlot derives the latest seen slot from the protocol.
+func (c *Chains) deriveLatestSeenSlot(protocol *Protocol) func() {
+	return protocol.Engines.Main.WithNonEmptyValue(func(mainEngine *engine.Engine) (shutdown func()) {
+		return lo.Batch(
+			mainEngine.Initialized.OnTrigger(func() {
+				c.LatestSeenSlot.Set(mainEngine.LatestCommitment.Get().Slot())
+			}),
+
+			protocol.Network.OnBlockReceived(func(block *model.Block, src peer.ID) {
+				c.LatestSeenSlot.Set(mainEngine.LatestAPI().TimeProvider().SlotFromTime(block.ProtocolBlock().Header.IssuingTime))
```
jonastheis: Why so complicated? And not just `c.LatestSeenSlot.Set(block.ID().Slot())`?

pkg/protocol/chains.go
```diff
@@ -0,0 +1,180 @@
...
+		c.HeaviestAttestedCandidate.WithNonEmptyValue(func(heaviestAttestedCandidate *Chain) (shutdown func()) {
+			return heaviestAttestedCandidate.StartEngine.ToggleValue(true)
+		}),
+
+		c.HeaviestVerifiedCandidate.WithNonEmptyValue(func(heaviestVerifiedCandidate *Chain) (shutdown func()) {
+			// only switch to the heaviest chain if the latest produced commitment is enough slots away from the forking point.
+			chainSwitchingCondition := func(_ *Commitment, latestProducedCommitment *Commitment) bool {
+				forkingPoint := heaviestVerifiedCandidate.ForkingPoint.Get()
+
+				return forkingPoint != nil && latestProducedCommitment != nil && (latestProducedCommitment.ID().Slot()-forkingPoint.ID().Slot()) > chainSwitchingThreshold
```
jonastheis: This is not exactly like the fork choice rule was implemented before. Right now you're checking whether the candidate chain is heavier for x slots when producing the commitments ourselves from that candidate engine. However, you instantiate a candidate engine immediately after another attested chain becomes heavier. Before, a candidate engine would only be instantiated when x attested slots in a row were heavier. So I think we should add this check additionally to `HeaviestAttestedCandidate` in order to make sure that we're not tricked into spawning a new engine

pkg/protocol/chain.go
```diff
@@ -0,0 +1,400 @@
...
+
+		return 0
+	}, latestSeenSlot))
+}
+
+// addCommitment adds the given commitment to this chain.
+func (c *Chain) addCommitment(newCommitment *Commitment) (shutdown func()) {
+	c.commitments.Set(newCommitment.Slot(), newCommitment)
+
+	c.LatestCommitment.Set(newCommitment)
```
jonastheis, line 312: How do we make sure that this commitment is really the latest for this chain?

pkg/protocol/chain.go
```diff
@@ -0,0 +1,400 @@
...
+func (c *Chain) dispatchBlockToSpawnedEngine(block *model.Block, src peer.ID) (dispatched bool) {
+	// abort if we do not have a spawned engine
+	engineInstance := c.Engine.Get()
+	if engineInstance == nil {
+		return false
+	}
+
+	// abort if the target slot is below the latest commitment
+	issuingTime := block.ProtocolBlock().Header.IssuingTime
+	targetSlot := engineInstance.APIForTime(issuingTime).TimeProvider().SlotFromTime(issuingTime)
```
jonastheis, line 331: Why not take the slot from the block directly?

pkg/protocol/chains.go
hmoog: Most of the comments are AI generated and completely up for discussion. I mostly focused on removing all the unnecessary quirks in the code that required additional explanations but to be honest, we should most probably even discuss the naming of the properties.

Maybe sth like "HeaviestUnverifiedCandidate" would already much better to convey the idea of what kind of "knowledge" this property is supposed to capture,

pkg/protocol/chains.go
hmoog: Same as previous comments - lets's just merged it?

pkg/protocol/chains.go
hmoog: same as before.

pkg/testsuite/mock/node.go
hmoog: Correct.

pkg/protocol/protocol_attestations.go
hmoog: Fixed.

pkg/protocol/protocol_blocks.go
hmoog: As I said in a previous comment; currently during shutdown but dependening on our shutdown strategy it might turn into a superflouous check that we should get rid of over time while hardening the code,

pkg/protocol/protocol.go
hmoog: The current code is not in its final format.

Ideally I would like to have a dedicated sub-structure for each "name space" (Attestations, Blocks, Chains, Engines, Commitments, WarpSync and so on) that not only forms a human-navigateable API but that also separates the logic in logically separated sub-domains.

Keeping each sub-domain Public for the possible exposure of API, makes imho a lot of sense even if some methods that are currently exposed turn out to be unnecessary and could be turned private instead - the API is just nice with this separation of sub components:

protocol.Engines.Main
protocol.Chains.Main
protocol.Attestations.Request(...)

pkg/protocol/protocol_commitments.go
hmoog: But its the other way round. You want to report errors that indicate a DB issues / storage issue and ignore the ones where you are actually just querying for a missing commitment.

What is wrong with toggling the log level? Thats exactly what you should do?

pkg/protocol/protocol_commitments.go
hmoog: That plays into the realm of how to handle our eviction strategy and yes, I think every node should solidify all commitments up till the genesis.

The easiest solution would be to add a check if the requested slot is below our root commitment and then resolve these requests against the permanent storage of our main engine instead of looking up the matching engine dynamically but this would have to be adressed in an additional PR.

pkg/protocol/protocol_warp_sync.go
hmoog: We can add them.

pkg/protocol/protocol_warp_sync.go
hmoog: I agree, everything should be reactive / closure / callback based.

pkg/tests/loss_of_acceptance_test.go
```diff
@@ -220,6 +222,7 @@ func TestLossOfAcceptanceWithRestartFromDisk(t *testing.T) {
 	node2 := ts.AddNode("node2")
 
 	ts.Run(true, nil)
+	node2.Protocol.SetLogLevel(log.LevelTrace)
```
jonastheis: remove

pkg/tests/loss_of_acceptance_test.go
```diff
@@ -126,6 +127,7 @@ func TestLossOfAcceptanceFromSnapshot(t *testing.T) {
...
```
jonastheis: remove

pkg/protocol/protocol_blocks.go
jonastheis: Could be addressed with #591

pkg/protocol/protocol_attestations.go
hmoog: It is called as soon as the chain is removed from the Chains collection.

WithElements sets up a behavior for each element of the chain that gets its shutdown function called as soon as the element is removed again.

pkg/protocol/protocol_attestations.go
hmoog: The original idea was to allow attestations to be "updated" if we for example receive more attestations from a different neighbor with a higher weight and for this, access to the weight needs to be threat safe + we want to keep the current value in the error case.

Since I currently just ported the existing logic without implementing such an additional update mechanism to keep the logic changes minimal, we could also synchronize over the attested flag but since that change is already planned, I would keep it as is.

pkg/protocol/chains.go
hmoog: If we want to do renames, we should revisit all properties and have a proper discussion around naming but I wouldnt address that in this PR.

pkg/protocol/commitment_verifier.go
hmoog: I am not sure which comments you mean as this method didn't have a comment before.

If you are talking about the comments about the cumulative weight and the corresponding cutoff condition, then I don't get it either because this calculation and the corresponding condition where replaced by a different check?!?

pkg/protocol/protocol.go
hmoog: Because we don't just want to process the block in our engine but we also want to feed it to our latestSeenBlocks and so on that listen to these kind of network events.

Ideally you want to inject things where they would usually arrive, to make sure you do exactly the same things with all blocks.

pkg/protocol/options.go
hmoog: Maybe - but also the scope of a different PR.

pkg/protocol/engine/engine.go
hmoog: This actually just unsubscribes from changes to the log level of the parent logger, so maybe we should rename that to unsubscribeFromParentLogger

pkg/protocol/commitments.go
hmoog: Why is it odd? It is used to create a persistent state across multiple events by capturing the shared value in a closure?

Thats exactly how you always do data flows. I.e. when you solidify in the BlockDAG, you create a shared variable unsolidParentsCounter and you trigger an event once this number reaches 0.

Its similar here where we want to share a persistent state across multiple events.

pkg/protocol/commitments.go
hmoog: Because thats the root of the chain.

pkg/protocol/engine/engine.go
hmoog: removed it

pkg/protocol/engine/engine.go
hmoog: we log it now

pkg/protocol/engine/engine.go
hmoog: fixed

pkg/protocol/engines.go
hmoog: I just copied the EngineManager as is and simply added the reactive variables to minimize changes for possible merged but since we are close to merge I made all the methods that are not used in external tests private.

pkg/protocol/engines.go
hmoog: because we do not want to setup a reactive behavior with a shutdown function, but we just want to setup a simple function call.

pkg/protocol/engines.go
hmoog: dunno, I changed it.

pkg/protocol/engines.go
hmoog: fixed

pkg/protocol/protocol.go
hmoog: yes

pkg/protocol/chainmanager/manager.go
hmoog: We have a PreSolidBlockFilter where we check the version

pkg/protocol/commitment.go
hmoog: yes

pkg/protocol/commitments.go
hmoog: Hum? I was under the assumption that a commitment that we produce ourselves will simply be seen as having the full weight of that commitment?

pkg/protocol/commitment.go
hmoog: Yes, but the shutdown methods in the setup will unregister themselves on eviction.

pkg/protocol/chain.go
hmoog: commitments are associated with a chain in solidification order, so they arrive here in the correct order,

pkg/protocol/commitment.go
hmoog: We will see this already through the tracelogging of the properties (a MainChild change, followed by a sequence of Chain updates).

Since the only time the MainChild is changed is here, I dont think we need an additional log. Additional logs should rather relate to data flow that does not result in dedicated variable updates.

pkg/protocol/commitment.go
hmoog: It doesn't really matter which chain is labeled how as the chain switching logic will eventually correct any mislabeling, but you are right that we could pre-select the heaviest child to have less state changes in case we happen to pick the wrong one first.

These kind of reorgs are so seldom that it felt less complicated in code to just leave the selection mechanism to the chainswitching part and just make sure things are uniquely labeled here.

pkg/protocol/commitment.go
hmoog: I fixed it

pkg/protocol/engines.go
hmoog: yep

pkg/protocol/chains.go
hmoog: I fixed it

pkg/protocol/protocol_blocks.go
```diff
@@ -0,0 +1,153 @@
...
+		b.LogTrace("received block", "blockID", block.ID(), "commitment", commitment.LogName())
+	})
+}
+
+// ProcessRequest processes the given block request.
+func (b *BlocksProtocol) ProcessRequest(blockID iotago.BlockID, from peer.ID) {
+	b.workerPool.Submit(func() {
+		block, exists := b.protocol.Engines.Main.Get().Block(blockID)
+
+		// TODO: CHECK why blocks can be nil here:
```
jonastheis: Let's remove this, there's an issue #556 that tracks this already. I wouldn't even add the check for nil or at least explicitly log it.

pkg/protocol/chainmanager/manager.go
jonastheis: Yes that is for blocks. This check here however, is meant for commitments and their respective compatibility of slot and protocol version.

pkg/protocol/protocol_warp_sync.go
jonastheis: Tracked in issue #597

pkg/protocol/options.go
jonastheis: Tracked in issue #598

pkg/protocol/protocol_warp_sync.go
jonastheis: I added them

pkg/protocol/protocol_attestations.go
jonastheis: I added them

PR #295, opened 2023-08-22
Add latest supported protocol version to Accounts ledger

piotrm50:
This PR adds latest supported protocol version to Accounts ledger and tracks its changes in AccountsDiff storage. 

Accompanying PR: https://github.com/iotaledger/iota.go/pull/488


jonastheis (changes requested):
The latest supported version of a validator alone does not yield sufficient information for a delegator to change delegation in case there's a vote for a protocol upgrade. This is because the version just describes the protocol version the validator supports but not the specific incarnation thereof. We need to also use the hash of the protocol parameters so that the version with its specific parameters is clear.

pkg/storage/prunable/accountdiffs.go
```diff
@@ -177,6 +185,26 @@ func (d *AccountDiff) readFromReadSeeker(reader io.ReadSeeker) (offset int, err
 	}
 	offset += 8
 
+	newVersionAndHashBytes := make([]byte, iotago.IdentifierLength+1)
```
jonastheis, line 188: Same here, use a constant to make it much easier maintainable.

pkg/storage/prunable/accountdiffs.go
```diff
@@ -177,6 +185,26 @@ func (d *AccountDiff) readFromReadSeeker(reader io.ReadSeeker) (offset int, err
...
+	if err = binary.Read(reader, binary.LittleEndian, newVersionAndHashBytes); err != nil {
+		return offset, ierrors.Wrap(err, "unable to read new version and hash bytes in the diff")
+	}
+	d.NewLatestSupportedVersionAndHash, _, err = iotago.VersionAndHashFromBytes(newVersionAndHashBytes)
+	if err != nil {
+		return offset, ierrors.Wrap(err, "unable to parse new version and hash bytes in the diff")
+	}
+	offset += len(newVersionAndHashBytes)
+
+	prevVersionAndHashBytes := make([]byte, iotago.IdentifierLength+1)
```
jonastheis, line 198: Here too

jonastheis (changes requested):

jonastheis (approved):

PR #297, opened 2023-08-23
Fix deficit and initial RMC value bugs

piotrm50:
Fix deficit calculation so it does not go below 0.
Fix RMC calculation after starting a new network from genesis snapshot. Previously RMC was set to zero from the first slot until RMC increased despite minRMC in protocol parameters being set to a non-zero value. 
Fix work score calculation - previously work score for a block was always equal to `1`, now it's correctly calculated in accordance to WorkScoreStructure part of ProtocolParameters. 
Added sample WorkScoreOptions for in snapshot presets.


pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
```diff
@@ -380,8 +401,9 @@ func (s *Scheduler) selectIssuer(start *IssuerQueue, slotIndex iotago.SlotIndex)
 
 			denom, err := safemath.SafeAdd(remainingDeficit, quantum-1)
```
cyberphysic4l, line 402:
```suggestion
			numerator, err := safemath.SafeAdd(remainingDeficit, quantum-1)
```

pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
```diff
@@ -380,8 +401,9 @@ func (s *Scheduler) selectIssuer(start *IssuerQueue, slotIndex iotago.SlotIndex)
...
 			if err != nil {
-				denom = s.maxDeficit()
+				denom = math.MaxInt64
```
cyberphysic4l, line 404:
```suggestion
				numerator = math.MaxInt64
```

pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
```diff
@@ -380,8 +401,9 @@ func (s *Scheduler) selectIssuer(start *IssuerQueue, slotIndex iotago.SlotIndex)
...
 			}
+
 			r, err := safemath.SafeDiv(denom, quantum)
```
cyberphysic4l, line 407:
```suggestion
			r, err := safemath.SafeDiv(numerator, quantum)
```

cyberphysic4l (approved):
just change the word denom as I incorrectly named that as it is a numerator in the division rather than denominator. Otherwise all looks ok to me.

PR #298, opened 2023-08-24
Add synced flag and trigger UpdatedStatus event in SyncManager

### LINKED ISSUE #270: SyncManager: Handle changes to the bootstrapped state to trigger updates

jkrvivian:
* Move `syncManager` from protocol to engine
* Change bootstrapped status for once
* Implement `Synced` update method and getter in `syncManager`
* Only access sync status via `syncManager`
* Trigger `updatedStatus` event if synced status changed

Close #270


pkg/protocol/engine/engine.go
```diff
@@ -273,23 +275,9 @@ func (e *Engine) Block(id iotago.BlockID) (*model.Block, bool) {
 	return modelBlock, modelBlock != nil
 }
 
+// IsBootstrapped returns true if the node is bootstrapped. Do not get bootstrapped status with this method, but always get it from SyncManager. This method is needed and exported to access component inside the engine, so we could test it with different bootstrapped conditions.
```
piotrm50, line 278: It's a little confusing to have a method and discourage its use. I don't see this method used anywhere in tests - are we certain that it needs to be exposed? Or that it needs to exist at all? Maybe we can pass the `optsIsBootstrappedFunc` directly to all the components that need it, instead of `e.IsBootstrapped`?

pkg/protocol/syncmanager/trivialsyncmanager/syncmanager.go
```diff
@@ -1,22 +1,27 @@
 package trivialsyncmanager
 
 import (
+	"time"
```
piotrm50, line 4: Move `syncmanager` directory under `engine` as all engine subcomponents are placed there.

piotrm50 (changes requested):
Looks good overall, but please answer the comments.

jonastheis (changes requested):

jonastheis (approved):

piotrm50 (approved):

PR #303, opened 2023-08-29
Resolve ledger mempool TODOs

piotrm50:
Use error handler instead of panicking.
Extend account test with delegation output claiming. Full test set will be implemented as part of the VM hardening that covers all edge cases. 
Remove some outdated TODOs.

### DIFF
--- a/pkg/tests/accounts_test.go
+++ b/pkg/tests/accounts_test.go
@@ -18,7 +18,7 @@ import (
 	iotago "github.com/iotaledger/iota.go/v4"
 )
 
-// TODO: implement tests for staking and delegation transitions
+// TODO: implement tests for staking and delegation transitions that cover edge cases - part of hardening phase.
 func Test_TransitionAccount(t *testing.T) {
 	oldGenesisOutputKey := utils.RandPubKey()
 	ts := testsuite.NewTestSuite(t, testsuite.WithAccounts(snapshotcreator.AccountDetails{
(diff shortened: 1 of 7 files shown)


pkg/tests/accounts_test.go
```diff
@@ -260,6 +258,51 @@ func Test_TransitionAccount(t *testing.T) {
 		ValidatorStake:  10000,
 	}, ts.Nodes()...)
 
+	// transition a delegation output to a delayed claiming state
+	inputForDelegationTransition, delegationTransitionOutputs, delegationTransitionWallets := ts.TransactionFramework.DelayedClaimingTransition("TX3:0", 0)
```
PhilippGackstatter, line 262: I thought about this transition to end epoch = 0, and it is fine. The end epoch only matters when claiming in delayed claiming state, not for the regular state. To get to the delayed claiming state, one has to go through a transition in which the `End Epoch` must be set to its correct value (which may be zero as it is here). In other words, before a delegatin output is in delayed claiming state the value of `End Epoch` does not matter and is not looked at. Thus using a default of 0 for the end epoch is fine. We could also use any other value, but 0 makes reasonable sense. 

Delayed Claiming itself is identified by a non-zero delegation ID rather than anything based on end epoch, so it's fine from that point of view as well. When transitioning a delegation output in epoch 0 to delayed claiming, one sets end epoch to 0, which seems weird, but is correct.

PhilippGackstatter (approved):
Looks good to me!

PR #306, opened 2023-08-29
Update to go 1.21 and update modules

muXxer:


jonastheis (commented):
Everything seems fine but the Docker network does not work. It seems that nodes are not issuing anything for some reason.

jonastheis (approved):
Seems to be that it's unrelated to this PR. With 7ec1520eb4a14e39228fdf14c633a8c4d685becf it still worked

PR #311, opened 2023-08-30
Adapt to Block Issuer Keys interface changes

PhilippGackstatter:
Companion PR to iotaledger/iota.go#489.


pkg/utils/rand.go
```diff
@@ -231,16 +231,16 @@ func RandOutputOnAddressWithAmount(outputType iotago.OutputType, address iotago.
...
 
-func RandPubKey() ed25519.PublicKey {
-	return ed25519.PublicKey(RandBytes(32))
+func RandBlockIssuerKey() iotago.BlockIssuerKey {
+	return iotago.BlockIssuerKeyEd25519FromPublicKey(ed25519.PublicKey(RandBytes(32)))
 }
 
-func RandPubKeys() []ed25519.PublicKey {
+func RandBlockIssuerKeys() iotago.BlockIssuerKeys {
 	length := RandomIntn(10)
```
muXxer, line 239: Should that maybe be `length := RandomIntn(9) +1 ` so we always have at least 1 IssuerKey?

pkg/testsuite/testsuite.go
```diff
@@ -400,7 +400,12 @@ func (t *TestSuite) Run(failOnBlockFiltered bool, nodesOptions ...map[string][]o
 			}
 
 			if accountDetails.AccountID.Empty() {
-				accountDetails.AccountID = blake2b.Sum256(accountDetails.IssuerKey[:])
+				blockIssuerKeyEd25519, ok := accountDetails.IssuerKey.(iotago.BlockIssuerKeyEd25519)
```
muXxer, line 403: It doesn't need to be of type ed25519 I guess.
We could also directly hash the IssuerKeyBytes.

pkg/protocol/snapshotcreator/snapshotcreator.go
```diff
@@ -71,7 +70,15 @@ func CreateSnapshot(opts ...options.Option[Options]) error {
 	for _, accountData := range opt.Accounts {
 		// Only add genesis validators if an account has both - StakedAmount and StakingEndEpoch - specified.
 		if accountData.StakedAmount > 0 && accountData.StakingEpochEnd > 0 {
-			accounts.Set(blake2b.Sum256(accountData.IssuerKey[:]), &account.Pool{
+
+			blockIssuerKeyEd25519, ok := accountData.IssuerKey.(iotago.BlockIssuerKeyEd25519)
```
muXxer, line 74: It doesn't need to be of type ed25519 I guess.
We could also directly hash the IssuerKeyBytes.

pkg/utils/rand.go
muXxer: Or is it good to also test the "no block issuer keys available" case? :D

PR #312, opened 2023-08-30
Add retry loop to `RequestAttestations`

### LINKED ISSUE #259: Resolve protocol TODOs
* [x] ~Finish `ban source`  in onForkDetected~ (Move to #310 )
* [ ] Decide if we need to retry when `RequestAttestations`
* [x] ~Copy over old slots from the old engine to the new one in `switchEngines`~ (Move to #300)

file: `pkg/protocol/protocol_fork.go`

---
* [x] Distinguish between workerpool for handling requested/requesting and wp for handling received stuff

file: `pkg/protocol/protocol_network.go`

jkrvivian:
* Add a retry loop with ticker to `RequestAttestations`, an error is returned if attestation is not received within the loop.
* Add options to max retry and retry time interval to protocol options.

Close #259


pkg/protocol/protocol_fork.go
```diff
@@ -213,9 +212,6 @@ func (p *Protocol) processFork(fork *chainmanager.Fork) (anchorBlockIDs iotago.B
 	defer unhook()
 	defer wp.Shutdown()
 
-	ctx, cancel := context.WithTimeout(p.context, 2*time.Minute)
-	defer cancel()
```
jonastheis, line 217: Taking the parent context `p.context` and getting a new one with a timeout is actually necessary here as the child context gets also cancelled if the parent gets cancelled. This is important for shutting down the node properly without hanging this goroutine until the final timeout passed. Also this whole timeout is for limiting the time we will try to process the fork. There might be hundreds of slots to verify etc so that sets an upper limit to that as well

pkg/protocol/protocol_fork.go
```diff
@@ -247,32 +243,41 @@ func (p *Protocol) processFork(fork *chainmanager.Fork) (anchorBlockIDs iotago.B
...
-			case result.actualCumulativeWeight < mainChainCommitment.CumulativeWeight():
-				heavierCount--
-			default:
-				heavierCount = 0
-			}
-
-			if decided, doSwitch := forkChoiceRule(heavierCount); decided {
-				return anchorBlockIDs, doSwitch, false, nil
+		ticker := time.NewTicker(p.optsAttestationRequesterTryInterval)
+		defer ticker.Stop()
```
jonastheis, line 247: As the `defer` is function scoped, it might actually not be properly called in this setup right now. I'd recommend putting everything from the ticker and the loop into its own function. 

This should also allow you to get rid of the `received` variable but instead just return the result or an error. 
The switch should still be evaluated on this level so that the function really just does request (potentially multiple) times and waits for the response without having actual logic in it

jonastheis (changes requested):
Looks good, but some small things to correct and simplify

jonastheis (approved):
Just some nitpicking

PR #316, opened 2023-08-31
Autopeering & network stack and protocol flattening

### LINKED ISSUE #49: Port autopeering w/ Kademlia discovery
Kademlia hop distance w/out mana distance

karimodm:
- All peer identity.ID has been migrated to libp2p IDs. There is no reason why we should have specialized IDs for our nodes while all the connection's authentication works using p2p identities and handshakes.
- All network addresses became p2p's multiaddresses: flexible and unique way to bind to interfaces and connect to new peers, with p2p identification by default. 
- Flattened protocol structure in one-step single connection design: 
  - Everything on one port and connection.
  - Single protocol registration and handshake.
  - Removed dummy services.
  - Removed two-steps connect -> dial / accepted matcher logic.
- Autopeering via Kademlia DHT on the same connection.
- Discovery SeedPeers via expirable known peers in PeerDB.
- Removed old code hive.go/autopeering dependency completely.
- maxPeers parameter.
- knownPeers & bootstrapPeers parameters: the former is the manualpeering set of peers the node will constantly try to keep a connection against, the latter is the initial list of peers used for Kademlia DHT discovery.

Fixes #49


pkg/network/p2p/manager.go
```diff
@@ -48,125 +67,142 @@ func WithNoDefaultTimeout() ConnectPeerOption {
...
+	)
 
-// AddInbound tries to add a neighbor by accepting an incoming connection from that peer.
-func (m *Manager) AddInbound(ctx context.Context, p *peer.Peer, group NeighborsGroup,
-	connectOpts ...ConnectPeerOption,
-) error {
-	m.log.Debugw("adding inbound neighbor", "peer", p.ID())
-	return m.addNeighbor(ctx, p, group, m.acceptPeer, connectOpts)
+	if err := m.peerDB.UpdatePeer(peer); err != nil {
+		return ierrors.Wrapf(err, "failed to update peer %s", peer.ID)
```
muXxer, line 167: Shouldn't we close the stream here as well?

pkg/network/peer.go
```diff
@@ -0,0 +1,142 @@
...
+type Peer struct {
+	ID            p2ppeer.ID
+	PublicKey     ed25519.PublicKey
+	PeerAddresses []ma.Multiaddr
+	ConnStatus    *atomic.Value
+	RemoveCh      chan struct{}
+	DoneCh        chan struct{}
+}
+
+func NewPeerFromAddrInfo(addrInfo *p2ppeer.AddrInfo) (*Peer, error) {
```
muXxer, line 42: This can never return an error.

pkg/network/autopeering/autopeering.go
```diff
@@ -0,0 +1,155 @@
...
+		util.Advertise(m.ctx, m.routingDiscovery, m.networkID, discovery.TTL(5*time.Minute))
+
+		go m.discoveryLoop()
+
+		m.isStarted.Store(true)
+	})
+}
+
+// Stop terminates internal background workers. Calling multiple times has no effect.
+func (m *Manager) Stop() (err error) {
```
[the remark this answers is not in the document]
alexsporn: if you remove the label you need to remove the `( )` too

components/p2p/component.go
```diff
@@ -11,16 +11,20 @@ import (
 	"github.com/libp2p/go-libp2p/core/host"
 	"go.uber.org/dig"
 
+	ma "github.com/multiformats/go-multiaddr"
```
alexsporn, line 14: why the custom import label?

components/p2p/params.go
```diff
@@ -6,8 +6,8 @@ import (
 
 // ParametersP2P contains the definition of configuration parameters used by the p2p plugin.
 type ParametersP2P struct {
-	// BindAddress defines on which address the p2p service should listen.
-	BindAddress string `default:"0.0.0.0:14666" usage:"the bind address for p2p connections"`
+	// BindAddress defines on which multi address the p2p service should listen on.
+	BindAddress string `default:"/ip4/0.0.0.0/tcp/14666" usage:"the multi address for p2p connections"`
```
alexsporn, line 10: the default should bind both ipv4 and ipv6. Is this covered by this multiaddr?

components/protocol/component.go
```diff
@@ -7,15 +7,15 @@ import (
 	"github.com/labstack/gommon/bytes"
 	"go.uber.org/dig"
 
+	p2ppeer "github.com/libp2p/go-libp2p/core/peer"
```
alexsporn, line 10: why the custom import name? there should be no collision with hive.go anymore if you use the default `peer`

tools/docker-network/.env
```diff
@@ -16,5 +16,5 @@ COMMON_CONFIG="
 --protocol.snapshot.path=/app/data/snapshot.bin
 "
 MANUALPEERING_CONFIG="
---p2p.knownPeers='[{\"address\":\"peer_master:14666\", \"publicKey\":\"F9G7rptEGXrNtHmdXoLHK6JcWovoSJiebtxeTJ5N1j4b\"}, {\"address\":\"peer_master2:14666\", \"publicKey\":\"HPgXMEvZPQkRAgRFa8Ns6cjYMJTQcG2zxWxuXAnkYMBW\"}, {\"address\":\"faucet:14666\", \"publicKey\":\"DKThLsdMKVLSWaVb2M8mPyz5mAdHxrG3mvLRMk3FY1T5\"}, {\"address\":\"peer_nomana:14666\", \"publicKey\":\"FXyrkXLUQrVvhoswhQ2KEc8xjT9GYPFJrZkDEq51E5zb\"}, {\"address\":\"peer_nomana2:14666\", \"publicKey\":\"8nFcKDqA4n7UmDc8Jr7m5xSzFcq5EftGxA9wmB1C8PoV\"}]'
```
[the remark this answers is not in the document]
karimodm: No, knownPeers is still a slice of strings containing multiaddresses. I just add only one peer to let autopeering do the rest of the job.

tools/docker-network/.env
```diff
@@ -1,6 +1,6 @@
 COMMON_CONFIG="
 -c config.json
---logger.level=info
+--logger.level=debug
```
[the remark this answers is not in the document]
karimodm: Yes, I always found annoying not to have debug output in the docker network by default.

pkg/network/p2p/manager.go
```diff
@@ -2,19 +2,38 @@ package p2p
...
+	ma "github.com/multiformats/go-multiaddr"
+
 	"github.com/iotaledger/hive.go/ierrors"
 	"github.com/iotaledger/hive.go/logger"
 	"github.com/iotaledger/hive.go/runtime/syncutils"
 	"github.com/iotaledger/iota-core/pkg/network"
 )
 
+const (
+	protocolID               = "iota-core/0.0.1"
```
alexsporn, line 23: can we do proper semantic versioning here? it should be 1.0.0

components/p2p/component.go
karimodm: This is how it is imported in the libp2p literature and example code.

pkg/network/manualpeering/manualpeering.go
```diff
@@ -68,52 +29,38 @@ type KnownPeer struct {
 type Manager struct {
 	p2pm              *p2p.Manager
 	log               *logger.Logger
-	local             *peer.Local
 	startOnce         sync.Once
 	isStarted         atomic.Bool
 	stopOnce          sync.Once
 	stopMutex         syncutils.RWMutex
 	isStopped         bool
```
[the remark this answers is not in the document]
karimodm: Without an atomic bool it could happen that you still read false when it was actually true, for something read so infrequently and without consequences if you process just another loop I thought it was not really necessary. For the `isStarted` this reasoning could also apply, but there you might actually panic during initialization, as you are trying to process something on a component that should be started but you read it is not. I can change it, but I believe it is unnecessary.

pkg/network/p2p/manager.go
karimodm: Yes we should, as we otherwise will leave a negotiated stream in the air without adding it to our list of known neighbors.

components/p2p/params.go
muXxer: @alexsporn that can be ignored. It's fixed in my PR.

components/p2p/component.go
karimodm: Changed

tools/docker-network/.env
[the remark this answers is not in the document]
muXxer: Yeah I know that it's still a slice, I was just wondering why you removed the others. But okay, the autopeering handles it, now I get the reason for the change :D

pkg/network/manualpeering/manualpeering.go
```diff
@@ -71,14 +71,16 @@ func (m *Manager) AddPeers(peerAddrs ...multiaddr.Multiaddr) error {
 // RemovePeer removes a peer from the list of known peers.
 func (m *Manager) RemovePeer(peerID peer.ID) error {
 	m.knownPeersMutex.Lock()
-	defer m.knownPeersMutex.Unlock()
 
 	kp, exists := m.knownPeers[peerID]
 	if !exists {
 		return nil
```
muXxer, line 77: Deadlock alarm :scream:

muXxer (changes requested):

muXxer (approved):

PR #317, opened 2023-08-31
Fix issuer key to account id conversion

PhilippGackstatter:
Fix issuer key to account id conversion which broke the docker-network.

### DIFF
--- a/pkg/protocol/snapshotcreator/snapshotcreator.go
+++ b/pkg/protocol/snapshotcreator/snapshotcreator.go
@@ -70,7 +70,12 @@ func CreateSnapshot(opts ...options.Option[Options]) error {
 	for _, accountData := range opt.Accounts {
 		// Only add genesis validators if an account has both - StakedAmount and StakingEndEpoch - specified.
 		if accountData.StakedAmount > 0 && accountData.StakingEpochEnd > 0 {
-			accountID := blake2b.Sum256(accountData.IssuerKey.BlockIssuerKeyBytes())
+			blockIssuerKeyEd25519, ok := accountData.IssuerKey.(iotago.BlockIssuerKeyEd25519)
+			if !ok {
+				panic("block issuer key must be of type ed25519")
+			}
+			ed25519PubKey := blockIssuerKeyEd25519.ToEd25519PublicKey()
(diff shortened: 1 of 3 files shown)


pkg/protocol/snapshotcreator/snapshotcreator.go
```diff
@@ -70,7 +70,12 @@ func CreateSnapshot(opts ...options.Option[Options]) error {
 	for _, accountData := range opt.Accounts {
 		// Only add genesis validators if an account has both - StakedAmount and StakingEndEpoch - specified.
 		if accountData.StakedAmount > 0 && accountData.StakingEpochEnd > 0 {
-			accountID := blake2b.Sum256(accountData.IssuerKey.BlockIssuerKeyBytes())
+			blockIssuerKeyEd25519, ok := accountData.IssuerKey.(iotago.BlockIssuerKeyEd25519)
```
karimodm, line 73: I wonder if this converstion should be a method attached to the IssuerKey, hiding the type casting. The method would also return an error to be managed appropriately by the caller.
wdyt?

PhilippGackstatter: This conversion is only used in testing code and not needed in other parts, so I don't think it makes sense to "pollute" the BlockIssuerKey interface with that.
This PR was intended as a quick fix for the docker-network. Ideally, I'll find a way to not do the block issuer key -> ed25519 pubkey conversion at all and use the `BlockIssuerKeyBytes` for the account ID hashing directly. I just did the conversion because I ran out of time and didn't want to leave the docker-network broken.

karimodm: We have more than one place in the code where we have to juggle hive.go's crypto types, hence why I asked if it could be useful to have some helper code. This is fine for now though!

karimodm (approved):

pkg/protocol/snapshotcreator/snapshotcreator.go
PhilippGackstatter: I see what you meant now. Specifically adding a method on the Block Issuer Key interface is something I wouldn't do, but helper code does sound useful indeed.

PR #318, opened 2023-08-31
Update performance tracker and rewards

### LINKED ISSUE #258: Resolve performance TODOs
* [x] In `BlockAccepted`, 
	* [x] check if the block is validation block
	* [ ] ~include wannabe validator performance in the snapshot~
* [ ] ~Finish `EligibleValidatorCandidates`~
* [x] In `aggregatePerformanceFactors`, scale the result by the number of slots per epoch

file: `pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go`

---
* [x]  Move parameters to protocol params.
* [x]  Decide default value for parameters.
* [x]  In `ValidatorReward` and `DelegatorReward`, double check reward calculation: the epoch should be returned by the reward calculations and we should only loop until the current epoch, not epochEnd.
* [x]  In `poolReward`, deal with the potential overflow issue

file: `pkg/protocol/sybilprotection/sybilprotectionv1/performance/rewards.go`

---
* [x]  In `importPerformanceFactor`, 
	* [x] discuss: lower number of accounts to `uint8`
	* [x] discuss: the upper bound of performance factor, maybe `uint16` is enough
* [x]  In `exportPoolRewards`, adjust the loop to calculate rewards

file: `pkg/protocol/sybilprotection/sybilprotectionv1/performance/snapshot.go`

daria305:
Fixes #258 #155 
 - updated rewards with current TIP version
 -  properly define reward parameters
 - extended performance tracking to also consider subslot activity with `model.ValidatorPerformance`
 - included hash and version to `ValidatorPerformance`
 - implemented rewards test scenarios
 - use safmeth for rewards calculation
 - limited rewards import with pruning threshold


pkg/protocol/sybilprotection/sybilprotectionv1/performance/snapshot.go
```diff
@@ -264,32 +265,31 @@ func (t *Tracker) exportPoolRewards(pWriter *utils.PositionedWriter, targetEpoch
 	if err := pWriter.WriteValue("pool rewards epoch count", epochCount, true); err != nil {
 		return ierrors.Wrap(err, "unable to write epoch count")
 	}
-	// TODO: make sure to adjust this loop if we add pruning later.
-	for epoch := targetEpoch; epoch > iotago.EpochIndex(0); epoch-- {
+
+	for epoch := targetEpoch; epoch > lo.Return1(t.latestPrunedEpoch()); epoch-- {
```
jonastheis, line 269: This is imo not necessary and also not correct. Iiuc we want to export 365 epochs. However, this last pruned epoch might return all epochs forever in an extreme case where a node never pruned. Therefore, I'd say it's better to take the last 365 epoch (or as many as possible until and including 0) and simply have the check when we retrieve the rewards map implicit via the store itself. If for some reason an epoch is not available it'll fail (should never happen as we don't prune them for at least 365 epochs for this store).

pkg/protocol/sybilprotection/sybilprotectionv1/performance/snapshot.go
```diff
@@ -242,7 +243,7 @@ func (t *Tracker) exportPerformanceFactor(pWriter *utils.PositionedWriter, start
 		}); err != nil {
 			return ierrors.Wrapf(err, "unable to write performance factors for slot index %d", currentSlot)
 		}
-
+		fmt.Println("writing pf account count", accountsCount)
```
jonastheis, line 246: Remove print

pkg/protocol/sybilprotection/sybilprotectionv1/performance/snapshot.go
karimodm: Amended the for loop. I hope I understood you correctly =)

pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go
```diff
@@ -65,26 +71,69 @@ func (t *Tracker) TrackValidationBlock(block *blocks.Block) {
...
 	}
 }
 
+func (t *Tracker) updateSlotPerformanceBitMap(pf *model.ValidatorPerformance, slotIndex iotago.SlotIndex, issuingTime time.Time) *model.ValidatorPerformance {
+	if pf == nil {
+		return pf
+	}
+	subslotIndex := t.subslotIndex(slotIndex, issuingTime)
+	// set bit at subslotIndex to 1 to indicate activity in that subslot
+	pf.SlotActivityVector = pf.SlotActivityVector | (1 << subslotIndex)
```
oliviasaa, line 123: I don't know if this is intentional, but the last bit of SlotActivityVector is always 0, no? since subslotIndex is defined in line 134 starting at 1

karimodm: Yep, you are right! Started the indexing from 0 instead of 1

pkg/protocol/sybilprotection/sybilprotectionv1/performance/rewards.go
```diff
@@ -156,43 +188,99 @@ func (t *Tracker) rewardsForAccount(accountID iotago.AccountID, epochIndex iotag
...
-	return uint64(increaseAccuracy(totalValidatorsStake, profitMarginExponent) / (totalValidatorsStake + totalPoolStake))
+func (t *Tracker) calculateProfitMargin(totalValidatorsStake, totalPoolStake iotago.BaseToken, epoch iotago.EpochIndex) uint64 {
+	return uint64(scaleUpWithExponent(totalValidatorsStake, t.apiProvider.APIForEpoch(epoch).ProtocolParameters().RewardsParameters().ProfitMarginExponent) / (totalValidatorsStake + totalPoolStake))
 }
 
-// increaseAccuracy shifts the bits of the given value to the left by the given amount, so that the value is moved to the power of 2^accuracyShift.
-// TODO make sure that we handle overflow here correctly if the inserted value is > 2^(64-accuracyShift).
-func increaseAccuracy[V iotago.BaseToken | iotago.Mana | uint64](val V, shift uint64) V {
+// scaleUpWithExponent shifts the bits of the given value to the left by the given amount, so that the value is moved to the power of 2^accuracyShift.
+func scaleUpWithExponent[V iotago.BaseToken | iotago.Mana | uint64](val V, shift uint8) V {
```
oliviasaa, line 256: shouldn't we move this function to the safemath package?

pkg/protocol/sybilprotection/sybilprotectionv1/performance/rewards.go
```diff
@@ -156,43 +188,99 @@ func (t *Tracker) rewardsForAccount(accountID iotago.AccountID, epochIndex iotag
...
+	poolCoeff, err := safemath.SafeAdd(result1, result2)
+	if err != nil {
+		return 0, ierrors.Wrapf(err, "failed to calculate pool coefficient due to overflow for slot %d", slot)
 	}
 
-	return (aux2 >> finalRewardScalingExponent) - fixedCost
+	return uint64(poolCoeff), nil
 }
 
 // calculateProfitMargin calculates the profit margin of the pool by firstly increasing the accuracy of the given value, so the profit margin is moved to the power of 2^accuracyShift.
```
oliviasaa, line 250:
```suggestion
// calculateProfitMargin calculates a common profit margin for all validators by firstly increasing the accuracy of the given value, so the profit margin is moved to the power of 2^accuracyShift.
```

pkg/protocol/sybilprotection/sybilprotectionv1/performance/rewards.go
```diff
@@ -65,19 +56,49 @@ func (t *Tracker) ValidatorReward(validatorID iotago.AccountID, stakeAmount iota
...
 		}
 
-		unDecayedEpochRewards := uint64(rewardsForAccountInEpoch.FixedCost) +
-			decreaseAccuracy(poolStats.ProfitMargin*uint64(rewardsForAccountInEpoch.PoolRewards), profitMarginExponent) +
-			decreaseAccuracy(increasedAccuracyComplement(poolStats.ProfitMargin, profitMarginExponent)*uint64(rewardsForAccountInEpoch.PoolRewards), profitMarginExponent)*
-				uint64(stakeAmount)/
-				uint64(rewardsForAccountInEpoch.PoolStake)
+		// all rewards go to delegators, as the fixed cost was too high
+		if rewardsForAccountInEpoch.PoolRewards < rewardsForAccountInEpoch.FixedCost {
+			return 0, 0, 0, nil
```
oliviasaa, line 61: I always have problems testing things in this repo, but isn't this `return` zeroing the validator rewards for the whole loop (from `epochStart` to `epochEnd`)?

pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go
```diff
@@ -144,46 +188,90 @@ func (t *Tracker) ApplyEpoch(epoch iotago.EpochIndex, committee *account.Account
...
+
+		if pf.BlockIssuedCount > t.apiProvider.APIForEpoch(epoch).ProtocolParameters().RewardsParameters().ValidatorBlocksPerSlot {
+			// we harshly punish validators that issue any blocks more than allowed
+
+			return 0
+		}
+		epochPerformanceFactor += uint64(slotPerformanceFactor)
+	}
+
+	return epochPerformanceFactor
```
oliviasaa, line 215:
```suggestion
	return epochPerformanceFactor >> SlotsPerEpochExponent
```
no?

oliviasaa: actually, maybe that's not the best place to add this `SlotsPerEpochExponent`; maybe it's best to do this shift when we use the performance factor and divide it by `validatorBlocksPerSlot`. Anyway, I don't see the `SlotsPerEpochExponent` being accounted for anywhere in the code, am I missing something?

pkg/protocol/sybilprotection/sybilprotectionv1/performance/rewards.go
karimodm: Yes, this would return 0 validator rewards. Is this correct?

pkg/protocol/sybilprotection/sybilprotectionv1/performance/rewards.go
```diff
@@ -156,43 +188,99 @@ func (t *Tracker) rewardsForAccount(accountID iotago.AccountID, epochIndex iotag
...
 
-// decreaseAccuracy reversts the accuracy operation of increaseAccuracy by shifting the bits of the given value to the right by the profitMarginExponent.
+// scaleDownWithExponent reversts the accuracy operation of scaleUpWithExponent by shifting the bits of the given value to the right by the profitMarginExponent.
 // This is a lossy operation. All values less than 2^accuracyShift will be rounded to 0.
-func decreaseAccuracy[V iotago.BaseToken | iotago.Mana | uint64](val V, shift uint64) V {
+func scaleDownWithExponent[V iotago.BaseToken | iotago.Mana | uint64](val V, shift uint8) V {
+	result := val >> shift
+	// if the result is greater than the original value, we have an overflow
+	if val < result {
+		panic("overflow on a bit shift operation")
```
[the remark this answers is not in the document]
karimodm: It cannot =)

pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go
karimodm: I've applied your suggestion; indeed `SlotsPerEpochExponent` wasn't used anywhere except when subtracted to the generationRateExponent to shift to the final result of the rewards https://github.com/iotaledger/iota.go/blob/81e2dc34386938e2fbd862ce3d21800470f38df5/mana.go#L69

karimodm: This test now fails https://github.com/iotaledger/iota-core/blob/178f30e00dff24b4581250e6471a781dfe714e00/pkg/protocol/sybilprotection/sybilprotectionv1/performance/tracker_test.go#L9

pkg/protocol/sybilprotection/sybilprotectionv1/performance/rewards.go
oliviasaa: I thought about just zeroing the reward for the epoch in question (the one where `rewardsForAccountInEpoch.PoolRewards`<`rewardsForAccountInEpoch.FixedCost`)

I have to think a little bit about whether this is a problem or not, maybe that's too harsh, idk...

pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go
```diff
@@ -144,46 +192,90 @@ func (t *Tracker) ApplyEpoch(epoch iotago.EpochIndex, committee *account.Account
...
+	pf.SlotActivityVector = pf.SlotActivityVector | (1 << t.subslotIndex(slotIndex, issuingTime))
+
+	return pf
+}
+
+// subslotIndex returns the index for timestamp corresponding to subslot created dividing slot on validatorBlocksPerSlot equal parts.
+func (t *Tracker) subslotIndex(slot iotago.SlotIndex, issuingTime time.Time) int {
+	valBlocksNum := t.apiProvider.APIForEpoch(t.latestAppliedEpoch).ProtocolParameters().RewardsParameters().ValidatorBlocksPerSlot
+	subslotDur := time.Duration(t.apiProvider.APIForEpoch(t.latestAppliedEpoch).TimeProvider().SlotDurationSeconds()) * time.Second / time.Duration(valBlocksNum)
+	slotStart := t.apiProvider.APIForEpoch(t.latestAppliedEpoch).TimeProvider().SlotStartTime(slot)
```
jonastheis, line 278: Just get the API once here instead of 3 times

pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go
```diff
@@ -144,46 +192,90 @@ func (t *Tracker) ApplyEpoch(epoch iotago.EpochIndex, committee *account.Account
...
-		return nil, false
+	validatorPerformance, err := validatorPerformances.Load(block.ProtocolBlock().IssuerID)
+	if err != nil {
+		t.errHandler(ierrors.Errorf("failed to load performance factor for account %s", block.ProtocolBlock().IssuerID))
+	}
+	// key not found
+	if validatorPerformance == nil {
+		validatorPerformance = model.NewValidatorPerformance()
+	}
+	updatedPerformance := t.updateSlotPerformanceBitMap(validatorPerformance, block.ID().Index(), block.ProtocolBlock().IssuingTime)
```
jonastheis, line 246: don't see the need for this function as it's just one line

pkg/protocol/sybilprotection/sybilprotectionv1/performance/rewards.go
```diff
@@ -61,23 +50,64 @@ func (t *Tracker) ValidatorReward(validatorID iotago.AccountID, stakeAmount iota
...
 		if poolStats == nil {
 			return 0, 0, 0, ierrors.Errorf("pool stats for epoch %d and validator accountID %s are nil", epochIndex, validatorID)
 		}
 
-		unDecayedEpochRewards := uint64(rewardsForAccountInEpoch.FixedCost) +
-			decreaseAccuracy(poolStats.ProfitMargin*uint64(rewardsForAccountInEpoch.PoolRewards), profitMarginExponent) +
-			decreaseAccuracy(increasedAccuracyComplement(poolStats.ProfitMargin, profitMarginExponent)*uint64(rewardsForAccountInEpoch.PoolRewards), profitMarginExponent)*
-				uint64(stakeAmount)/
-				uint64(rewardsForAccountInEpoch.PoolStake)
+		// all rewards go to delegators, as the fixed cost was too high
```
jonastheis, line 58: I don't understand this comment. isn't the fixed cost for the validator meaning that `no` rewards go to the delegators?

jonastheis (changes requested):
The code in this whole component is a little bit too verbose at times (and at other times not enough) and hard to read imo. Also I didn't really check the calculations as I don't know what they should be.

pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go
oliviasaa: I found the reason why the test was failing after this suggestion, we have to change something in testsuite_test.go

pkg/protocol/sybilprotection/sybilprotectionv1/performance/testsuite_test.go
```diff
@@ -122,55 +115,184 @@ func (t *TestSuite) ApplyEpochActions(epochIndex iotago.EpochIndex, actions map[
...
+	profitMarging := t.calculateProfitMargin(totalValidatorsStake, totalStake)
+	t.epochStats[epochIndex] = &model.PoolsStats{
+		TotalStake:          totalStake,
+		TotalValidatorStake: totalValidatorsStake,
+		ProfitMargin:        profitMarging,
+	}
+	t.poolRewards[epochIndex] = make(map[string]*model.PoolRewards)
+
+	for alias, action := range actions {
+		epochPerformanceFactor := action.SlotPerformance * action.ActiveSlotsCount
```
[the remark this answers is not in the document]
oliviasaa: now the tests should pass

pkg/protocol/sybilprotection/sybilprotectionv1/performance/rewards.go
oliviasaa: this is a form of punishing the validator for adding an unreasonably high fixed cost. if fixed cost>pool reward, we "slash" the validator reward

oliviasaa: however, the way it is now, the validator will get zero for the whole time period, not for the specific epochs where this happened. It's not an urgent matter, but I feel like this can be a problem in the future with the random committees

oliviasaa: since the validator won't know in advance the order of magnitude of his reward for every single epoch, I think this could become a punishment for unlucky people in the future :/

pkg/protocol/sybilprotection/sybilprotectionv1/performance/snapshot.go
oliviasaa: I thought about it and I'd remove lines 59 to 61 and instead, I'd put 63 to 107 inside 

`if rewardsForAccountInEpoch.PoolRewards >= rewardsForAccountInEpoch.FixedCost`

oliviasaa: this would zero only the epoch in question, right?

oliviasaa: btw, I can't find this condition in the delegator's reward function. Please see my suggestion for lines 150-177

oliviasaa: btw, now that I think about it, this check does no work!
See, suppose the pool reward was 15 and the fixed cost was 10. Then the function `poolReward` would spit you out 5. 
When you do this check, `rewardsForAccountInEpoch.PoolRewards<rewardsForAccountInEpoch.FixedCost=TRUE`, and the validator will be punished

oliviasaa: this logic at the end of funtion `poolReward` is kind of broken IMO (lines 240 to 244):

	```
if poolRewardFixedCost < fixedCost {
		return poolRewardFixedCost, nil
	}

	return poolRewardFixedCost - fixedCost, nil
```

File: pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go [hunk 13]
```diff
@@ -144,46 +192,90 @@ func (t *Tracker) ApplyEpoch(epoch iotago.EpochIndex, committee *account.Account
...
+	if updatedPerformance.BlockIssuedCount == t.apiProvider.APIForSlot(block.ID().Index()).ProtocolParameters().RewardsParameters().ValidatorBlocksPerSlot {
+		// no need to store larger number and we can fit into uint8
+		updatedPerformance.BlockIssuedCount = t.apiProvider.APIForSlot(block.ID().Index()).ProtocolParameters().RewardsParameters().ValidatorBlocksPerSlot + 1
```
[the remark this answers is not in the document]
daria305: yes, but the plus 1 is needed so we keep the record if node exceeded the performance, I simplified it like this
```
if validatorPerformance.BlockIssuedCount < apiForSlot.ProtocolParameters().RewardsParameters().ValidatorBlocksPerSlot+1 {
        validatorPerformance.BlockIssuedCount++
}
```

pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go
```diff
@@ -110,25 +147,36 @@ func (t *Tracker) ApplyEpoch(epoch iotago.EpochIndex, committee *account.Account
...
 
 	committee.ForEach(func(accountID iotago.AccountID, pool *account.Pool) bool {
-		intermediateFactors := make([]uint64, 0, epochEndSlot+1-epochStartSlot)
+		validatorPerformances := make([]*model.ValidatorPerformance, 0, epochEndSlot+1-epochStartSlot)
 		for slot := epochStartSlot; slot <= epochEndSlot; slot++ {
-			performanceFactorStorage, err := t.performanceFactorsFunc(slot)
+			validatorSlotPerformances, err := t.validatorPerformancesFunc(slot)
 			if err != nil {
-				intermediateFactors = append(intermediateFactors, 0)
+				validatorPerformances = append(validatorPerformances, nil)
```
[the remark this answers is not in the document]
daria305: I am not sure if I understood, slots [epochStart, ...] doesn't match array indexes [0, 1,...] here I would need to add another index like i `while` loop or use map and both solution seem less clean

pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go
```diff
@@ -110,25 +147,36 @@ func (t *Tracker) ApplyEpoch(epoch iotago.EpochIndex, committee *account.Account
 	}
 
 	committee.ForEach(func(accountID iotago.AccountID, pool *account.Pool) bool {
-		intermediateFactors := make([]uint64, 0, epochEndSlot+1-epochStartSlot)
+		validatorPerformances := make([]*model.ValidatorPerformance, 0, epochEndSlot+1-epochStartSlot)
```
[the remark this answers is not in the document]
daria305: thx, I haven't noticed we have this not only in seconds

pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go
```diff
@@ -19,8 +22,9 @@ type Tracker struct {
 	poolStatsStore           *epochstore.Store[*model.PoolsStats]
 	committeeStore           *epochstore.Store[*account.Accounts]
 
-	performanceFactorsFunc func(slot iotago.SlotIndex) (*slotstore.Store[iotago.AccountID, uint64], error)
-	latestAppliedEpoch     iotago.EpochIndex
+	validatorPerformancesFunc func(slot iotago.SlotIndex) (*slotstore.Store[iotago.AccountID, *model.ValidatorPerformance], error)
+	latestPrunedEpoch         func() (iotago.EpochIndex, bool)
```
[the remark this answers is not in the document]
daria305: forgot to remove

pkg/protocol/sybilprotection/sybilprotectionv1/performance/snapshot.go
```diff
@@ -264,32 +269,31 @@ func (t *Tracker) exportPoolRewards(pWriter *utils.PositionedWriter, targetEpoch
 	if err := pWriter.WriteValue("pool rewards epoch count", epochCount, true); err != nil {
 		return ierrors.Wrap(err, "unable to write epoch count")
 	}
-	// TODO: make sure to adjust this loop if we add pruning later.
-	for epoch := targetEpoch; epoch > iotago.EpochIndex(0); epoch-- {
+
+	for epoch := targetEpoch; epoch > iotago.EpochIndex(lo.Max(0, int(targetEpoch)-365)); epoch-- {
```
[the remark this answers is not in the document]
daria305: Ok, for now it's just constant, I left todo with issue you mentioned

pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go
```diff
@@ -93,7 +126,11 @@ func (t *Tracker) ApplyEpoch(epoch iotago.EpochIndex, committee *account.Account
 	epochStartSlot := timeProvider.EpochStart(epoch)
 	epochEndSlot := timeProvider.EpochEnd(epoch)
 
-	profitMargin := calculateProfitMargin(committee.TotalValidatorStake(), committee.TotalStake())
+	profitMargin, err := t.calculateProfitMargin(committee.TotalValidatorStake(), committee.TotalStake(), epoch)
+	if err != nil {
+		panic(ierrors.Wrapf(err, "failed to calculate profit margin for epoch %d", epoch))
```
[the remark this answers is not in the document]
daria305: I propagated all panics except two panics in `HookSth` and in one place when we fail to reuse the same committee as it is already a fallback strategy, and it's really should never happen

pkg/protocol/sybilprotection/sybilprotectionv1/performance/rewards.go
oliviasaa: I thought that I had already commented this somewhere, but maybe I commented in the wrong place or something. Where is this `rewardsForAccountInEpoch.PoolRewards` being set? because the `poolReward` function returns the pool reward value after subtracting the fixed cost (see line 244) and then this check doesn't make sense.

oliviasaa: (I can't find where you update the `PoolRewards` struct)

oliviasaa: Also, I suggested (somewhere :p) to just not add any value for this single epoch to the validator rewards, instead of zeroing the rewards for the whole period

jonastheis (approved):

PR #330, opened 2023-09-05
Port peers management from hornet

muXxer:


components/p2p/component.go
```diff
@@ -101,108 +123,199 @@ func provide(c *dig.Container) error {
...
-		libp2pHost, err := golibp2p.New(
-			golibp2p.ListenAddrStrings(ParamsP2P.BindAddress),
-			golibp2p.Identity(nodePrivateKey),
-			golibp2p.ConnectionManager(connManager),
-			golibp2p.NATPortMap(),
+		// also see "defaults.go" in go-libp2p for the default values
+		createdHost, err := libp2p.New(
+			libp2p.ListenAddrStrings(ParamsP2P.BindMultiAddresses...),
+			libp2p.Identity(nodePrivateKey),
+			libp2p.Transport(tcp.NewTCPTransport),
```
karimodm, line 276: Why do we need this?

components/p2p/params.go
```diff
@@ -22,19 +27,26 @@ type ParametersP2P struct {
...
-	// KnownPeers defines the map of peers to be used as known peers.
-	KnownPeers string `usage:"map of peers that will be used as known peers"`
-	// BootstrapPeers defines the map of peers to be used as discovery for other peers.
-	BootstrapPeers string `usage:"map of peers to be used as discovery for other peers"`
+	// Defines the static peers this node should retain a connection to (CLI).
+	Peers []string `default:"" usage:"the static peers this node should retain a connection to (CLI)"`
+	// Defines the aliases of the static peers (must be the same length like CfgP2PPeers) (CLI).
+	PeerAliases []string `default:"" usage:"the aliases of the static peers (must be the same amount like \"p2p.peers\""`
+	// Defines the peers to be used as discovery for other peers (CLI).
+	BootstrapPeers []string `default:"" usage:"peers to be used as discovery for other peers (CLI)"`
```
karimodm, line 47: Do we need an additional parameter for this? Or we can just use Peers as bootstrap.

pkg/onchangemap/onchangemap.go
```diff
@@ -0,0 +1,194 @@
+package onchangemap
```
karimodm, line 1: Wouldn't this belong in hive?

karimodm (commented):
I've resolved a few conflicts with some changes I've made to the underlying PR. 
This looks good!
Only thing we miss would be handling peer adding and removing from the REST API.

components/p2p/params.go
muXxer: Good question.
I like to have another parameter tbh.
This way you can have BootstrapPeers defined by the default configuration, that don't get added as peers in every "peering.json" of every node, but still can be used for discovery.

components/p2p/component.go
muXxer: To prevent the usage of the other default transport mechanisms like udp/websocket/quic.
Our tooling like the node-docker-setup for example or the wiki explains to open the port for TCP only anyways.

karimodm (approved):

PR #344, opened 2023-09-15
Implicit account creation

### LINKED ISSUE #278: Implicit account creation mechanism
https://hackmd.io/@IF/Sy1ZPiJ2n

cyberphysic4l:
Implements implicit account creation https://hackmd.io/@IF/Sy1ZPiJ2n

implicit account addresses added in https://github.com/iotaledger/iota.go/pull/503

closes #278


pkg/protocol/engine/commitmentfilter/accountsfilter/commitmentfilter.go
```diff
@@ -155,12 +155,23 @@ func (c *CommitmentFilter) evaluateBlock(block *blocks.Block) {
 	switch signature := block.ProtocolBlock().Signature.(type) {
 	case *iotago.Ed25519Signature:
 		if !accountData.BlockIssuerKeys.Has(iotago.BlockIssuerKeyEd25519FromPublicKey(signature.PublicKey)) {
-			c.events.BlockFiltered.Trigger(&commitmentfilter.BlockFilteredEvent{
-				Block:  block,
-				Reason: ierrors.Wrapf(ErrInvalidSignature, "block issuer account %s does not have public key %s in slot %d", block.ProtocolBlock().IssuerID, signature.PublicKey, block.ProtocolBlock().SlotCommitmentID.Index()),
+			// if the block issuer does not have the public key in the slot commitment, check if it has an implicit account with the corresponding address
+			var hasAddress = false
+			accountData.BlockIssuerKeys.Range(func(element iotago.BlockIssuerKey) {
```
muXxer, line 160: Can't we stop the loop if we found the address? maybe we should add a return bool parameter to the `Range` function to abort it?

pkg/protocol/engine/ledger/ledger/ledger.go
```diff
@@ -437,15 +446,25 @@ func (l *Ledger) prepareAccountDiffs(accountDiffs map[iotago.AccountID]*model.Ac
...
+			accountDiff.BlockIssuerKeysAdded = createdOutput.Output().FeatureSet().BlockIssuer().BlockIssuerKeys
+			accountDiff.NewExpirySlot = createdOutput.Output().FeatureSet().BlockIssuer().ExpirySlot
+			if stakingFeature := createdOutput.Output().FeatureSet().Staking(); stakingFeature != nil {
+				accountDiff.ValidatorStakeChange = int64(stakingFeature.StakedAmount)
+				accountDiff.StakeEndEpochChange = int64(stakingFeature.EndEpoch)
+				accountDiff.FixedCostChange = int64(stakingFeature.FixedCost)
+			}
+		// for basic outputs (implicit accounts), get block issuer keys from the address in the unlock conditions.
+		case iotago.OutputBasic:
+			address, _ := createdOutput.Output().UnlockConditionSet().Address().Address.(*iotago.ImplicitAccountCreationAddress)
```
muXxer, line 463: Are we 100% sure this typecast will always work?

cyberphysic4l: Yes, a basic output will only ever end up in the createdAccounts map if it has this address unlock condition. Is it better to still check and panic if the type cast fails?

PhilippGackstatter: Added a comment there.

Issue #284, opened 2023-08-15
Harmonize serialization utils, methods on objects and more


jonastheis:
```[tasklist]
### Tasks
- [x] Currently we have and use multiple tools for serialization throughout the code base. We're using everything except `marshalutil` heavily. Therefore, we should try to remove dependencies to `marshalutil` to decrease the amount of code and thus error surface.
- [x] Go away from manual `binary.Read` and `binary.Write` in favor of utils provided in `stream` package
- [x] Remove objects' `.FromBytes()` method in favor of stateless `ObjectFromBytes()` function to avoid quirks with Go generics
- [x] Remove `storable` package from hive.go
- [x] Remove `PositionedWriter` and replace with `stream.WriteCollection`
- [x] introduce/generate wrapper to conveniently store basic types with our `kvstore.TypedKVStore`
- [x] Revisit `time.Time` serialization (seconds vs nanoseconds)
- [x] Introduce better way to serialize and deserialize `BlockIssuerKeys`
```

jonastheis:
In [this commit](https://github.com/iotaledger/iota.go/commit/b61e53dbf4890fe5a9ea7606fe6e4d6904eba9ce ) the serialization of `BlockIssuerKeys` has been improved. However, it showcases one problem well:
- many objects can be read/written in the context of a `reader` (e.g. for a snapshot). While the total object size might be variable, the object is still able to deserialize itself without the need to always prepend a size (just by following the order and reading the individual fields from the reader). This becomes important when we are writing lots of objects, for example outputs, for a ledger state snapshot. 
- When and how exactly do we utilize `serix`? Some objects are sometimes serialized with `serix` but other times not. Especially when reading from a `reader` currently `serix` does not support that.

PR #351, opened 2023-09-18
Encode each attestation by version when sending over the wire

alexsporn:


pkg/network/protocols/core/protocol.go
```diff
@@ -200,10 +202,37 @@ func (p *Protocol) onAttestations(commitmentBytes []byte, attestationsBytes []by
...
-	// TODO: there could be multiple versions of attestations in the same packet
-	if _, err := lo.PanicOnErr(p.apiProvider.APIForVersion(iotago.Version(commitmentBytes[0]))).Decode(attestationsBytes, &attestations, serix.WithValidation()); err != nil {
-		p.Events.Error.Trigger(ierrors.Wrap(err, "failed to deserialize attestations"), id)
+	if len(attestationsBytes) < 4 {
+		p.Events.Error.Trigger(ierrors.Errorf("failed to deserialize attestations, invalid attestation count"), id)
+
+		return
+	}
+
+	attestationCount := binary.LittleEndian.Uint32(attestationsBytes[0:4])
```
jonastheis, line 211: Checking for the constant, using `binary.LittleEndian` package and having a manual `readOffset` seems to be a bit cumbersome. I think this would be much nicer with the `stream` package, here specifically `ReadCollection`

pkg/network/protocols/core/protocol.go
```diff
@@ -73,16 +75,16 @@ func (p *Protocol) SendSlotCommitment(cm *model.Commitment, to ...peer.ID) {
...
-	if len(attestations) > 0 {
-		// TODO: there are multiple attestations potentially spanning multiple epochs/versions, we need to use the correct API for each one
-		iotagoAPI = lo.PanicOnErr(p.apiProvider.APIForVersion(attestations[0].ProtocolVersion))
-	} else {
-		iotagoAPI = p.apiProvider.APIForSlot(cm.Index()) // we need an api to serialize empty slices as well
+	encodedAttestations := marshalutil.New()
+	encodedAttestations.WriteUint32(uint32(len(attestations)))
+	for _, att := range attestations {
+		iotagoAPI := lo.PanicOnErr(p.apiProvider.APIForVersion(att.ProtocolVersion))
+		encodedAttestations.WriteBytes(lo.PanicOnErr(iotagoAPI.Encode(att)))
```
jonastheis, line 82: Instead of marshalutil we could use `stream.WriteCollection`

alexsporn: a bytes buffer does not implement WriteSeeker, so could not use it here

jonastheis: Ok then let's implement something like this https://stackoverflow.com/a/73679110 in the context of #284 so that we're able to use a bytes buffer with the stream package and replace this later

jonastheis (approved):

PR #375, opened 2023-09-27
Use OutputBuilders from iota.go in transaction test framework

jkrvivian:
iota.go PR: https://github.com/iotaledger/iota.go/pull/532


pkg/testsuite/transactions_framework.go
```diff
@@ -342,159 +329,107 @@ func (t *TransactionFramework) DefaultAddress() iotago.Address {
...
 }
 
-func WithBlockIssuerKeys(keys iotago.BlockIssuerKeys) options.Option[iotago.AccountOutput] {
-	return func(accountOutput *iotago.AccountOutput) {
-		blockIssuer := accountOutput.FeatureSet().BlockIssuer()
-		if blockIssuer == nil {
-			panic("cannot set block issuer keys to account without BlockIssuer feature")
-		}
-		blockIssuer.BlockIssuerKeys = keys
+func AddBlockIssuerKey(key iotago.BlockIssuerKey) options.Option[builder.AccountOutputBuilder] {
```
daria305, line 376: Shouldn't we use name  WithBlockIssuerKey to keep the convention? Then you need to update WithBlockIssuerKeys. Maybe WithSetBlockIssuerKeys, as both methods differ, becuase socond replace the keys, not add many

pkg/testsuite/transactions_framework.go
```diff
@@ -342,159 +329,107 @@ func (t *TransactionFramework) DefaultAddress() iotago.Address {
...
+func WithAccountConditions(conditions iotago.AccountOutputUnlockConditions) options.Option[builder.AccountOutputBuilder] {
+	return func(accountBuilder *builder.AccountOutputBuilder) {
+		for _, condition := range conditions.MustSet() {
+			switch condition.Type() {
+			case iotago.UnlockConditionStateControllerAddress:
+				accountBuilder.StateController(condition.(*iotago.StateControllerAddressUnlockCondition).Address)
+			case iotago.UnlockConditionGovernorAddress:
+				accountBuilder.Governor(condition.(*iotago.GovernorAddressUnlockCondition).Address)
+			}
+		}
```
daria305, line 426: just checking if WithAccountIncreasedFoundryCounter, WithAccountIncreasedFoundryCounter and WithAccountImmutableFeatures were skipped on purpose?

daria305 (changes requested):

pkg/testsuite/transactions_framework.go
jkrvivian: Maybe the other way around, only this is adding a new element to the original value while others are replacing it. So I'll keep `WithBlockIssuerKeys` and rename `AddBlockIssuerKey` to `WithAddBlockIssuerKey`.

pkg/testsuite/transactions_framework.go
jkrvivian: I removed it since it's not used during refactoring. :stuck_out_tongue: 
Will add it back.

daria305 (approved):

alexsporn (approved):

PR #392, opened 2023-10-01
Fix Delegation Stake Change Calculation

PhilippGackstatter:
Fix the delegation stake calculation upon slot commitment.

There were two issues:
- Using the Delegation ID of a Delegation Output directly, which may be zero, so it must be derived from the output ID first.
- A Delegation Output that was transitioned to Delayed Claiming was still counted as staking to that validator. However, that should not be the case. Consider the following case:
    - Create a Delegation Output `A` with a `Amount = Delegated Amount = 1000` and `Validator Address = 0x123`. That increases `0x123`'s delegated stake by 1000. Now we transition to delayed claiming and keep `Delegated Amount` the same, but remove all the value except for the required storage deposit, say, to `Amount = 10`. In the same transition we can create a new Delegation Output `B` with `Amount = Delegated Amount = 990` to the same validator. Now we've effectively delegated to the same validator twice with the same `990` tokens, and that stake is active as long as we don't consume `A`. We could repeat this procedure for a long time. Hence, as soon as we transition to delayed claiming, we need to subtract the delegated stake from the validator, so no Delegation Output can ever delegate the same tokens twice.

### DIFF
--- a/pkg/protocol/engine/ledger/ledger/ledger.go
+++ b/pkg/protocol/engine/ledger/ledger/ledger.go
@@ -473,7 +473,7 @@ func (l *Ledger) processCreatedAndConsumedAccountOutputs(stateDiff mempool.State
 	consumedAccounts = make(map[iotago.AccountID]*utxoledger.Output)
 	destroyedAccounts = ds.NewSet[iotago.AccountID]()
 
-	createdAccountDelegation := make(map[iotago.ChainID]*iotago.DelegationOutput)
+	newAccountDelegation := make(map[iotago.ChainID]*iotago.DelegationOutput)
 
 	stateDiff.CreatedStates().ForEachKey(func(outputID iotago.OutputID) bool {
 		createdOutput, errOutput := l.Output(outputID)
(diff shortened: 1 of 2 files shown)


pkg/protocol/engine/ledger/ledger/ledger.go
```diff
@@ -499,11 +499,17 @@ func (l *Ledger) processCreatedAndConsumedAccountOutputs(stateDiff mempool.State
 			}
 
 			createdAccounts[accountID] = createdOutput
-
 		case iotago.OutputDelegation:
-			// the delegation output was created => determine later if we need to add the stake to the validator
-			delegation, _ := createdOutput.Output().(*iotago.DelegationOutput)
-			createdAccountDelegation[delegation.DelegationID] = delegation
+			// The Delegation Output output was created or transitioned => determine later if we need to add the stake to the validator.
```
muXxer, line 503:
```suggestion
			// The DelegationOutput was created or transitioned => determine later if we need to add the stake to the validator.
```

muXxer (approved):

PR #394, opened 2023-10-02
Add BIC to throughput and add 1 to quantum

cyberphysic4l:
This PR makes two simple but important changes to throughput in the scheduler. 

1. We include the Block Issuance Credits (BIC) in the throughput for a block issuer in the DRR scheduler. Previously, it was only based on the stored and potential Mana on the account output of the issuer, but now we also include the BIC in this value. BIC is locked to the account to an even greater extent than the stored and potential Mana so it makes sense to include it for the throughput allowance. For any user that may wish to pre-allot their Mana, this means they do not need to choose between allotting the Mana and using it for their throughput, if they allot it they get it towards their throughput as well now.
2. We now always have a quantum of at least one for a block issuer, even if they have zero Mana. The storage deposit on the account output is sufficient spam protection, so this does not amount to giving free throughput. Although the issuers are "zero-mana" issuers, they are not free to create. The schedule will still operate in the same fashion in times of congestion, prioritising issuers with higher Mana.


piotrm50 (changes requested):
MinMana should be removed from ProtocolParameters since we always assume that everyone has at least MinMana.

piotrm50 (approved):

PR #399, opened 2023-10-03
UTXO-level MCA threshold for unincluded conflicts 

### LINKED ISSUE #124: UTXO-level TSC condition

karimodm:
Fixes #124 

Introduce an MaxCommittableAge delay to evict Transaction and Outputs from the Mempool.
This delay is needed to objectively determine if a block is spending or promoting a rejected or pending conflict that has fallen below the commitment point. In fact, a race exists between a node that attempts to book a transaction having committed the conflict and a node that  didn't yet: the former won't consider the transaction solid as the input has been already evicted from the mempool, while the latter will book the transaction under the rejected or pending conflict; leading to a subjective discrepancy in what block has been booked.
The same race exists on Tangle level: whenever a block is promoting a conflict that is about to be committed.

The solution lies then in enforcing that every block or transaction can only be booked under a conflict that is still available with respect to the referenced commit.
For this we have to delay eviction of transactions and outputs until a block cannot possibly belong to a slot and still include a commitment in which the promoted or spent conflict was uncommitted: such boundary is MaxCommittableAge. 

- Orphaned conflicts are annotated with the commitment slot that excluded them.
- Such slot gets propagated to the UTXO future cone.
- Whenever a block gets booked, a transaction is considered orphaned if it tries to spend a conflict that has been orphaned with an index smaller than what the block commits to. If so, the attachment block is marked as invalid.
- Similarly, a block does not inherit conflicts carrying an orphanage index smaller than what the block commits to.


piotrm50 (approved):

hmoog (approved):
LGTM

PR #400, opened 2023-10-04
Extract block issuer from node in tests and components

cyberphysic4l:
This PR replaces the blockfactory package with a the blockhandler package, which simply accepts pre-signed blocks but does not modify them or create new ones.

We also add a blockissuer module to the mock package of the testsuite which is mostly copied over from the blockfactory package, and we extract this blockissuer module from the mock node in the tests. Validator nodes still own a block issuer for issuing their validation blocks only, but for any other block issuance, an external block issuer must be used in all tests now. This functionality will later be extended to include a more complete user/wallet style interaction in the tests to more accurately reflect the user/node flow rather than the entirely node-centric testing prior to this PR.


components/restapi/core/component.go
```diff
@@ -39,14 +39,6 @@ const (
 	// GET returns block metadata.
 	RouteBlockMetadata = "/blocks/:" + restapipkg.ParameterBlockID + "/metadata"
 
-	// RouteBlocks is the route for creating new blocks.
-	// POST creates a single new block and returns the new block ID.
-	// The block is parsed based on the given type in the request "Content-Type" header.
-	// By providing only the protocolVersion and payload transaction user can POST a transaction.
-	// MIMEApplicationJSON => json.
-	// MIMEVendorIOTASerializer => bytes.
-	RouteBlocks = "/blocks"
```
alexsporn, line 48: We still need this route, but it should pass the blocks to the blockhandler. Else there is no way for clients to send pre-signed blocks.

components/restapi/core/blocks.go
```diff
@@ -61,3 +64,67 @@ func blockIssuance(_ echo.Context) (*apimodels.IssuanceBlockHeaderResponse, erro
...
+
+	switch mimeType {
+	case echo.MIMEApplicationJSON:
+		// Do not validate here, the parents might need to be set
+		if err := deps.Protocol.CurrentAPI().JSONDecode(bytes, iotaBlock); err != nil {
+			return nil, ierrors.Wrapf(httpserver.ErrInvalidParameter, "invalid block, error: %w", err)
+		}
+
+	case httpserver.MIMEApplicationVendorIOTASerializerV2:
+		version, _, err := iotago.VersionFromBytes(bytes)
```
alexsporn, line 94: Switch this to use `ProtocolBlockFromBytes(deps.Protocol)`

components/restapi/core/blocks.go
```diff
@@ -61,3 +64,67 @@ func blockIssuance(_ echo.Context) (*apimodels.IssuanceBlockHeaderResponse, erro
...
+
+	bytes, err := io.ReadAll(c.Request().Body)
+	if err != nil {
+		return nil, ierrors.Wrapf(httpserver.ErrInvalidParameter, "invalid block, error: %w", err)
+	}
+
+	switch mimeType {
+	case echo.MIMEApplicationJSON:
+		// Do not validate here, the parents might need to be set
+		if err := deps.Protocol.CurrentAPI().JSONDecode(bytes, iotaBlock); err != nil {
```
alexsporn, line 89: Add `serix.WithValidation()` and remove comment, since we are now expecting a fully valid block here

components/validator/issuer.go
```diff
@@ -35,17 +36,23 @@ func issueValidatorBlock(ctx context.Context) {
...
-		blockfactory.WithHighestSupportedVersion(deps.Protocol.LatestAPI().Version()),
-	)
+	// create the validation block here using the validation block builder from iota.go
+
+	validationBlock, err := builder.NewValidationBlockBuilder(deps.Protocol.CurrentAPI()).
+		IssuingTime(blockIssuingTime).
+		SlotCommitmentID(latestCommitment.ID()).
+		ProtocolParametersHash(protocolParametersHash).
+		HighestSupportedVersion(deps.Protocol.LatestAPI().Version()).
+		Build()
```
alexsporn, line 46: I am missing `LatestFinalizedSlot`, the issuer and the signing of the block here

alexsporn: Also tipselection is missing

cyberphysic4l: Added basic tip selection here as I don't think we have special validator tip selection yet.

alexsporn (approved):

PR #408, opened 2023-10-09
Implement candidate registration and TopStakers seat manager.

piotrm50:
This PR adds support for top staker committee selection as well as registering candidates for next epoch committee selection.

Candidates are registered whenever they issue a single block before Epoch Nearing Threshold in the previous epoch.


pkg/protocol/commitment_verifier.go
```diff
@@ -20,7 +20,7 @@ type CommitmentVerifier struct {
 }
 
 func NewCommitmentVerifier(mainEngine *engine.Engine, lastCommonCommitmentBeforeFork *model.Commitment) *CommitmentVerifier {
-	committeeAtForkingPoint := mainEngine.SybilProtection.SeatManager().Committee(lastCommonCommitmentBeforeFork.Slot()).Accounts().IDs()
+	committeeAtForkingPoint := lo.Return1(mainEngine.SybilProtection.SeatManager().CommitteeInSlot(lastCommonCommitmentBeforeFork.Slot())).Accounts().IDs()
```
alexsporn: `CommitteeInSlot(...)` can return nil now, so this looks dangerous if you ignore the bool return variable. (This pattern is repeated everywhere where `Commitee(...)` was used before)

pkg/protocol/sybilprotection/seatmanager/topstakers/topstakers.go
```diff
@@ -0,0 +1,260 @@
...
+
+// SeatManager is a sybil protection module for the engine that manages the weights of actors according to their stake.
+type SeatManager struct {
+	apiProvider iotago.APIProvider
+	events      *seatmanager.Events
+
+	committeeStore   *epochstore.Store[*account.Accounts]
+	onlineCommittee  ds.Set[account.SeatIndex]
+	inactivityQueue  timed.PriorityQueue[account.SeatIndex]
+	lastActivities   *shrinkingmap.ShrinkingMap[account.SeatIndex, time.Time]
```
alexsporn: should we refactor the activity tracking into an own file? This is kind of duplicated code between poa and topstakers now

pkg/protocol/sybilprotection/seatmanager/topstakers/topstakers.go
```diff
@@ -0,0 +1,260 @@
...
+	sort.Slice(candidatePools, func(i, j int) bool {
+		if candidatePools[i].pool.PoolStake != candidatePools[j].pool.PoolStake {
+			return candidatePools[i].pool.PoolStake > candidatePools[j].pool.PoolStake
+		}
+
+		if candidatePools[i].pool.ValidatorStake != candidatePools[j].pool.ValidatorStake {
+			return candidatePools[i].pool.ValidatorStake > candidatePools[j].pool.ValidatorStake
+		}
+
+		// TODO: add more tie-breaking
```
alexsporn: maybe prioritize candidates that haven't added an endepoch to its staking feature?

pkg/protocol/sybilprotection/seatmanager/topstakers/topstakers.go
```diff
@@ -0,0 +1,260 @@
...
+
+		// two candidates never have the same account ID because they come in a map
+		return bytes.Compare(candidatePools[i].accountID[:], candidatePools[j].accountID[:]) > 0
+	})
+
+	// Create new Accounts instance that only included validators selected to be part of the committee.
+	accounts := account.NewAccounts()
+	for _, candidatePool := range candidatePools[:s.optsSeatCount] {
```
alexsporn: you better check we have more candidates than seats here

pkg/protocol/sybilprotection/seatmanager/topstakers/topstakers.go
```diff
@@ -0,0 +1,260 @@
...
+	defer s.activityMutex.RUnlock()
+
+	return s.onlineCommittee
+}
+
+func (s *SeatManager) SeatCount() int {
+	s.committeeMutex.RLock()
+	defer s.committeeMutex.RUnlock()
+
+	return int(s.optsSeatCount)
```
alexsporn, line 154: same as before, what happens if we have less candidates?

pkg/protocol/sybilprotection/seatmanager/topstakers/topstakers.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	return nil
+}
+
+func (s *SeatManager) SetCommittee(epoch iotago.EpochIndex, validators *account.Accounts) error {
+	s.committeeMutex.Lock()
+	defer s.committeeMutex.Unlock()
+
+	if validators.Size() != int(s.optsSeatCount) {
```
alexsporn, line 194: Again here, what would happen if we don't have enough candidates?

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
```diff
@@ -108,8 +98,42 @@ func (o *SybilProtection) Shutdown() {
 	o.TriggerStopped()
 }
 
-func (o *SybilProtection) TrackValidationBlock(block *blocks.Block) {
-	o.performanceTracker.TrackValidationBlock(block)
+func (o *SybilProtection) TrackBlock(block *blocks.Block) {
+	accountData, exists, err := o.ledger.Account(block.ProtocolBlock().IssuerID, block.SlotCommitmentID().Slot())
```
alexsporn, line 109: why fetch the accountData before even checking if it is a candidacy announcement? Check first if we need to do the work or not.

pkg/storage/prunable/prunable.go
```diff
@@ -167,7 +171,7 @@ func (p *Prunable) Rollback(targetSlot iotago.SlotIndex) error {
 			}
 
 			if shouldRollback {
-				if err := p.committee.DeleteEpoch(epoch); err != nil {
+				if err = p.committee.DeleteEpoch(epoch); err != nil {
```
alexsporn: why this change? it is a local var captured only for the if.

pkg/storage/prunable/prunable_slot.go
```diff
@@ -52,6 +53,11 @@ func (p *Prunable) RootBlocks(slot iotago.SlotIndex) (*slotstore.Store[iotago.Bl
 	), nil
 }
 
+func (p *Prunable) CommitteeCandidates(epoch iotago.EpochIndex) (kvstore.KVStore, error) {
+	// Slot 0 does not exist, but we need to add it to the prefix to avoid random clashes with other keys.
+	return p.prunableSlotStore.Get(epoch, byteutils.ConcatBytes(iotago.SlotIndex(0).MustBytes(), kvstore.Realm{epochPrefixCommitteeCandidates}))
```
alexsporn: this is hacky, better use the epoch startSlot so it at least matches the other stores.

alexsporn (changes requested):
This kind of introduces a lot of new TODOs

pkg/protocol/sybilprotection/seatmanager/topstakers/topstakers.go
piotrm50: I added some more things there. It wasn't there initially because the data wasn't available. Now I changed what RotateCommittee receives which imo makes it clearer.

pkg/protocol/commitment_verifier.go
piotrm50: We need to discuss what to do if the committee does not exists for some reason, because this should not happen, but if it does then it's pretty severe.

pkg/protocol/sybilprotection/seatmanager/topstakers/topstakers.go
piotrm50: This should always return the same number as this is the number of validators we will have at all times.

pkg/protocol/sybilprotection/seatmanager/topstakers/topstakers.go
piotrm50: This method is only used for re-using the committee for the previous slot, so here we require the exact number of validators as we use them as they come.

pkg/protocol/sybilprotection/seatmanager/topstakers/topstakers.go
piotrm50: Good catch! Changed it to re-use the committee from  the previous slot in this case as discussed before.

alexsporn (approved):

alexsporn (approved):

PR #419, opened 2023-10-09
Remove `Output(…)` from ledger interface in favour of always calling `OutputOrSpent(…)`

alexsporn:


muXxer:
Or maybe we do that logic 
```
	if spent != nil {
		output = spent.Output()
	}
```
in the former `Output()`?

alexsporn:
> Or maybe we do that logic
> 
> ```
> 	if spent != nil {
> 		output = spent.Output()
> 	}
> ```
> 
> in the former `ledger.Output(...)`?

Thing is, the `ledger.Output` is used internally for processing state diffs and those should not do the spent magic.
We can refactor the internal call to not use this, but the question is, do we really need 2 functions in the ledger interface?

alexsporn:
Will implement differently in a separate PR

PR #423, opened 2023-10-11
Store proofs per output that the outputID corresponds to its content

alexsporn:
Depends on https://github.com/iotaledger/iota.go/pull/563


pkg/protocol/engine/ledger/ledger/ledger.go
```diff
@@ -610,14 +611,13 @@ func (l *Ledger) processStateDiffTransactions(stateDiff mempool.StateDiff) (spen
 		{
 			// input side
 			for _, inputRef := range inputRefs {
-				inputState, outputErr := l.Output(inputRef.OutputID())
-				if outputErr != nil {
+				stateWithMetadata, stateError := l.memPool.StateMetadata(inputRef)
```
muXxer, line 614: why do we use the mempool directly here now?

pkg/protocol/engine/utxoledger/output.go
```diff
@@ -89,10 +93,29 @@ func (o *Output) Output() iotago.Output {
 	return o.output
 }
 
+func (o *Output) OutputIDProof() *iotago.OutputIDProof {
+	o.proofOnce.Do(func() {
+		if o.outputProof == nil {
+			api := o.apiProvider.APIForSlot(o.blockID.Slot())
```
muXxer, line 99: why the slot of the blockID? the output itself has the slot already.

muXxer (approved):

pkg/protocol/engine/ledger/ledger/ledger.go
alexsporn: because we are applying the state from the mempool actually

PR #425, opened 2023-10-11
Resolve TODOs in test suite

### LINKED ISSUE #251: TODOs in testsuite
* [ ] `pkg/testsuite/storage_accountdiffs.go`
	* Double check if `cmp` is enough for the test
* [ ] `pkg/testsuite/testsuite.go`
	* Remove `WithLivenessThreshold`, `WithMinCommittableAge`, common parameters should be used

jonastheis:
- Refactor protocol parameter definition via testsuite to be more explicit and using iota.go ProtocolParameter options instead of individual parameters
- remove `cmp.Equals` and remove package in favor of `assert.Equals`

Fixes #251


alexsporn (approved):
I would revert the changes in `pkg/testsuite/mock/blockissuer.go`

PR #430, opened 2023-10-13
Add Protocol Upgrade test

### LINKED ISSUE #189: Protocol upgrades: test switching to a new protocol version
Write end-to-end test to test the entire functionality of the `UpgradeOrchestrator` in different edge case scenarios.
- attestations over upgrade boundaries (with different versions)
-

jonastheis:
This extends `Test_Upgrade_Signaling` to not only test for the signaling of votes but also to continue and test for the upgrade to actually happen in the voted upon epoch. Nodes that didn't upgrade can't process the blocks while nodes that upgraded will only accept blocks with the new version from the upgrade epoch. 

Other changes:
- rename `CurrentAPI` to `CommittedAPI` in `APIProvider`
- introduce filter for correct protocol version to filter
- introduce filter for correct protocol version to process commitments

Depends on https://github.com/iotaledger/iota.go/pull/567, https://github.com/iotaledger/inx-app/pull/90

Fixes #189


pkg/testsuite/mock/blockissuer.go
```diff
@@ -99,9 +99,11 @@ func (i *BlockIssuer) CreateValidationBlock(ctx context.Context, alias string, i
 		blockParams.BlockHeader.IssuingTime = &issuingTime
 	}
 
+	currentAPI := node.Protocol.APIForTime(*blockParams.BlockHeader.IssuingTime)
```
alexsporn: What about the `ProtocolVersion` in the `blockParams.BlockHeader`?

alexsporn (approved):

PR #440, opened 2023-10-17
Scheduler todos

### LINKED ISSUE #253: TODOs in scheduler
* [x] In `dropTail`, extract find longest mana-scaled queue function to util, 
        `pkg/protocol/engine/congestioncontrol/scheduler/drr/drrbuffer.go`
* [x] Define max block work instead of using MaxBlockSize
        `pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go`
* [x] Define work function for Block
        `pkg/protocol/engine/blocks/block.go`

cyberphysic4l:
closes #253 

Adds an optimisation of removing empty queues of issuers with max deficit. This saves us iterating over the queues of inactive issuers but we still get track of their deficit in the deficits map.


pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
```diff
@@ -570,8 +564,16 @@ func (s *Scheduler) removeIssuer(issuerID iotago.AccountID, err error) {
...
 
 	s.deficits.Delete(issuerID)
+	s.basicBuffer.RemoveIssuerQueue(issuerID)
+}
 
-	s.basicBuffer.RemoveIssuer(issuerID)
+func (s *Scheduler) getOrCreateIssuer(accountID iotago.AccountID) *IssuerQueue {
+	issuerQueue, created := s.basicBuffer.GetOrCreateIssuerQueue(accountID)
+	if created {
+		s.deficits.Set(accountID, 0)
```
piotrm50: If the idea is to remove queue once the issuer reaches MaxDeficit, then the queue does not exist, so if the issuer issues a block and it reaches the scheduler, we call `getOrCreateIssuer` which creates a new queue and sets deficit to 0. Or am I missing something? We should check if the deficit exists and set it to 0 if it doesn't.

piotrm50 (changes requested):

pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
cyberphysic4l: Good catch, you are correct! This would have meant issuer's deficits going to 0 every time their queue went empty :o
Updating this now to use GetOrCreate method of the shrinkingmap to set the deficit only if it doesn't exist.

piotrm50 (approved):

PR #441, opened 2023-10-18
Feat: Reset logic for engine to roll-back in memory.

hmoog:
This PR introduces a mechanism to Reset an engine to its clean state after the last commitment, without having to restart the entire engine.

This PR only addresses the parts of the code that are relevant for the commitments as other things will eventually get evicted anyway, but we should maybe revisit the TODOs that I left in the code in a separate PR, once this is no longer a blocker.


karimodm (approved):
Seems to work stably when trying to catch up with tens of thousands of commits.

pkg/protocol/engine/accounts/accountsledger/manager.go
```diff
@@ -355,6 +355,33 @@ func (m *Manager) AddAccount(output *utxoledger.Output, blockIssuanceCredits iot
...
+
+	versionSignalsToDelete := make([]iotago.SlotIndex, 0)
+	m.latestSupportedVersionSignals.ForEach(func(slot iotago.SlotIndex, _ *shrinkingmap.ShrinkingMap[iotago.AccountID, *model.SignaledBlock]) {
+		if slot > m.latestCommittedSlot {
+			versionSignalsToDelete = append(versionSignalsToDelete, slot)
+		}
+	})
+
+	for _, slot := range blockBurnsToDelete {
+		m.blockBurns.Delete(slot)
```
karimodm: Why not deleting them straight away instead of collecting them first?

pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -308,6 +309,31 @@ func (m *Manager) Rollback(targetSlot iotago.SlotIndex) error {
...
+	})
+
+	pendingAttestationsToClear := make([]iotago.SlotIndex, 0)
+	m.pendingAttestations.ForEach(func(slot iotago.SlotIndex, _ *shrinkingmap.ShrinkingMap[iotago.AccountID, *iotago.Attestation]) {
+		if slot > m.lastCommittedSlot {
+			pendingAttestationsToClear = append(pendingAttestationsToClear, slot)
+		}
+	})
+
+	for _, slot := range futureAttestationsToClear {
```
karimodm: Why not deleting straight away?

pkg/protocol/engine/clock/blocktime/relativetime.go
```diff
@@ -81,6 +81,17 @@ func (c *RelativeTime) Advance(newTime time.Time) (updated bool) {
 	return true
 }
 
+// Reset resets the time value to the given time (resetting monotonicity of the relative time).
+func (c *RelativeTime) Reset(newTime time.Time) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
```
karimodm, line 87: Why do you lock here and not in other components' Resets?

pkg/protocol/engine/eviction/state.go
```diff
@@ -323,6 +324,23 @@ func (s *State) PopulateFromStorage(latestCommitmentSlot iotago.SlotIndex) {
...
+			storage.ForEach(func(blockID iotago.BlockID, commitmentID iotago.CommitmentID) bool {
+				rootBlocksToClear = append(rootBlocksToClear, blockID)
+
+				return true
+			})
+		}
+	})
+
+	for _, blockID := range rootBlocksToClear {
+		s.RemoveRootBlock(blockID)
```
karimodm: remove on collection?

pkg/protocol/engine/mempool/v1/mempool.go
```diff
@@ -197,6 +197,42 @@ func (m *MemPool[VoteRank]) stateDiff(slot iotago.SlotIndex) (*StateDiff, error)
 	return lo.Return1(m.stateDiffs.GetOrCreate(slot, func() *StateDiff { return NewStateDiff(slot, kv) })), nil
 }
 
+// Reset resets the component to a clean state as if it was created at the last commitment.
+func (m *MemPool[VoteRank]) Reset() {
+	stateDiffsToDelete := make([]iotago.SlotIndex, 0)
```
karimodm: same

pkg/protocol/engine/notarization/slotnotarization/slotmutations.go
```diff
@@ -61,15 +61,20 @@ func (m *SlotMutations) Evict(index iotago.SlotIndex) error {
...
+	m.acceptedBlocksBySlot.ForEachKey(func(slot iotago.SlotIndex) bool {
+		if slot > m.latestCommittedIndex {
+			slotsToReset = append(slotsToReset, slot)
+		}
+
+		return true
+	})
+
+	for _, slot := range slotsToReset {
+		m.acceptedBlocksBySlot.Delete(slot)
```
karimodm: delete directly

pkg/protocol/engine/engine.go
```diff
@@ -216,6 +216,41 @@ func New(
...
 }
 
+func (e *Engine) ProcessBlockFromPeer(block *model.Block, source peer.ID) {
+	e.Filter.ProcessReceivedBlock(block, source)
+	e.Events.BlockProcessed.Trigger(block.ID())
+}
+
+// Reset resets the component to a clean state as if it was created at the last commitment.
+func (e *Engine) Reset() {
+	e.Workers.WaitChildren()
```
karimodm: Because of the lack of locking on all the `Reset` function you are probably assuming that no task will be submitted to any component after this `e.Workers.WaitChildren()`.
How are we 100% sure of that?

karimodm (changes requested):

pkg/protocol/engine/accounts/accountsledger/manager.go
hmoog: Because the foreach grabs a read lock which prevents us from writing to the collection in the same step.

pkg/protocol/engine/attestation/slotattestation/manager.go
hmoog: Because the foreach grabs a read lock which prevents us from writing to the collection in the same step.

pkg/protocol/engine/clock/blocktime/relativetime.go
hmoog: The "RelativeTime" is not a "component" but a self contained data structure that currently lacks the ability to just "Set" a specific time in the past. So the "Reset" method might be used outside the scope of a reset call for the engine components simply to reset the time to a specific value that is lower than the current time, but if you think its unnecessary we can also remove it.

pkg/protocol/engine/eviction/state.go
hmoog: Foreach grabs a read lock.

pkg/protocol/engine/mempool/v1/mempool.go
hmoog: Foreach grabs a read lock.

pkg/protocol/engine/notarization/slotnotarization/slotmutations.go
hmoog: Foreach grabs a read lock.

pkg/protocol/engine/engine.go
hmoog: You have to make sure that access to the engine is blocked from the outside, inside you don't have any guarantees, which is something we anyway need to ensure to not accidentaly feed old blocks to the engine after the reset.

Putting a mutex there would make it thread safe for concurrent use but it would not get rid of this requirement, so I think it might be better to put that mental burden on the dev to make sure its correct rather than assuming correctness through the presence of a mutex and making the dev ignore this requirement.

pkg/protocol/engine/accounts/accountsledger/manager.go
karimodm: It does not: it only holds the lock to create a clone of the map to iterate over, but the lock is released by the time the callback is called. The ForEachKey was designed exactly to be able to support something of this kind.

pkg/protocol/engine/clock/blocktime/relativetime.go
karimodm: I still find this discrepancy a bit odd. The Relativetime is still a subcomponent of the Clock, which is an engine-specific component. It is kinda odd that no component holds a lock while resetting but this does, either all have or none, no?

pkg/core/buffer/unsolid_commitment_buffer.go
```diff
@@ -136,3 +136,9 @@ func (u *UnsolidCommitmentBuffer[V]) GetValuesAndEvict(commitmentID iotago.Commi
 
 	return values
 }
+
+// Reset resets the component to a clean state as if it was created at the last commitment.
+func (u *UnsolidCommitmentBuffer[V]) Reset() {
+	u.blockBuffers.Clear()
+	u.commitmentBuffer.Each(func(key iotago.CommitmentID, _ types.Empty) { u.commitmentBuffer.Remove(key) })
```
piotrm50, line 146: Lock `commitmentBufferMutex` or add a comment as to why it's not locked.

pkg/protocol/engine/accounts/accountsledger/manager.go
```diff
@@ -355,6 +355,33 @@ func (m *Manager) AddAccount(output *utxoledger.Output, blockIssuanceCredits iot
 	return nil
 }
 
+// Reset resets the component to a clean state as if it was created at the last commitment.
+func (m *Manager) Reset() {
+	blockBurnsToDelete := make([]iotago.SlotIndex, 0)
+	m.blockBurns.ForEachKey(func(slot iotago.SlotIndex) bool {
+		if slot > m.latestCommittedSlot {
```
piotrm50: why only if slot >  m.latestCommittedSlot? Anything below is already evicted so this check is not needed. I think you can simply call `.Clear()` on this one.

pkg/protocol/engine/accounts/accountsledger/manager.go
```diff
@@ -355,6 +355,33 @@ func (m *Manager) AddAccount(output *utxoledger.Output, blockIssuanceCredits iot
...
+			blockBurnsToDelete = append(blockBurnsToDelete, slot)
+		}
+
+		return true
+	})
+
+	versionSignalsToDelete := make([]iotago.SlotIndex, 0)
+	m.latestSupportedVersionSignals.ForEach(func(slot iotago.SlotIndex, _ *shrinkingmap.ShrinkingMap[iotago.AccountID, *model.SignaledBlock]) {
+		if slot > m.latestCommittedSlot {
```
piotrm50: Same here: anything below latest committed slot is evicted anyway, so you can just call `.Clean()`.

pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -319,6 +320,31 @@ func (m *Manager) Rollback(targetSlot iotago.SlotIndex) error {
 	return nil
 }
 
+// Reset resets the component to a clean state as if it was created at the last commitment.
+func (m *Manager) Reset() {
+	futureAttestationsToClear := make([]iotago.SlotIndex, 0)
+	m.futureAttestations.ForEach(func(slot iotago.SlotIndex, _ *shrinkingmap.ShrinkingMap[iotago.AccountID, *iotago.Attestation]) {
```
piotrm50: Since future attestations contain only entries for slots that are not committed, we can clear the whole thing.

pkg/protocol/engine/attestation/slotattestation/manager.go
```diff
@@ -319,6 +320,31 @@ func (m *Manager) Rollback(targetSlot iotago.SlotIndex) error {
...
+		if slot > m.lastCommittedSlot {
+			futureAttestationsToClear = append(futureAttestationsToClear, slot)
+		}
+	})
+
+	pendingAttestationsToClear := make([]iotago.SlotIndex, 0)
+	m.pendingAttestations.ForEach(func(slot iotago.SlotIndex, _ *shrinkingmap.ShrinkingMap[iotago.AccountID, *iotago.Attestation]) {
```
piotrm50: Pending attestations contain only attestations that only belong to slots that were committed, so actually it might require a different reset strategy. I'm not sure about that so it's better to check with Jonas.

pkg/protocol/engine/booker/inmemorybooker/booker.go
```diff
@@ -116,6 +116,9 @@ func (b *Booker) Queue(block *blocks.Block) error {
 	return nil
 }
 
+// Reset resets the component to a clean state as if it was created at the last commitment.
+func (b *Booker) Reset() {}
```
piotrm50: Missing implementation? If nothing to reset, then add it in a comment please.

pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
```diff
@@ -244,6 +244,11 @@ func (s *Scheduler) AddBlock(block *blocks.Block) {
 	}
 }
 
+// Reset resets the component to a clean state as if it was created at the last commitment.
+func (s *Scheduler) Reset() {
+	// TODO: clear all buffers? currently irrelevant because the scheduler is not relevant for commitments.
```
piotrm50: But it's relevant for tip selection, because tips are added to the tip manager only after being scheduled. So I think we should clear all buffers first, and only then reset the TipManager. Otherwise it might happen that some old blocks will be added as tips even after resetting everything.

pkg/protocol/engine/eviction/state.go
```diff
@@ -323,6 +324,23 @@ func (s *State) PopulateFromStorage(latestCommitmentSlot iotago.SlotIndex) {
 	}
 }
 
+func (s *State) Reset() {
+	rootBlocksToClear := make([]iotago.BlockID, 0)
+	s.rootBlocks.ForEach(func(slot iotago.SlotIndex, storage *shrinkingmap.ShrinkingMap[iotago.BlockID, iotago.CommitmentID]) {
```
piotrm50: After we reset it like this, for some time we're not going to have root blocks for the earliest part of the active root block window available. I think under certain conditions it might be a problem, and cause discrepancies in block solidity. Wdyt?

pkg/protocol/engine/mempool/v1/mempool.go
piotrm50: I think we can clean the whole structure as it only contains data for uncommitted slots.

> Comment: We can not, because the StateDiffs roll something back on disk, and since we evict delayed (once you can no longer reference things, the stuff before the committed slot needs to stay.

pkg/protocol/engine/mempool/v1/mempool.go
```diff
@@ -197,6 +197,42 @@ func (m *MemPool[VoteRank]) stateDiff(slot iotago.SlotIndex) (*StateDiff, error)
...
+	attachmentsToDelete := make([]iotago.SlotIndex, 0)
+	m.attachments.ForEach(func(slot iotago.SlotIndex, _ *shrinkingmap.ShrinkingMap[iotago.BlockID, *SignedTransactionMetadata]) {
+		if slot > m.lastEvictedSlot {
+			attachmentsToDelete = append(attachmentsToDelete, slot)
+		}
+	})
+
+	for _, slot := range stateDiffsToDelete {
+		if stateDiff, deleted := m.stateDiffs.DeleteAndReturn(slot); deleted {
+			if err := stateDiff.Reset(); err != nil {
```
piotrm50: Why do we need to reset the state diff instance as well? I thought we remove the instance completely anyway.

pkg/protocol/engine/notarization/slotnotarization/slotmutations.go
piotrm50: Clear the whole thing as it only contains uncommitted data.

piotrm50 (changes requested):

pkg/protocol/engine/clock/blocktime/relativetime.go
hmoog: I will add mutexes everwhere its cleaner anyway

pkg/protocol/engine/eviction/state.go
hmoog: Let's clear one thing up since the naming is weird: the `lastEvictedSlot` is the last committed slot (we evict delayed).

Let's furthermore recap what root blocks are: Root blocks are accepted blocks of a slot that got referenced by at least one accepted block of a later slot. The reason why we track them is because we want to have a compressed perception of "all blocks that are relevant for the solidification of later slots" and they are mainly used to reduce the size of snapshots and to have a perception of the cutoff for our walks (since there is no longer a single root in the tangle once we snapshot).

We can easily prune all slots that are following our latest commitment because those are blocks that we will have to solidify again and they will become root blocks once they got referenced by accepted blocks of a later slot (through normal means).

Since we do not remove the root blocks from the committed slot, we might have dirty entries here that were referenced by blocks that were eventually orphaned but since those were still accepted blocks of our committed slot, it is perfectly legal to reference them in following slots.

In theory we would have to fill the root blocks of our target epoch to contain ALL blocks of that epoch since we don't know what chain we are syncing with and what root blocks this chain requires. That's why we provide the root blocks from the outside after the reset as part of the warp-sync process to make sure that all the required root blocks are available.

If we are resetting as part of a restart from a snapshot or disk, then we want to sync with the exact same chain and we just want to maintain the existing root block that we restored from disk (which is the previously discussed default behavior for the committed slot).

So I think this is perfectly fine.

pkg/protocol/engine/eviction/state.go
hmoog: I changed the code so everywhere where we can actually delete inline, we do it.

pkg/protocol/engine/mempool/v1/mempool.go
hmoog: I changed the code so everywhere where we can actually delete inline or clear, we do it.

pkg/protocol/engine/mempool/v1/mempool.go
hmoog: The StateDiffs roll something back on disk

pkg/protocol/engine/booker/inmemorybooker/booker.go
hmoog: I added a comment everywhere, where its empty just to comply with an upcoming interface.

pkg/protocol/engine/attestation/slotattestation/manager.go
hmoog: I talked to Jonas and we dont need to clean the pending ones at all.

pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
hmoog: I implemented the reset strategy together with Andrew

karimodm (approved):

piotrm50 (approved):

Issue #206, opened 2023-07-24
Evil Accounts spammer


karimodm:

daria305:
TODO:
 - [x] extend transaction builder in iota.go with accounts
 - [x] simple account spammer, we create new accounts from faucet funds, we do not handle those accounts
 - [ ]  evil spammer should allow to create account and remember it for further usage, and spams
 - [ ] solve bootstrapping with implicit account
 - [x]  account handling tools separated from evil spammer, like `go run . accounts` 
 - [x] solution to keep the state with basic spammer, it creates `wallet.dat` file storing the wallet state
        - both accounts tool and evil spammer use the same accountsWallet read from `wallet.dat` file so we can 
 - [ ] created account that can be used with previous functionality of evil spammer, with alias name we provide in options what account should be used for each spam:
        -      `go run . accounts --create A --amount 100000`
        -      `go run . basic --spammer tx --rate 2 --duration 60s --issuer A` <- here we will use mana from account
 - [x] make sure mana is not lost during block creation for evil spammer: `AllotRequiredManaAndStoreRemainingManaInOutput` might be useful
 - [ ] handle allotment strategies to pay for block issuance:
       - allot exact burnt mana cost, use `BuildAndSwapToBlockBuilder`
       - changing allotment amount, testing negative BIC balance
       - allot in advance for data payload spam 
 - [ ] handle other block types:  validation blocks

Evil spammer fixes:
 - [x] adapt to new outputID calculation principles
 - [ ] calculate the target number of required funds and handle funds requesting for longer spams, in `RequestFreshBigFaucetWallet` and similar functions

Fix issues regarding REST API:
 - [x] fix RMC endpoint nil reference error
 - [x] `BlockMetadata` endpoint and providing statuses does not work properly, need it for Awaiting ForConfirmation functionality of evil spammer
 - [x] Add `slotIndex` to  `BlockIssuer` endpoint to point to specific slot we need the commitment for, so that respnce from `congestion` endpoint always matches

daria305:
Example of planned script usage
```bash
./evil-spammergo \
accounts \
create --alias A --amount 1000 --noBIF true \ # noBIF - creates account that is not able to issue
create --alias B --amount 500 --implicit true \
convert --alias B \  # transition to account from implicit
allot --from A --to B --amount 100 \ # if no from then get mana from faucet output, from can be alias of existing account
update --alias B --bik hexkey --addmana 1000 --addtoken 1000 --expiryslot 15 \
[...]

PR #444, opened 2023-10-18
Fixes for REST API, RMC and visualizer

daria305:
Set of fixes regarding REST API (some of them identified in issue #206):

- Retain errors from BlockFiltered event.
  - Block can be filtered out after it's attached to the BlockDAG, the failure reason should be kept and can be checked via BlockMetadata api. 
 - RMC used for mana burn calculation should be from the same slot as commitment of the block, fixes in REST api and commitment filter
 -  add missing RMC in evilspammer block building, and set maxBurnedMana correctly
 - extend BlockIssuance endpoint with optional query param `slotIndex`, we need it to make sure both responses: congestion and blockIssuance should point to the same slot (as RMC uses slot index of the commitment) 
 - visualizer fixes

This PR is based on https://github.com/iotaledger/iota.go/pull/572


components/restapi/core/accounts.go
```diff
@@ -23,26 +22,23 @@ func congestionForAccountID(c echo.Context) (*apimodels.CongestionResponse, erro
...
-	if err != nil {
-		rmcSlot = 0
-	}
-	rmc, err := deps.Protocol.MainEngineInstance().Ledger.RMCManager().RMC(rmcSlot)
+
+	rmc, err := deps.Protocol.MainEngineInstance().Ledger.RMCManager().RMC(commitmentSlot)
 	if err != nil {
-		return nil, ierrors.Wrapf(err, "failed to get RMC for slot: %d", rmcSlot)
+		return nil, ierrors.Wrapf(err, "failed to get RMC for slot: %d", commitmentSlot)
 	}
```
jkrvivian:
```
latestCommitment := deps.Protocol.MainEngineInstance().SyncManager.LatestCommitment()
```
We can get the RMC from the commitment instead of calling RMCManager again.

jkrvivian (changes requested):

jkrvivian (approved):

karimodm (approved):
lgtm

PR #448, opened 2023-10-19
Wallet-centric account testing

### LINKED ISSUE #292: Fix accounts test changing expiry slot
The Accounts test needs to test transitioning an account to an earlier expiry slot, but this is currently not done. The genesis account output should have expiry slot math.MaxUint64, and then be transitioned to latestCommittedSlot+MaxCommittableAge, then commit until then and destroy the account.

cyberphysic4l:
The aim of these changes is to make the testing flow in the accounts tests be centred around wallets (like firefly wallets). The wallet models the following:
- hierarchical deterministic key management
- signing transactions
- signing blocks
- keeping track of owned unspent outputs

Wallets are used throughout the tests package for creation of transactions and issuance of basic blocks. 

Nodes have a blockissuer module of their own for validation, so nodes issue validation blocks.

closes #292


pkg/tests/accounts_test.go
```diff
@@ -3,40 +3,36 @@ package tests
...
 	"github.com/iotaledger/iota-core/pkg/protocol/engine/accounts"
-	"github.com/iotaledger/iota-core/pkg/protocol/snapshotcreator"
 	"github.com/iotaledger/iota-core/pkg/testsuite"
 	"github.com/iotaledger/iota-core/pkg/testsuite/mock"
+	"github.com/iotaledger/iota-core/pkg/testsuite/snapshotcreator"
 	"github.com/iotaledger/iota-core/pkg/utils"
 	iotago "github.com/iotaledger/iota.go/v4"
 )
 
 // TODO: implement tests for staking and delegation transitions that cover edge cases - part of hardening phase.
```
PhilippGackstatter: Btw, we could probably remove that line now that we have #466.

pkg/testsuite/mock/wallet.go
```diff
@@ -0,0 +1,129 @@
...
+		Testing:      t,
+		Name:         name,
+		Node:         node,
+		outputs:      make(map[string]*utxoledger.Output),
+		transactions: make(map[string]*iotago.Transaction),
+		keyManager:   keyManager,
+	}
+}
+
+func (w *Wallet) AddBlockIssuer(accountID iotago.AccountID) {
```
PhilippGackstatter:
```suggestion
func (w *Wallet) SetBlockIssuer(accountID iotago.AccountID) {
```

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -0,0 +1,386 @@
...
+			&iotago.CommitmentInput{
+				CommitmentID: w.Node.Protocol.MainEngineInstance().Storage.Settings().LatestCommitment().Commitment().MustID(),
+			},
+		}),
+		WithOutputs(iotago.Outputs[iotago.Output]{accountOutput}),
+	))
+
+	// register the outputs in the wallet
+	w.registerOutputs(transactionName, signedTransaction.Transaction)
+
```
PhilippGackstatter: Couldn't this be done in `createSignedTransactionWithOptions` so it's impossible to forget?

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -0,0 +1,386 @@
...
+				AccountID: implicitAccountID,
+			},
+			&iotago.CommitmentInput{
+				CommitmentID: w.Node.Protocol.MainEngineInstance().Storage.Settings().LatestCommitment().Commitment().MustID(),
+			},
+		}),
+		WithInputs(utxoledger.Outputs{input}),
+		WithOutputs(iotago.Outputs[iotago.Output]{accountOutput}),
+		WithSlotCreated(creationSlot),
+	))
```
PhilippGackstatter:
```suggestion
	signedTransaction := lo.PanicOnErr(w.createSignedTransactionWithOptions(
		WithContextInputs(iotago.TxEssenceContextInputs{
			&iotago.BlockIssuanceCreditInput{
				AccountID: implicitAccountID,
			},
			&iotago.CommitmentInput{
				CommitmentID: w.Node.Protocol.MainEngineInstance().Storage.Settings().LatestCommitment().Commitment().MustID(),
			},
		}),
		WithInputs(utxoledger.Outputs{input}),
		WithOutputs(iotago.Outputs[iotago.Output]{accountOutput}),
		WithSlotCreated(creationSlot),
		func(txBuilder *builder.TransactionBuilder) {
			txBuilder.AllotAllMana(creationSlot, implicitAccountID)
		},
	))
```

If you rewrite it like that, you can get rid of `createSignedTransactionWithAllotmentAndOptions` so there's less duplication.

pkg/tests/accounts_test.go
```diff
@@ -3,40 +3,36 @@ package tests
...
-func Test_TransitionAccount(t *testing.T) {
+func Test_TransitionAndDestroyAccount(t *testing.T) {
 	oldGenesisOutputKey := utils.RandBlockIssuerKey()
+
+	// add a genesis account to the snapshot which expires in slot 1.
```
PhilippGackstatter: No longer true now that you've set it to `MaxSlotIndex`.

pkg/tests/accounts_test.go
```diff
@@ -51,68 +47,73 @@ func Test_TransitionAccount(t *testing.T) {
...
-		}),
-		testsuite.WithOutputs(append(accountOutputs, equalOutputs...)),
-		testsuite.WithAllotments(iotago.Allotments{&iotago.Allotment{
-			AccountID: genesisAccountOutput.AccountID,
-			Value:     0,
-		}}),
-	))
 
-	var block1Slot iotago.SlotIndex = 1
+	// defaults block issuer issues a block containing the transaction in slot 1.
```
PhilippGackstatter:
```suggestion
	// default block issuer issues a block containing the transaction in slot 1.
```

pkg/tests/accounts_test.go
```diff
@@ -51,68 +47,73 @@ func Test_TransitionAccount(t *testing.T) {
...
 	genesisCommitment := iotago.NewEmptyCommitment(ts.API.ProtocolParameters().Version())
 	genesisCommitment.ReferenceManaCost = ts.API.ProtocolParameters().CongestionControlParameters().MinReferenceManaCost
-
-	block1 := ts.IssueBasicBlockAtSlotWithOptions("block1", block1Slot, genesisCommitment, blockIssuer, node1, tx1)
-
+	block1 := ts.IssueBasicBlockAtSlotWithOptions("block1", block1Slot, ts.DefaultWallet(), tx1, mock.WithSlotCommitment(genesisCommitment))
 	latestParent := ts.CommitUntilSlot(ts.BlockID("block1").Slot(), block1)
 
+	// assert diff of a genesis account, it should have a new output ID, same expiry slot and a new block issuer key.
```
PhilippGackstatter:
```suggestion
	// assert diff of the genesis account, it should have a new output ID, same expiry slot and a new block issuer key.
```

There's only one "genesis account" iiuc.

PhilippGackstatter (approved):
Looks good to me!

Nice that you factored the keymanager out of the wallet and made the transactions separate from the accounts test. That is much nicer to read now!

pkg/testsuite/mock/wallet_transactions.go
cyberphysic4l: Nice, that is much cleaner

pkg/testsuite/mock/wallet_transactions.go
cyberphysic4l: For the recipient wallet, we have to do it here because we don't pass the recipient wallet itself to `createSignedTransactionWithOptions`. It's replacing the step that the firefly wallet would have to do to check for new outputs received on its addresses, but maybe we could automate this a little more in the wallet, I will see if I can think of something.

cyberphysic4l: I moved the registerOutputs for the sending wallet into the `createSignedTransactionWithOptions` at least, but we still need to register them for the recipient wallet explicitly for now.

PR #453, opened 2023-10-20
Added ListenToAcceptedTransactions to INX

alexsporn:


components/inx/server_utxo.go
```diff
@@ -244,34 +246,34 @@ func (s *Server) ListenToLedgerUpdates(req *inx.SlotRangeRequest, srv inx.INX_Li
...
 		latestCommitment := deps.Protocol.MainEngineInstance().SyncManager.LatestCommitment()
 
-		if startIndex > latestCommitment.Slot() {
+		if startSlot > latestCommitment.Slot() {
 			// no need to send previous milestone diffs
 			return 0, nil
 		}
 
 		// Stream all available milestone diffs first
 		pruningIndex := deps.Protocol.MainEngineInstance().SyncManager.LatestFinalizedSlot()
```
muXxer: two things, why is it called `pruningIndex` instead of slot and why is it the latest finalized slot? :thinking:

muXxer (changes requested):

muXxer (approved):

PR #463, opened 2023-10-25
Set the error codes in api responses

### LINKED ISSUE #429: Verify error handling in restapi/core pkg
Some of the handlers in the `restapi/core` pkg are returning errors that are not wrapping `httpserver.Err*` or `echo.Err*` errors. 
If we do not use the correct underlying errors the API will always return a `500 Internal Server Error` instead of the correct status code.

e.g.
```
{"error":{"code":"500","message":"internal server error. error: block not found: 0x2b9ab0b8c2ea3937fdc31c79b80be7c1f3a70e5bf4b05b5f75a38b128358023d09000000"}}
```

`ierrors.Errorf("block not found: %s", blockID.ToHex())`
should be changed to 
`ierrors.Wrapf(echo.ErrNotFound, "block not found: %s", blockID.ToHex())`

jkrvivian:
Close #429

### DIFF
--- a/components/restapi/core/accounts.go
+++ b/components/restapi/core/accounts.go
@@ -19,17 +19,17 @@ import (
 func congestionForAccountID(c echo.Context) (*apimodels.CongestionResponse, error) {
 	accountID, err := httpserver.ParseAccountIDParam(c, restapipkg.ParameterAccountID)
 	if err != nil {
-		return nil, err
+		return nil, ierrors.Wrapf(echo.ErrBadRequest, "failed to parse account ID %s: %s", c.Param(restapipkg.ParameterAccountID), err)
 	}
 
 	commitment := deps.Protocol.MainEngineInstance().SyncManager.LatestCommitment()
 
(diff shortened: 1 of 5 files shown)


cyberphysic4l (approved):

components/restapi/core/accounts.go
```diff
@@ -19,17 +19,17 @@ import (
 func congestionForAccountID(c echo.Context) (*apimodels.CongestionResponse, error) {
 	accountID, err := httpserver.ParseAccountIDParam(c, restapipkg.ParameterAccountID)
 	if err != nil {
-		return nil, err
+		return nil, ierrors.Wrapf(echo.ErrBadRequest, "failed to parse account ID %s: %s", c.Param(restapipkg.ParameterAccountID), err)
```
alexsporn: no need to wrap errors that come from the `httpserver` package since those are already properly wrapped with underlying http errors.
These changes can be reverted in the whole PR.

PR #477, opened 2023-10-31
Fix context based on static code analysis tools run

### LINKED ISSUE #248: Check captured variable in closures

karimodm:
Closes #248 

Two tools are provided: `detect_captured_variables` that detects if any variable is captured by a closure without an explicit copy (2000+ cases are reported in iota-core only), and `detect_modified_after_captured_variables` that only reports if such captured variable is modified somewhere else in the same file. I've reviewed the report of the second tool, being the results of the first too verbose.

_We should go through all the cases reported by the `detect_captured_variables` tool, one by one._ I haven't done that for lack of time and the cost/potential benefit of such task.

The `detect_modified_after_captured_variables` code analysis tool has some clear limitations:
- It is not scope-aware: it will detect a variable being modified if another with the same name is declared in another scope, even in another function.
- It is just a static-analysis tool: if the variable the closure is capturing is modified in another file, through a method that receives it as a pointer, or through any other indirect reference (e.g. pointer) the tool won't detect the modification.

The following finds have been identified, but they are not immediately dangerous:
```
./pkg/protocol/engine/ledger/ledger/ledger.go:567: variable 'accountDiffs' captured by closure and modified at ./pkg/protocol/engine/ledger/ledger/ledger.go:596
./pkg/protocol/protocol_fork.go:108: variable 'detachProcessCommitment' captured by closure and modified at ./pkg/protocol/protocol_fork.go:122
./pkg/protocol/protocol_fork.go:109: variable 'detachMainEngineSwitched' captured by closure and modified at ./pkg/protocol/protocol_fork.go:143
```


karimodm:
Move to a new repo https://github.com/iotaledger/iota-core-tools , so removed the tooling from here.

muXxer (approved):

PR #479, opened 2023-10-31
Add support for GenesisSlot protocol parameters to have a genesis != 0

### LINKED ISSUE #431: Add a GenesisSlot parameter to the protocol
We should be able to bootstrap a network with a genesis slot different than 0.
This allows us to e.g. migrate Shimmer and IOTA and keep the old milestone indexes to have a coherent tangle history.
```[tasklist]
### Tasks
- [x] Add GenesisSlot parameter to the ProtocolParameters
- [x] Use the GenesisSlot in the TimeProvider
- [x] Adapt all special cases of 0 as genesis to use GenesisSlot instead
```

alexsporn:
Depends on https://github.com/iotaledger/iota.go/pull/592


components/validator/issuer.go
```diff
@@ -109,7 +109,7 @@ func reviveChain(issuingTime time.Time) (*iotago.Commitment, iotago.BlockID, err
 	apiForSlot := deps.Protocol.APIForSlot(lastCommittedSlot)
 
 	// Get a rootblock as recent as possible for the parent.
-	parentBlockID := iotago.EmptyBlockID
+	parentBlockID := iotago.EmptyBlockID //TODO: use GenesisBlockID
```
karimodm: Can it be resolved?

pkg/protocol/engine/engine.go
```diff
@@ -183,7 +183,7 @@ func New(
 				}
 
 				// Only mark any pruning indexes if we loaded a non-genesis snapshot
-				if e.Storage.Settings().LatestFinalizedSlot() > 0 {
+				if e.Storage.Settings().LatestFinalizedSlot() > e.Storage.GenesisRootBlockID().Slot() {
```
karimodm, line 186: Don't we have protoParams.GenesisSlot available here?

pkg/protocol/engine/ledger/ledger/ledger.go
```diff
@@ -180,11 +180,14 @@ func (l *Ledger) CommitSlot(slot iotago.SlotIndex) (stateRoot iotago.Identifier,
 
 	// Update the Accounts ledger
 	// first, get the RMC corresponding to this slot
-	maxCommittableAge := l.apiProvider.APIForSlot(slot).ProtocolParameters().MaxCommittableAge()
-	rmcIndex, _ := safemath.SafeSub(slot, maxCommittableAge)
-	rmcForSlot, err := l.rmcManager.RMC(rmcIndex)
+	protocolParams := l.apiProvider.APIForSlot(slot).ProtocolParameters()
+	rmcSlot, _ := safemath.SafeSub(slot, protocolParams.MaxCommittableAge())
```
karimodm: Maybe worth mentioning that you are ignoring the underflow because SafeSub returns 0 in such case.

pkg/testsuite/snapshotcreator/snapshotcreator.go
```diff
@@ -120,6 +121,10 @@ func CreateSnapshot(opts ...options.Option[Options]) error {
 	)
 	defer engineInstance.Shutdown()
 
+	if opt.AddGenesisRootBlock {
+		engineInstance.EvictionState.AddRootBlock(api.ProtocolParameters().GenesisBlockID(), genesisCommitment.ID())
```
karimodm, line 125: If this option is not provided the genesis root block in the eviction state will default to `NewBlockID(p.basicProtocolParameters.GenesisSlot, EmptyIdentifier)` which is the same you are calling here.
```
// GenesisBlockID defines the block ID of the genesis block.
func (p *V3ProtocolParameters) GenesisBlockID() BlockID {
	return NewBlockID(p.basicProtocolParameters.GenesisSlot, EmptyIdentifier)
}
```

Am I missing something?

karimodm (changes requested):

components/validator/issuer.go
piotrm50: I think so, I've created an accompanying PR in inx-validator. Conflicts on this file should be resolved by taking version from develop.

piotrm50 (approved):
Before merging this one, merge inx-validator PR first and make a release of it, so that Ansible deployment and Docker network continue to work correctly.

pkg/testsuite/snapshotcreator/snapshotcreator.go
alexsporn: Correct. IMHO the default in the eviction state should be changed. We should not default to a genesis root block at any point in time.

pkg/protocol/engine/ledger/ledger/ledger.go
alexsporn: Added it. I stumbled myself here when refactoring the code.

pkg/protocol/engine/engine.go
alexsporn: Yes, I used this one because it was shorter than `e.Storage.Settings().APIProvider().CommittedAPI().ProtocolParameters().GenesisSlot()`, but I can change it if you like that better.

karimodm (approved):

PR #287, opened 2023-08-15
Address Accounts Ledger TODOs 

karimodm:
Resolves #261 

This PR won't resolve "Add in memory shrink version of slot diffs": the reason for this is that the slotDiff is a complex structure that stores diffs for the accounts and the information about what accounts have been destroyed. Adding a caching layer in form a shrinking map will render the code quite more complex, without any obvious benefits, as accounts can be resolved from storage using key lookups; hence most likely cached by RocksDB quite efficiently for often-used accounts.
We will optimize this if and when this ends up being an actual bottleneck.

PR #483, opened 2023-11-02
Use safemath in accounts

### LINKED ISSUE #223: Safemath everywhere
- use addition whenever possible
- when it makes sense compute with safemath
- don't panic, always return ierrors across the code
```[tasklist]
### Tasks
- [x] `pkg/core/account/accounts.go`
- [x] `pkg/protocol/engine/accounts/accountsledger/manager.go`
- [x] `pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go`
```

karimodm:
Closes #223 

- `pkg/core/account/accounts.go` -> addressed by this PR
- `pkg/protocol/engine/accounts/accountsledger/manager.go` -> addressed by #287 
- `pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go` -> addressed by #283

### DIFF
--- a/pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
+++ b/pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
@@ -161,7 +161,11 @@ func (o *SybilProtection) CommitSlot(slot iotago.SlotIndex) (committeeRoot, rewa
 				panic(fmt.Sprintf("committee for current epoch %d not found", currentEpoch))
 			}
 
-			committeeAccounts := committee.Accounts()
+			committeeAccounts, err := committee.Accounts()
+			if err != nil {
+				return iotago.Identifier{}, iotago.Identifier{}, ierrors.Wrapf(err, "failed to get accounts from committee for epoch %d", currentEpoch)
+			}
+
 			committeeAccounts.SetReused()
 			if err = o.seatManager.SetCommittee(nextEpoch, committeeAccounts); err != nil {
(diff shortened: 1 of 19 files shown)


pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
```diff
@@ -276,7 +280,13 @@ func (o *SybilProtection) slotFinalized(slot iotago.SlotIndex) {
 	epochEndSlot := timeProvider.EpochEnd(epoch)
 	if slot+apiForSlot.ProtocolParameters().EpochNearingThreshold() == epochEndSlot &&
 		epochEndSlot > o.lastCommittedSlot+apiForSlot.ProtocolParameters().MaxCommittableAge() {
-		newCommittee := o.selectNewCommittee(slot)
+		newCommittee, err := o.selectNewCommittee(slot)
+		if err != nil {
+			// TODO: should we fail "harder" here?
```
piotrm50: The method inside panicks on error. Should return those errors instead and panic here? 
I suppose it can't be exploited to cause panic for everyone in the network as someone would need A LOT of stake, which should be protected by mana generation calculations etc. But what else can we do in this case?

piotrm50 (approved):

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
karimodm: Yep, decided to return all the errors and panic here :+1:

PR #484, opened 2023-11-02
Validator tip selection

### LINKED ISSUE #406: Validator Tip Selection
We need to define a tip selection algorithm for validators which selects `CommitteeSize` validation blocks as tips, one for each validator in the committee from the last subslot, and `BlockMaxParents`==8 basic blocks as tips. 

The current problem with validation blocks tips is that the maxLen contraint in the serix tag for the parents field of the blocks can not be dynamic at the moment. If we have a max committee size, we can simply set this maxLen to ensure even a full committee would not violate that constraint. 

The incentives and finalisation taskforces have a number of proposals for the validator tip selection, so further discussion is required with them to decide on one which is sufficiently straightforward to implement while still achieving some of our important objectives.

cyberphysic4l:
For validator tip selection, we will use a larger number of maximum parents. 
This PR:
- renames those parameters as they were confusing names before.
- adds this simple validator tip selection to the block issuer in the test suite.

closes https://github.com/iotaledger/iota-core/issues/406

### DIFF
--- a/components/restapi/core/blocks.go
+++ b/components/restapi/core/blocks.go
@@ -45,7 +45,7 @@ func blockMetadataByID(c echo.Context) (*apimodels.BlockMetadataResponse, error)
 }
 
 func blockIssuanceBySlot(slotIndex iotago.SlotIndex) (*apimodels.IssuanceBlockHeaderResponse, error) {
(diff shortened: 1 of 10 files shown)


cyberphysic4l:
closes #406

piotrm50 (approved):

PR #485, opened 2023-11-02
Fix node shutdown and cleanup Docker

piotrm50:


tools/docker-network/run.sh
```diff
@@ -1,17 +1,16 @@
 #!/bin/bash
-set -e
+
 
 # Create a function to join an array of strings by a given character
 function join { local IFS="$1"; shift; echo "$*"; }
 
 # All parameters can be optional now, just make sure we don't have too many
 if [[ $# -gt 4 ]] ; then
```
karimodm: This needs to change too, as we now have only 1 possible parameter.

karimodm (changes requested):

karimodm (approved):

Issue #451, opened 2023-10-20
Nil pointer in TipSel


karimodm:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0xe9f673]

goroutine 168 [running]:
github.com/iotaledger/iota-core/pkg/model.(*Block).ProtocolBlock(...)
        /scratch/pkg/model/block.go:89
github.com/iotaledger/iota-core/pkg/protocol/engine/tipselection/v1.DynamicLivenessThreshold.func1({0x2c0dea0, 0xc01397dc80})
        /scratch/pkg/protocol/engine/tipselection/v1/provider.go:45 +0x33
github.com/iotaledger/iota-core/pkg/protocol/engine/tipselection/v1.(*TipSelection).classifyTip(0xc0003dcb00, {0x2c0dea0, 0xc01397dc80})
        /scratch/pkg/protocol/engine/tipselection/v1/tip_selection.go:166 +0xf9
github.com/iotaledger/hive.go/runtime/event.(*Event1[...]).Trigger.func1(0xc0005cf4a0?)
        /go/pkg/mod/github.com/iotaledger/hive.go/runtime@v0.0.0-20231019175611-f85bce9102d1/event/events.go:88 +0x1b4
github.com/iotaledger/hive.go/ds/orderedmap.(*OrderedMap[...]).ForEach(0xc0105e7b48, 0xc0105e7b20?)
        /go/pkg/mod/github.com/iotaledger/hive.go/ds@v0.0.0-20231019175611-f85bce9102d1/orderedmap/orderedmap.go:130 +0xc4
github.com/iotaledger/hive.go/runtime/event.(*Event1[...]).Trigger(0xc039f6a750?, {0x2c0dea0?, 0xc01397dc80?})
        /go/pkg/mod/github.com/iotaledger/hive.go/runtime@v0.0.0-20231019175611-f85bce9102d1/event/events.go:70 +0x8a
github.com/iotaledger/iota-core/pkg/protocol/engine/tipmanager/v1.(*TipManager).setupBlockMetadata(0xc000394e00, 0xc01397dc80)
        /scratch/pkg/protocol/engine/tipmanager/v1/tipmanager.go:143 +0x146
[...]

PR #489, opened 2023-11-03
Add log to debug tipselection nilpointer.

piotrm50:
Code print in a situation in which nil pointer exception is about to happen in TipSelection and prints the problematic Block.
This is to help debugging #451.

### DIFF
--- a/pkg/protocol/engine/tipmanager/v1/tipmanager.go
+++ b/pkg/protocol/engine/tipmanager/v1/tipmanager.go
@@ -152,7 +152,7 @@ func (t *TipManager) forEachParentByType(block *blocks.Block, consumer func(pare
 			}
 
 			if parentBlock.ModelBlock() == nil {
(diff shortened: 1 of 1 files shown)


jonastheis (commented):
Would be good to also print the information of blocks.Block. Maybe there's something in there that further helps pinpoint the problem

piotrm50:
> Would be good to also print the information of blocks.Block. Maybe there's something in there that further helps pinpoint the problem

Do you mean not the parent, but the block from the parameter? Currently `.String()` is called on `*blocks.Block()`.

jonastheis:
> > Would be good to also print the information of blocks.Block. Maybe there's something in there that further helps pinpoint the problem
> 
> Do you mean not the parent, but the block from the parameter? Currently `.String()` is called on `*blocks.Block()`.

Exactly. Sorry for being so inconcise

jonastheis (approved):

PR #499, opened 2023-11-07
Implement test that locks and unlocks an account due to negative BIC.

piotrm50:


pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -321,7 +321,7 @@ func (w *Wallet) CreateBasicOutputsEquallyFromInputs(transactionName string, out
 	return signedTransaction
 }
 
-func (w *Wallet) createSignedTransactionWithOptions(transactionName string, opts ...options.Option[builder.TransactionBuilder]) (*iotago.SignedTransaction, error) {
+func (w *Wallet) CreateSignedTransactionWithOptions(transactionName string, opts ...options.Option[builder.TransactionBuilder]) (*iotago.SignedTransaction, error) {
```
cyberphysic4l: I would prefer to keep this as a private method and use utility functions for specific types of transactions. This just helps keep the tests themselves a bit cleaner and clearer.

pkg/tests/accounts_test.go
```diff
@@ -452,6 +457,219 @@ func Test_ImplicitAccounts(t *testing.T) {
...
+	)
+	defer ts.Shutdown()
+
+	// Add a validator node to the network. This will add a validator account to the snapshot.
+	node1 := ts.AddValidatorNode("node1")
+	node2 := ts.AddNode("node2")
+
+	validatorBIC := iotago.MaxBlockIssuanceCredits / 2
+	// Add a default block issuer to the network. This will add another block issuer account to the snapshot.
+	walletBIC := iotago.BlockIssuanceCredits(499)
```
cyberphysic4l: I think it's better to just start with zero BIC so that we are certain they will go negative regardless of RMC and the WorkScore params.

pkg/tests/accounts_test.go
```diff
@@ -452,6 +457,219 @@ func Test_ImplicitAccounts(t *testing.T) {
...
+	wallet := ts.AddGenesisWallet("default", node2, walletBIC)
```
cyberphysic4l: I think it would be nicer to create two new Wallets in the snapshot, called wallet1 and wallet2 or something, rather than using the default wallet and the wallet attached to node 1.

cyberphysic4l:
```suggestion
	wallet1 := ts.AddGenesisWallet("wallet 1", node2, walletBIC)
    wallet2 := ts.AddGenesisWallet("wallet 2", node2, walletBIC)
```

pkg/tests/accounts_test.go
```diff
@@ -452,6 +457,219 @@ func Test_ImplicitAccounts(t *testing.T) {
...
+		block1Commitment.ReferenceManaCost = ts.API.ProtocolParameters().CongestionControlParameters().MinReferenceManaCost
+		block11 := ts.IssueBasicBlockAtSlotWithOptions("block1.1", block1Slot, ts.Wallet("node1"), &iotago.TaggedData{}, mock.WithSlotCommitment(block1Commitment))
+		block12 := ts.IssueBasicBlockAtSlotWithOptions("block1.2", block1Slot, wallet, &iotago.TaggedData{}, mock.WithStrongParents(block11.ID()), mock.WithSlotCommitment(block1Commitment))
+
+		// Commit BIC burns and check account states.
+		latestParent = ts.CommitUntilSlot(ts.BlockID("block1.2").Slot(), block12)
+
+		burned := iotago.BlockIssuanceCredits(block11.WorkScore()) * iotago.BlockIssuanceCredits(ts.API.ProtocolParameters().CongestionControlParameters().MinReferenceManaCost)
+
+		validatorBIC -= burned
```
cyberphysic4l: This would become wallet1BIC and wallet2BIC rather having validatorBIC. There is no reason for this other one to be a validator so its more clear this way.

pkg/tests/accounts_test.go
```diff
@@ -452,6 +457,219 @@ func Test_ImplicitAccounts(t *testing.T) {
...
+			Credits:         accounts.NewBlockIssuanceCredits(walletBIC, block1Slot),
+			ExpirySlot:      iotago.MaxSlotIndex,
+			OutputID:        blockIssuerAccountOutput.OutputID(),
+			BlockIssuerKeys: wallet.BlockIssuer.BlockIssuerKeys(),
+		}, ts.Nodes()...)
+	}
+
+	block3Slot := latestParent.ID().Slot()
+
+	// Allot some mana to the locked account to unlock it.
```
cyberphysic4l, line 587: Specify which account is going to issue the block that unlocks the negative account.

pkg/tests/accounts_test.go
```diff
@@ -452,6 +457,219 @@ func Test_ImplicitAccounts(t *testing.T) {
...
+	{
+		allottedBIC := iotago.BlockIssuanceCredits(500)
+		tx1, err := wallet.CreateSignedTransactionWithOptions(
```
cyberphysic4l: create a wallet method called AllotManaToWallet() rather than using the createSignedTransactionWithOptions() method.

pkg/tests/accounts_test.go
cyberphysic4l: It is the wallet of the validator node1 that does it in this case

pkg/tests/accounts_test.go
cyberphysic4l: Add a comment here to to highlight that the locked wallet is creating the transaction, but not issuing the block.

pkg/tests/accounts_test.go
```diff
@@ -452,6 +457,219 @@ func Test_ImplicitAccounts(t *testing.T) {
...
+			mock.WithSlotCreated(block3Slot),
+			mock.WithAllotments(iotago.Allotments{&iotago.Allotment{
+				AccountID: wallet.BlockIssuer.AccountID,
+				Mana:      iotago.Mana(allottedBIC),
+			}}),
+		)
+		require.NoError(t, err)
+
+		block3Commitment := node1.Protocol.MainEngineInstance().Storage.Settings().LatestCommitment().Commitment()
+		block31 := ts.IssueBasicBlockAtSlotWithOptions("block3.1", block3Slot, ts.Wallet("node1"), tx1, mock.WithStrongParents(latestParent.ID()), mock.WithSlotCommitment(block3Commitment))
```
cyberphysic4l: Add comment to highlight which node is issuing the block.

cyberphysic4l (approved):

pkg/tests/accounts_test.go
```diff
@@ -851,3 +864,233 @@ func Test_NegativeBIC_LockedAccountOutput(t *testing.T) {
...
+			OutputID:        wallet2AccountOutput.OutputID(),
+			ExpirySlot:      iotago.MaxSlotIndex,
+			BlockIssuerKeys: wallet2.BlockIssuer.BlockIssuerKeys(),
+		}, ts.Nodes()...)
+	}
+
+	block3Slot := latestParent.ID().Slot()
+
+	// UNLOCK THE ACCOUNT
+	// The locked wallet 2 is preparing and signs the transaction, but it's issued by wallet 1 whose account is not locked.
```
cyberphysic4l: The comment is reversed, should be wallet 1 creates tx and wallet 2 issues. Same with below comment.

pkg/tests/accounts_test.go
```diff
@@ -1017,7 +1017,8 @@ func Test_NegativeBIC_AccountOwnedBasicOutputLocked(t *testing.T) {
 	block3Slot := latestParent.ID().Slot()
 
 	// UNLOCK THE ACCOUNT
-	// The locked wallet 2 is preparing and signs the transaction, but it's issued by wallet 1 whose account is not locked.
+	// The locked wallet 2 is preparing and signs the transaction,
```
cyberphysic4l, line 1020: This one is still reversed, it's wallet 1 that is locked

cyberphysic4l (approved):

PR #502, opened 2023-11-07
Send SlotCommitmentID for each BlockID in the WarpSyncResponse

alexsporn:
We add all received blockIDs as root blocks but since we were missing the corresponting `SlotCommitmentID` we were wrongly storing the root blocks with an empty commitment ID. This PR now changes warpsync to send all blockIDs grouped by the corresponding SlotCommitmentID, so that we can use that when adding the blocks to the root blocks.

This PR simplifies the protobuf `WarpSyncRequest` by introducing a serix-serialized `WarpSyncPayload` that hold all the fields instead of serializing/deserializing every single field.


jonastheis (approved):

karimodm (approved):
Looks proper!

PR #506, opened 2023-11-08
Enable dPoS

### LINKED ISSUE #494: Finalization threshold on _seated_ committee members
Thresholds should not be computed on maximum committee size but on seated committee members: this is in case we could not select enough committee members.

### LINKED ISSUE #495: Enable top-stakers dPoS

jonastheis:
This PR enables the previously implemented top stakers committee selection. It adapts to the changes from  https://github.com/iotaledger/iota.go/pull/598 and also allows for committees that are smaller than `TargetCommitteeSize` in case there's fewer candidates available.

It adds end-to-end test for following scenarios:
- committee rotation with candidacy announcements at different times (slots) within the epoch
- committee reuse due to no candidates announced but slot finalized
- committee reuse due to no slot finalization at `EpochNearingThreshold`
- committee rotation to a smaller committee than `TargetCommitteeSize`

Fixes #494 and #495


piotrm50 (approved):
So far looks good :) Just need to remove some printlns and maybe run it in a docker network for a while with shorter epochs.

piotrm50 (approved):
lgtm

PR #518, opened 2023-11-13
Accounts tests upgrades

cyberphysic4l:


pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -426,37 +469,82 @@ func (w *Wallet) AllotManaFromInputs(transactionName string, allotments iotago.A
...
+	inputDelegation, ok := input.Output().(*iotago.DelegationOutput)
+	if !ok {
+		panic(fmt.Sprintf("output with alias %s is not *iotago.AccountOutput", inputName))
+	}
+
+	rewardMana, _, _, err := w.Node.Protocol.MainEngineInstance().SybilProtection.DelegatorReward(
+		inputDelegation.ValidatorAddress.AccountID(),
+		inputDelegation.DelegatedAmount,
+		inputDelegation.StartEpoch,
+		inputDelegation.EndEpoch,
```
PhilippGackstatter, line 495: `EndEpoch` might be 0 when the Delegation Output is in its regular state, so this will not always have the intended behavior. We can fix that in another PR, though.

PhilippGackstatter (approved):

PR #526, opened 2023-11-15
Fix genesisSlot shenanigans in snapshot export & import

karimodm:


alexsporn (approved):

pkg/protocol/engine/attestation/slotattestation/snapshot.go
```diff
@@ -51,7 +51,7 @@ func (m *Manager) Export(writer io.WriteSeeker, targetSlot iotago.SlotIndex) err
 	}
 
 	if _, isValid := m.computeAttestationCommitmentOffset(targetSlot); !isValid {
-		if err := stream.Write(writer, uint64(0)); err != nil {
```
jonastheis, line 54: You should use the type and the `stream` package here and then you get type safety for free

PR #529, opened 2023-11-16
Fixed INX failing to send the UTXO for accepted tx

### LINKED ISSUE #528: mempool.TransactionMetadata.Inputs().ForEach broken because of genesis slot
I guess the change with the genesis slot broke the network. when the evil spammer starts to spam (I don't know what exactly is spammed), the inx server in the valdiator node throws the following error:
```
docker-network-node-1-validator-1  | 2023-11-16T14:36:56Z       ERROR   INX     app@v0.0.0-20231110191152-7135670285dc/component.go:115 error creating payload: commitment is before genesis: slot 0 is before genesis slot 5: failed to load commitment with slot: 0
```

The only code part where this error is thrown is here:
```
func (c *Commitments) Load(slot iotago.SlotIndex) (commitment *model.Commitment, err error) {
	genesisSlot := c.apiProvider.CommittedAPI().ProtocolParameters().GenesisSlot()
	if slot < genesisSlot {
		return nil, ierrors.Wrapf(ErrCommitmentBeforeGenesis, "slot %d is before genesis slot %d", slot, genesisSlot)
	}

	return c.store.Get(slot)
}
```

but I don't even understand how this is actually called from here where the error is thrown in the end:
```
		var consumed []*inx.LedgerSpent
		if err := transactionMetadata.Inputs().ForEach(func(stateMetadata mempool.StateMetadata) error {
			spentOutput, ok := stateMetadata.State().(*utxoledger.Output)
			if !ok {
				return ierrors.Errorf("unexpected state metadata type: %T", stateMetadata.State())
			}

			inxSpent, err := NewLedgerSpent(utxoledger.NewSpent(spentOutput, transactionMetadata.ID(), slot))
			if err != nil {
				return err
			}
			consumed = append(consumed, inxSpent)

			return nil
		}); err != nil {
			Component.LogErrorf("error creating payload: %v", err)
			cancel()
		}
```
I mean the ForEach does that magically, but my IDE can't follow because of all the funny interfaces and generics and whatnot.
Anyway, should it even be possible in the first place that something got included in the Inputs that breaks the Call to ForEach?
This breaks the network because the inx-indexer doesn't get updates via INX this way.

alexsporn:


muXxer:
Fixes #529

muXxer (approved):

PR #530, opened 2023-11-19
includes a test comparing rewards to float results

### LINKED ISSUE #368: New tests for the rewards calculations
the tests for the rewards calculation are not that extensive, we should write new ones. Also, it's a good idea to compare results with the ones resulting from floating point operations

oliviasaa:


oliviasaa:
I have no idea what's going on with this unit test that is failing

oliviasaa:
also, I had to use `api.ComputedFinalReward()` and `api.ComputedInitialReward()`, otherwise the test would fail. I'm having trouble figuring out if those values are being fed to the code manually (thus, they are just arbitrary values) or if they are calculated incorrectly and that's why the tests fail when I don't use them.

piotrm50:
The failing tests seem to be unrelated to your changes. It seems to be a data race in TipSelection component that we need to investigate separetely.

```
2023-11-19T17:42:36.9547917Z --- FAIL: TestLossOfAcceptanceFromGenesis (2.56s)
2023-11-19T17:42:36.9548122Z     testing.go:1465: race detected during execution of test
```

pkg/protocol/sybilprotection/sybilprotectionv1/performance/testsuite_test.go
```diff
@@ -260,14 +264,38 @@ func (t *TestSuite) calculatePoolReward(epoch iotago.EpochIndex, totalValidators
...
+	performanceFactorFloat := float64(performanceFactor) / float64(params.ValidationBlocksPerSlot())
+	// theta := float64(params.RewardsParameters().ManaShareCoefficient)
+	// S := float64(params.TokenSupply())
+	// gamma := float64(params.ManaParameters().GenerationRate) * math.Pow(2.0, -float64(params.ManaParameters().GenerationRateExponent))
+	d := math.Pow(2.0, float64(params.SlotsPerEpochExponent()))
+	betaDelta := d * float64(params.SlotDurationInSeconds()) / 3.0 / (365 * 24 * 60 * 60)
+
+	targetRewardFloat := float64(t.api.ComputedFinalReward())
+
+	// targetRewardFloat := theta * S * gamma * d
```
cyberphysic4l: We should get this part to work. As we discussed, this probably has something to do with different protocol parameters being used in different places, but otherwise there is a mistake somewhere in either calculation, so I think we need to figure this out rather than just leave it out.

cyberphysic4l (changes requested):

cyberphysic4l (approved):

Issue #266, opened 2023-08-04
Revisit the weight we're using in slotattestation.Commit


jkrvivian:
Double check which weight we're using, the current one, or the one of the slot of the attestation/commitmentID? 

file: `pkg/protocol/engine/attestation/slotattestation/manager.go`

jonastheis:
Addressed with #531

Issue #255, opened 2023-08-03
Resolve TODOs in Commitment Verifier


jkrvivian:
* [x] What happens if the committee rotated after the fork?
* [x]  `verifyAttestations`: this might differ if we have Accounts with changing weights depending on the SlotIndex/epoch

jonastheis:
This can be done together with #210

jonastheis:
Addressed with #531

PR #531, opened 2023-11-20
Engine switching with committee rotation

### LINKED ISSUE #210: Test engine switching over epoch boundaries with rotating committee
One possible test case: a minority partition does not finalize in time and keeps the current committee. Eventually the partition is merged, after the committee changed. The nodes of that partition should switch correctly and adapt the new committee. Also noteworthy: the attestations at epoch boundary and the weights being correct for the new committee.

jonastheis:
This PR addresses several problems around engine switching with committee rotation across epoch boundaries:
1. `AttestationManager`: Use the committee of the slot of the attestation in contrast to the slot that we're committing. This is because at the time the attestation was created, the committee might have been different from the current one (due to rotation at epoch boundary).
2. `CommitmentVerifier`: Update `validatorAccountsData` if fork happened across epoch boundaries. We try to use the latest accounts data (at `lastCommonSlotBeforeFork`) for the current epoch. This is necessary because the committee might have rotated at the epoch boundary and different validators might be part of it. In case anything goes wrong we keep using previously known accounts data (initially set to the accounts data  of the validators for the epoch of the last common commitment before the fork). It could be, that we don't know about a validator when we verify the attestations: we then simply ignore this attestation (just like with public key changes) and weight of the chain will build up slower.
3. Fixes some problems with race conditions with force accepting and confirmation/finalization that lead to wrong commitments for a node syncing up with warp sync as weight was not fully propagated when the slot was force committed. 
4. Implements test for scenario where a node is separated into a partition and due to no finalization reuses the committee while the majority partition goes on to select a new committee due to finalization. After the partition is merged the node of the minority partition successfully joins the majority partition.
5. Some minor fixes here and there and improving readability of some code

Fixes #210, #266, #255


pkg/protocol/commitment_verifier.go
```diff
@@ -58,27 +75,49 @@ func (c *CommitmentVerifier) verifyCommitment(commitment *model.Commitment, atte
...
+	apiForSlot := c.engine.APIForSlot(commitment.Slot())
+	commitmentEpoch := apiForSlot.TimeProvider().EpochFromSlot(commitment.Slot())
+	if commitmentEpoch > c.epoch {
+		c.epoch = commitmentEpoch
+
+		committee, exists := c.engine.SybilProtection.SeatManager().CommitteeInEpoch(commitmentEpoch)
+		if exists {
+			validatorAccounts, err := committee.Accounts()
+			if err == nil {
+				validatorAccountsData, err := c.engine.Ledger.PastAccounts(validatorAccounts.IDs(), c.lastCommonSlotBeforeFork)
```
karimodm, line 92: So if you have a committee for the new epoch you take those accounts with the data at the last common slot before the fork. If instead you don't have a committee for the new epoch you are going to keep using the previous (or initial) `validatorAccountsData`; you are therefore assuming that the committee should be re-used and it was not rotated.
If I reed the following code correctly, in either case you would need accurate account data at `lastCommonSlotBeforeFork` _**ONLY** in terms of signing keys_: to verify attestations' signatures.

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
```diff
@@ -148,64 +148,68 @@ func (o *SybilProtection) CommitSlot(slot iotago.SlotIndex) (committeeRoot iotag
...
-	if timeProvider.EpochEnd(currentEpoch) == slot+maxCommittableAge {
-		if _, committeeExists := o.seatManager.CommitteeInEpoch(nextEpoch); !committeeExists {
-			// If the committee for the epoch wasn't set before due to finalization of a slot,
-			// we promote the current committee to also serve in the next epoch.
-			committeeAccounts, err := o.reuseCommittee(currentEpoch, nextEpoch)
-			if err != nil {
-				return iotago.Identifier{}, iotago.Identifier{}, ierrors.Wrapf(err, "failed to reuse committee for epoch %d", nextEpoch)
+	// Determine the committee root.
+	{
+		// If the committed slot is `maxCommittableSlot`
```
karimodm:
```suggestion
		// If the committed slot is `maxCommittableAge`
```

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
```diff
@@ -148,64 +148,68 @@ func (o *SybilProtection) CommitSlot(slot iotago.SlotIndex) (committeeRoot iotag
...
-	var targetRewardsEpoch iotago.EpochIndex
-	if apiForSlot.TimeProvider().EpochEnd(currentEpoch) == slot {
-		targetRewardsEpoch = nextEpoch
-	} else {
-		targetRewardsEpoch = currentEpoch
-	}
+	// Determine the rewards root.
+	{
+		targetRewardsEpoch := currentEpoch
+		if slot == currentEpochEndSlot {
```
karimodm, line 200: Why would we calculate the rewards for the next epoch? I don't get this.

pkg/storage/prunable/prunable.go
```diff
@@ -142,16 +142,16 @@ func (p *Prunable) Flush() {
...
-func (p *Prunable) Rollback(targetEpoch iotago.EpochIndex, pruningRange [2]iotago.SlotIndex) error {
-	if err := p.prunableSlotStore.PruneSlots(targetEpoch, pruningRange); err != nil {
-		return ierrors.Wrapf(err, "failed to prune slots in range [%d, %d] from target epoch %d", pruningRange[0], pruningRange[1], targetEpoch)
+func (p *Prunable) Rollback(targetEpoch iotago.EpochIndex, startPruneRange iotago.SlotIndex, endPruneRange iotago.SlotIndex) error {
+	if err := p.prunableSlotStore.PruneSlots(targetEpoch, startPruneRange, endPruneRange); err != nil {
+		return ierrors.Wrapf(err, "failed to prune slots in range [%d, %d] from target epoch %d", startPruneRange, endPruneRange, targetEpoch)
 	}
 
-	if err := p.rollbackCommitteesCandidates(targetEpoch, pruningRange[0]-1); err != nil {
+	if err := p.rollbackCommitteesCandidates(targetEpoch, startPruneRange); err != nil {
```
karimodm, line 149: You removed the original `-1` from the startPruneRange. Was it intended?

pkg/testsuite/testsuite.go
```diff
@@ -285,6 +286,28 @@ func (t *TestSuite) Nodes(names ...string) []*mock.Node {
 	return nodes
 }
 
+func (t *TestSuite) AccountsOfNodes(names ...string) []iotago.AccountID {
```
karimodm, line 313: Very useful helper function :+1:

karimodm (approved):
I have to really remark the very high quality of this PR: clean code, well commented and well designed. VERY GOOD JOB!!!! :+1: :+1: :+1: :+1:

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
jonastheis: Was asking myself the same question. I just refactored that code to be easier understandable what is happening at which time because before it was basically impossible to read

pkg/storage/prunable/prunable.go
jonastheis: I fixed a bug in the condition as it was `candidacySlot < targetSlot` which lead to the wrong slots being deleted. 

And yes, it was intended because I changed it to be inclusive `candidacySlot >= deletionStartSlot`. So we delete everything starting and including `deletionStartSlot`

pkg/protocol/commitment_verifier.go
jonastheis: Yes exactly

PR #536, opened 2023-11-20
Fix trigger of StateDiffApplied event before commitment is stored

muXxer:
This PR removes the `Ledger.StateDiffApplied` event, and instead publishes the info about created and consumed outputs via `Notarization.SlotCommited` event.

This way the inx `ListenToLedgerUpdates` propagates the info to the clients after the commitment was successfully stored in the node.


alexsporn (approved):
Don't completely like this being part of SlotCommitted, but don't see another solution right now.

PR #538, opened 2023-11-21
Fix Congestion endpoint to use AccountAddress

daria305:
Update GetCongestion endpoint to use AccountAddress, as only the client was changed to bech32 address

### DIFF
--- a/components/restapi/core/accounts.go
+++ b/components/restapi/core/accounts.go
@@ -32,14 +32,14 @@ func congestionByAccountAddress(c echo.Context) (*apimodels.CongestionResponse, 
 	}
 
 	hrp := deps.Protocol.CommittedAPI().ProtocolParameters().Bech32HRP()
-	address, err := httpserver.ParseBech32AddressParam(c, hrp, restapipkg.ParameterAddress)
+	address, err := httpserver.ParseBech32AddressParam(c, hrp, restapipkg.ParameterBech32Address)
 	if err != nil {
 		return nil, err
 	}
 
 	accountAddress, ok := address.(*iotago.AccountAddress)
 	if !ok {
-		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to assert account address %s", c.Param(restapipkg.ParameterAddress))
+		return nil, ierrors.Wrapf(httpserver.ErrInvalidParameter, "address %s is not an account address", c.Param(restapipkg.ParameterBech32Address))
 	}
 
 	accountID := accountAddress.AccountID()
@@ -116,14 +116,14 @@ func validators(c echo.Context) (*apimodels.ValidatorsResponse, error) {
 
 func validatorByAccountAddress(c echo.Context) (*apimodels.ValidatorResponse, error) {
 	hrp := deps.Protocol.CommittedAPI().ProtocolParameters().Bech32HRP()
-	address, err := httpserver.ParseBech32AddressQueryParam(c, hrp, restapipkg.ParameterAddress)
+	address, err := httpserver.ParseBech32AddressParam(c, hrp, restapipkg.ParameterBech32Address)
 	if err != nil {
-		return nil, ierrors.Wrapf(err, "failed to parse account address %s", c.Param(restapipkg.ParameterAddress))
+		return nil, err
 	}
 
 	accountAddress, ok := address.(*iotago.AccountAddress)
 	if !ok {
-		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to assert account address %s", c.Param(restapipkg.ParameterAddress))
+		return nil, ierrors.Wrapf(httpserver.ErrInvalidParameter, "address %s is not an account address", c.Param(restapipkg.ParameterBech32Address))
 	}
 
 	latestCommittedSlot := deps.Protocol.MainEngineInstance().SyncManager.LatestCommitment().Slot()
--- a/components/restapi/core/accounts.go
+++ b/components/restapi/core/accounts.go
@@ -16,7 +16,7 @@ import (
 	"github.com/iotaledger/iota.go/v4/nodeclient/apimodels"
 )
 
-func congestionForAccountID(c echo.Context) (*apimodels.CongestionResponse, error) {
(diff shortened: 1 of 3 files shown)


components/restapi/core/accounts.go
```diff
@@ -36,6 +31,18 @@ func congestionForAccountID(c echo.Context) (*apimodels.CongestionResponse, erro
 		}
 	}
 
+	hrp := deps.Protocol.CommittedAPI().ProtocolParameters().Bech32HRP()
+	address, err := httpserver.ParseBech32AddressParam(c, hrp, restapipkg.ParameterBech32Address)
+	if err != nil {
+		return nil, ierrors.Wrapf(err, "failed to parse bech32 address %s", c.Param(restapipkg.ParameterBech32Address))
+	}
+
+	accountAddress, ok := address.Clone().(*iotago.AccountAddress)
```
alexsporn: why are you cloning here?

components/restapi/core/accounts.go
```diff
@@ -36,6 +31,18 @@ func congestionForAccountID(c echo.Context) (*apimodels.CongestionResponse, erro
 		}
 	}
 
+	hrp := deps.Protocol.CommittedAPI().ProtocolParameters().Bech32HRP()
+	address, err := httpserver.ParseBech32AddressParam(c, hrp, restapipkg.ParameterBech32Address)
+	if err != nil {
+		return nil, ierrors.Wrapf(err, "failed to parse bech32 address %s", c.Param(restapipkg.ParameterBech32Address))
```
muXxer: I wouldn't wrap the error any further. the message from httpserver is already enough.

pkg/restapi/restapi.go
```diff
@@ -30,6 +30,9 @@ const (
 	// ParameterAccountID is used to identify an account by its ID.
 	ParameterAccountID = "accountID"
```
alexsporn: this can be removed. Please also adapt `RouteValidatorsAccount` to use an address

components/restapi/core/accounts.go
muXxer: Why do you need to clone?

muXxer (changes requested):

pkg/restapi/restapi.go
```diff
@@ -30,6 +30,9 @@ const (
...
 
+	// ParameterBech32Address is used to identify an account address by its bech32 address.
```
muXxer: It's not only account addresses. It's all adresses.

components/restapi/core/accounts.go
```diff
@@ -107,20 +114,29 @@ func validators(c echo.Context) (*apimodels.ValidatorsResponse, error) {
 	return resp, nil
 }
 
-func validatorByAccountID(c echo.Context) (*apimodels.ValidatorResponse, error) {
-	accountID, err := httpserver.ParseAccountIDParam(c, restapipkg.ParameterAccountID)
+func validatorByAccountAddress(c echo.Context) (*apimodels.ValidatorResponse, error) {
+	hrp := deps.Protocol.CommittedAPI().ProtocolParameters().Bech32HRP()
+	address, err := httpserver.ParseBech32AddressQueryParam(c, hrp, restapipkg.ParameterAddress)
```
alexsporn: This parses a query param, not a path param

alexsporn: Use `ParseBech32AddressParam` instead

components/restapi/core/accounts.go
```diff
@@ -36,6 +31,18 @@ func congestionForAccountID(c echo.Context) (*apimodels.CongestionResponse, erro
...
 
+	hrp := deps.Protocol.CommittedAPI().ProtocolParameters().Bech32HRP()
+	address, err := httpserver.ParseBech32AddressParam(c, hrp, restapipkg.ParameterAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	accountAddress, ok := address.(*iotago.AccountAddress)
+	if !ok {
+		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to assert account address %s", c.Param(restapipkg.ParameterAddress))
```
muXxer: I would say that is not an internal server error, but a user error. The user passed the wrong address type here.
`ErrInvalidParameter`
```suggestion
		return nil, ierrors.Wrapf(httpserver.ErrInvalidParameter, "address %s is not an account address", c.Param(restapipkg.ParameterAddress))
```

components/restapi/core/accounts.go
```diff
@@ -107,20 +114,29 @@ func validators(c echo.Context) (*apimodels.ValidatorsResponse, error) {
...
 }
 
-func validatorByAccountID(c echo.Context) (*apimodels.ValidatorResponse, error) {
-	accountID, err := httpserver.ParseAccountIDParam(c, restapipkg.ParameterAccountID)
+func validatorByAccountAddress(c echo.Context) (*apimodels.ValidatorResponse, error) {
+	hrp := deps.Protocol.CommittedAPI().ProtocolParameters().Bech32HRP()
+	address, err := httpserver.ParseBech32AddressQueryParam(c, hrp, restapipkg.ParameterAddress)
 	if err != nil {
-		return nil, ierrors.Wrapf(err, "failed to parse account ID %s", c.Param(restapipkg.ParameterAccountID))
+		return nil, ierrors.Wrapf(err, "failed to parse account address %s", c.Param(restapipkg.ParameterAddress))
```
muXxer:
```suggestion
		return nil, err
```

components/restapi/core/accounts.go
```diff
@@ -107,20 +114,29 @@ func validators(c echo.Context) (*apimodels.ValidatorsResponse, error) {
...
 	}
+
+	accountAddress, ok := address.(*iotago.AccountAddress)
+	if !ok {
+		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to assert account address %s", c.Param(restapipkg.ParameterAddress))
```
muXxer:
```suggestion
		return nil, ierrors.Wrapf(httpserver.ErrInvalidParameter, "address %s is not an account address", c.Param(restapipkg.ParameterAddress))
```

muXxer (approved):

alexsporn (approved):

PR #543, opened 2023-11-22
 Block time monotonicity check in filter

jonastheis:
This PR adds a check for increasing block time monotonicity to the filter (after a block is solid) and adds a test for it.

It also renames commitment filter to post solid filter and filter (block filter) to pre solid filter to make it easier understandable.


pkg/protocol/engine/blockdag/inmemoryblockdag/blockdag.go
```diff
@@ -203,8 +203,11 @@ func (b *BlockDAG) shouldAttach(data *model.Block) (shouldAttach bool, err error
 	}
 
 	// We already attached it before, but the parents are invalid, then we set the block as invalid.
-	if parentsValid, err := b.canAttachToParents(data); !parentsValid && storedBlock != nil {
-		storedBlock.SetInvalid()
+	if parentsValid, err := b.canAttachToParents(data); !parentsValid {
+		if storedBlock != nil {
```
piotrm50, line 207: Can it ever be nil? Above we check if storedBlockExists.

piotrm50 (approved):

pkg/protocol/engine/blockdag/inmemoryblockdag/blockdag.go
jonastheis: It is only non-nil if the block was marked as missing before or we attached it before. Otherwise it's always nil

pkg/protocol/engine/commitmentfilter/accountsfilter/commitmentfilter.go
```diff
@@ -58,6 +64,36 @@ func (c *CommitmentFilter) ProcessPreFilteredBlock(block *blocks.Block) {
 }
 
 func (c *CommitmentFilter) evaluateBlock(block *blocks.Block) {
+	// Block issuing time monotonicity: a block's issuing time needs to be greater than its parents issuing time.
+	for _, parentID := range block.Parents() {
+		var parentIssuingTime time.Time
+
+		if parentID == iotago.EmptyBlockID {
```
alexsporn: this is wrong. Genesis is not the `EmptyBlockID`. You can fetch the genesis blockID from the ProtocolParameters

piotrm50 (approved):

pkg/protocol/engine/commitmentfilter/accountsfilter/commitmentfilter.go
jonastheis: Removed this entire check. I realized that we actually do have the issuing time in the root blocks. Therefore, it is simpler and more performant to just use blocks from the block cache instead.

piotrm50 (approved):

PR #546, opened 2023-11-22
Fix and cleanup the REST API

muXxer:
This PR fixes a bug in inx-app that caused sendBlock to fail if a JSON payload was received.

It also uses the new constants from nodeclient, so we don't have to duplicate the endpoints all the time in iota-core and iota.go.

It also includes the fix for `omitempty` in `iota.go`.


components/restapi/core/blocks.go
```diff
@@ -85,7 +85,7 @@ func blockIssuance() (*apimodels.IssuanceBlockHeaderResponse, error) {
 func sendBlock(c echo.Context) (*apimodels.BlockCreatedResponse, error) {
 	iotaBlock, err := httpserver.ParseRequestByHeader(c, deps.Protocol.CommittedAPI(), iotago.BlockFromBytes(deps.Protocol))
 	if err != nil {
-		return nil, ierrors.Wrapf(err, "failed to parse iotablock")
+		return nil, ierrors.Wrapf(httpserver.ErrInvalidParameter, "failed to parse request: %w", err)
```
jkrvivian: `ErrInvalidParameter` or `ErrUnsupportedMediaType` is wrapped in the `err` from httpserver, we can wrap `err` directly.

muXxer: lol, you are right! :D
I commented the same in another PR and then I do the same mistake :sweat_smile: :D

jkrvivian (approved):

alexsporn (approved):

muXxer:
The docker network is broken because we have hardcoded values in the test framework. Since this is not caused by this PR, we will fix it in another one.

PR #547, opened 2023-11-22
Increase storage deposit parameters in the presets

muXxer:
The client team is confused about their implementation, because we are so cheap compared to the mainnet :trollface:


thibault-martinez (approved):
:trollface:

jonastheis (approved):

PR #549, opened 2023-11-23
Add Transaction Metadata endpoint

### LINKED ISSUE #456: Check if transaction metadata endpoint makes sense
IMO users sometimes don't care about the blocks which contain a transaction, but only about the transaction itself.
If it's possible, we should add a tx metadata endpoint.

jkrvivian:
Add `/transactions/{txID}/metadata` endpoint to return the state and failure reason of requested transaction.

Based on https://github.com/iotaledger/iota.go/pull/620

Close #456


components/restapi/core/blocks.go
```diff
@@ -33,6 +33,20 @@ func blockMetadataByBlockID(blockID iotago.BlockID) (*api.BlockMetadataResponse,
...
 
+func transactionMetadataByBlockID(blockID iotago.BlockID) (*api.TransactionMetadataResponse, error) {
+	blockMetadata, err := deps.Protocol.MainEngineInstance().Retainer.BlockMetadata(blockID)
+	if err != nil {
+		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to get block metadata %s: %s", blockID.ToHex(), err)
+	}
+
+	metadata := blockMetadata.TransactionMetadataResponse()
+	if metadata == nil {
+		return nil, ierrors.Wrapf(echo.ErrNotFound, "transaction not found")
```
muXxer, line 44: Please add the blockid here.
```suggestion
		return nil, ierrors.Wrapf(echo.ErrNotFound, "no transaction found for block ID %s", blockID.ToHex())
```

muXxer (changes requested):

muXxer (approved):

PR #550, opened 2023-11-23
Fix calculation of preset values

muXxer:
In #547 the storage deposit parameters were changed.

This caused the docker network to fail, because the values for the snapshot accounts were hardcoded and that caused "create snapshot" to panic due to insufficient balance.

This PR fixes the docker network, but the actual tests are also wrong. I need to talk to @cyberphysic4l how to fix that properly.

I would like to merge this PR with temporary workarounds in the tests, so that the client team can use the docker network in the meantime, and the tests pass.


tools/genesis-snapshot/presets/presets.go
```diff
@@ -15,238 +15,242 @@ import (
...
-			BlockIssuanceCredits: iotago.MaxBlockIssuanceCredits / 4,
-			StakingEpochEnd:      iotago.MaxEpochIndex,
-			FixedCost:            1,
-			StakedAmount:         mock.MinValidatorAccountAmount,
-			Mana:                 iotago.Mana(mock.MinValidatorAccountAmount),
-		},
-		snapshotcreator.AccountDetails{
-			/*
-				node-03-validator
+	// Feature is a preset for the feature network, genesis time ~20th of July 2023.
```
alexsporn: This sentence makes no sense, since we do `time.Now().Unix()-100_000`

muXxer: fixed

PhilippGackstatter (approved):

Issue #466, opened 2023-10-26
More accounts & validator tests


cyberphysic4l:
More end to end accounts tests are required to ensure everything is working in the VM and accounts manager. 

The tests should focus on user stories, with sequences of actions that we expect for account holders of all kinds. [This PR](https://github.com/iotaledger/iota-core/pull/448) introduces a wallet-centric framework for testing which should facilitate these sequences of account manipultations more intuitively.

PhilippGackstatter:
- Test that Reward Inputs can point to an account id that does not exist or has no rewards in the rewards data. When fetching rewards for accounts that don't exist (or have no rewards) the node should return 0 rather than an error. This is because reward inputs are **required** for removing a Staking Feature from an Account or for destroying a Delegation Output. Otherwise, if I delegated to some address that doesn't exist, I could not destroy my delegation output because the reward would error. (I haven't verified whether this is already implemented as-is or not, but it should be tested regardless).
- Test that Reward Inputs can only point to Accounts & Delegation Outputs. I tried to quickly find that this is checked somewhere but couldn't, so perhaps it's missing.
- Test that a transaction fails whenever the condition for claiming is not given. That is, coming up with scenarios where claiming is **not** allowed and try to claim (remove staking feature, destroy delegation output).
- Test the interplay of accounts with delegation in iota-core, e.g.:

```
// Step 1

Create Delegation Output A delegating to X

// Step 2 (separate epoch)

Create Account X with Staking Feature
Create Delegation Output B delegating to X
Create Delegation Output C delegating to X
Create Delegation Output D delegating to X

In this test X is never selected as a validator. Any Delegation Output that is destroyed should claim 0 rewards.

Assert expectations: X's total stake should be X + A + B + C + D. (I.e. make sure that A is counted even though it was created before X created its staking feature).

// Step 3

Destroy A
Create Delegation Output E delegating to X

Assert expectations: X's total stake should be X + B + C + D + E.

// Step 4

Transition B to Delayed Claiming.
Destroy C.

Assert expectations: X's total stake should be X + D + E.

// Step 5

Destroy Account.

Assert expectations: X should be removed from the "staking tracking data" (not sure what it's called exactly).

// Step 6

Destroy B, D & E.

Assert expectations.
```

Further variations of this test may be to check that removing the staking feature and re-adding it in the same transaction also correctly sets the rewards data and changed start and end epoch of the staker.

alexsporn:
[...]

PR #551, opened 2023-11-23
Add Reward Input tests

PhilippGackstatter:
Add tests that use the Reward Input in various scenarios.

part of #466.


pkg/tests/reward_test.go
```diff
@@ -0,0 +1,233 @@
...
+		}),
+		mock.WithOutputs(iotago.Outputs[iotago.Output]{nftOutput}),
+		mock.WithAllotAllManaToAccount(ts.CurrentSlot(), ts.DefaultWallet().BlockIssuer.AccountID),
+	)
+
+	ts.IssueBasicBlockWithOptions("block2", ts.DefaultWallet(), tx2, mock.WithStrongParents(latestParents...))
+
+	ts.Wait(node1, node2)
+
+	// TODO: Assertions do not pass for node2 - why?
```
cyberphysic4l: This one is probably worth figuring out, anything interesting in the logs?

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -390,7 +389,7 @@ func (w *Wallet) SendFundsFromAccount(transactionName string, accountOutputName
 		},
 		Features: iotago.BasicOutputFeatures{},
 	}}
-	signedTransaction := w.createSignedTransactionWithOptions(
+	signedTransaction := w.CreateSignedTransactionWithOptions(
```
cyberphysic4l: My reason for keeping this method private was to declutter the tests themselves a bit by creating methods methods that create specific types of transactions as the public methods used in the tests.

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -424,21 +423,31 @@ func (w *Wallet) ClaimValidatorRewards(transactionName string, inputName string)
 		panic(fmt.Sprintf("failed to calculate reward for output %s: %s", inputName, err))
 	}
 
-	signedTransaction := w.createSignedTransactionWithOptions(
+	accountOutput := builder.NewAccountOutputBuilderFromPrevious(inputAccount).
+		RemoveFeature(iotago.FeatureStaking).
+		Mana(input.StoredMana() + rewardMana).
```
cyberphysic4l: This doesn't take decay and potential Mana into account. It could cause transaction to fail in some cases if the stored Mana was large and it was decayed significantly.

cyberphysic4l (commented):
I would prefer to keep the `createSignedTransactionWithOptions()` method private and create methods for specific things we want to do. It just moves some of the complexity of the tests.

pkg/tests/reward_test.go
cyberphysic4l: We can leave this as is for now and add a TODO here referencing this issue https://github.com/iotaledger/iota-core/issues/580
And we check for failure at all nodes once that issue is closed rather than just node1

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -488,33 +500,66 @@ func (w *Wallet) ClaimDelegatorRewards(transactionName string, inputName string)
...
 		}),
-		WithAllotAllManaToAccount(w.currentSlot, w.BlockIssuer.AccountID),
+		WithOutputs(outputStates),
 	)
 
 	return signedTransaction
 }
 
+// Computes the Potential Mana that the output generates until target slot.
+func (w *Wallet) PotentialMana(api iotago.API, input *utxoledger.Output, targetSlot iotago.SlotIndex) iotago.Mana {
```
cyberphysic4l: This is nice. It would be nice to also do the same style of function for stored Mana rather where you just return `lo.PanicOnErr(apiForSlot.ManaDecayProvider().ManaWithDecay(input.StoredMana(), input.SlotCreated(), w.currentSlot))`

cyberphysic4l (approved):
Just my two suggestions
- creating a `w.StoredMana()` method for the wallet like the potential mana one you made.
- adding a comment linked to the issue for the TODO in rewards pointing to NFT test.

PR #557, opened 2023-11-24
Cleanup tests

muXxer:
This PR fixes the ugly workarounds from #550.

It adds `WalletOptions` to the testsuite as well as a `depositcalculator` package.

### DIFF
--- a/pkg/tests/accounts_test.go
+++ b/pkg/tests/accounts_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/iotaledger/iota-core/pkg/protocol/engine/accounts"
 	"github.com/iotaledger/iota-core/pkg/protocol/engine/blocks"
 	"github.com/iotaledger/iota-core/pkg/testsuite"
+	"github.com/iotaledger/iota-core/pkg/testsuite/depositcalculator"
 	"github.com/iotaledger/iota-core/pkg/testsuite/mock"
 	"github.com/iotaledger/iota-core/pkg/testsuite/snapshotcreator"
 	"github.com/iotaledger/iota-core/pkg/utils"
@@ -220,7 +221,14 @@ func Test_StakeDelegateAndDelayedClaim(t *testing.T) {
 	// set the expiry slot of the transitioned genesis account to the latest committed + MaxCommittableAge
 	newAccountExpirySlot := node1.Protocol.MainEngineInstance().Storage.Settings().LatestCommitment().Slot() + ts.API.ProtocolParameters().MaxCommittableAge()
 
-	validatorAccountAmount := mock.MinValidatorAccountAmount(ts.API.ProtocolParameters())
+	stakedAmount := iotago.BaseToken(10000)
+
+	validatorAccountAmount, err := depositcalculator.MinDeposit(ts.API.ProtocolParameters(), iotago.OutputAccount,
+		depositcalculator.WithAddress(&iotago.Ed25519Address{}),
+		depositcalculator.WithBlockIssuerKeys(1),
+		depositcalculator.WithStakedAmount(stakedAmount),
+	)
+	require.NoError(t, err)
 
 	var block1Slot iotago.SlotIndex = 1
 	tx1 := ts.DefaultWallet().CreateAccountFromInput(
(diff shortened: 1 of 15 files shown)


pkg/tests/accounts_test.go
```diff
@@ -221,16 +221,23 @@ func Test_StakeDelegateAndDelayedClaim(t *testing.T) {
...
 	tx1 := ts.DefaultWallet().CreateAccountFromInput(
 		"TX1",
 		"Genesis:0",
 		ts.DefaultWallet(),
 		mock.WithBlockIssuerFeature(iotago.BlockIssuerKeys{newAccountBlockIssuerKey}, newAccountExpirySlot),
-		mock.WithStakingFeature(10000, 421, 0, 10),
-		// TODO: Temporary "fix" for the tests, lets fix this in another PR, so we can at least use the docker network again
-		//mock.WithAccountAmount(mock.MinIssuerAccountAmount(ts.API.ProtocolParameters())),
-		mock.WithAccountAmount(mock.MinValidatorAccountAmount(ts.API.ProtocolParameters())),
+		mock.WithStakingFeature(stakedAmount, 421, 0, 10), // match amount and staked amount to simplify the tests
```
PhilippGackstatter:
```suggestion
		mock.WithStakingFeature(stakedAmount, 421, 0, 10),
```

That comment might not always be accurate anymore if the parameters change, so better remove it.

pkg/testsuite/depositcalculator/depositcalculator.go
```diff
@@ -0,0 +1,334 @@
...
+			Entries: iotago.StateMetadataFeatureEntries{
+				// configured length - map length prefix - key length prefix - value length prefix
+				"": make([]byte, opts.StateMetadataLength-int(serix.LengthPrefixTypeAsByte)+int(serix.LengthPrefixTypeAsByte)+int(serix.LengthPrefixTypeAsUint16)),
+			},
+		})
+	}
+
+	if opts.TagLength > 0 {
+		features = append(features, &iotago.TagFeature{Tag: make([]byte, opts.TagLength)})
+	}
```
PhilippGackstatter, line 213: For the metdata, the length we pass in is what the entire metadata feature ends up having in size. For the Tag Feature we'd have length + length prefix which is a bit inconsistent.

So I'd suggest to subtract the length prefix byte here.

If you do that, also update the documentation on `WithTagLength`.

pkg/testsuite/depositcalculator/depositcalculator.go
```diff
@@ -0,0 +1,335 @@
...
+	case iotago.OutputBasic, iotago.OutputAccount, iotago.OutputNFT, iotago.OutputDelegation:
+		unlockConditions = append(unlockConditions, iotago.UnlockCondition(&iotago.AddressUnlockCondition{Address: opts.Address}).(T))
+
+	case iotago.OutputAnchor:
+		unlockConditions = append(unlockConditions, iotago.UnlockCondition(&iotago.StateControllerAddressUnlockCondition{Address: opts.StateControllerAddress}).(T))
+		unlockConditions = append(unlockConditions, iotago.UnlockCondition(&iotago.GovernorAddressUnlockCondition{Address: opts.GovernorAddress}).(T))
+
+	case iotago.OutputFoundry:
+		unlockConditions = append(unlockConditions, iotago.UnlockCondition(&iotago.ImmutableAccountUnlockCondition{Address: &iotago.AccountAddress{}}).(T))
+	}
```
PhilippGackstatter, line 165: There are a bunch of scenarios where this would not yield the correct deposit if the mandatory address(es) is/are not set in the options. It would just silently return a too small value, because the serialized size would be smaller than on an actual output with the address set.
Either panic here, which is fine as long as we keep this as test-only code, or return an error if we want to move it to non-testing code later. I would go with the panics for now, that's simplest.

muXxer: No, it panics if you don't pass the correct addresses for the outputs, because the StorageScore code panics.
It guess it's fine to rely on those panics without any change?

pkg/tests/accounts_test.go
muXxer: Removed

pkg/testsuite/depositcalculator/depositcalculator.go
muXxer: I renamed the other funcs to clarify the difference.

PhilippGackstatter (approved):

PR #568, opened 2023-11-27
Fix: WarpSync

### LINKED ISSUE #532: Warpsync catch-up in live network with spam

### LINKED ISSUE #565: Solid blocks are re-requested and not found in block cache
In `TestLossOfAcceptanceFromGenesis` the following occurred:

Image

It seems that blocks get solid and then are not found in the block cache anymore.

When looking at the entire [log.txt](https://github.com/iotaledger/iota-core/files/13474445/log.txt) of the test, it seems that the node was warpsycing and solidifying the tangle from the tip at the same time. Which might lead to resetting the engine within the warpsync process. Thus, leading to some weird errors at any point in the data flow. 

This might be addressed with #293 as there's different modes for warpsync and normal mode operation.

hmoog:


pkg/protocol/protocol_warp_sync.go
```diff
@@ -213,12 +225,9 @@ func (w *WarpSyncProtocol) ProcessResponse(commitmentID iotago.CommitmentID, blo
 						targetEngine.BlockGadget.SetAccepted(block)
 
 						block.Notarized().OnUpdate(func(_ bool, _ bool) {
-							// Wait for all blocks to be notarized before forcing the commitment of the slot.
-							if notarizedBlocks.Add(1) != totalBlocks {
-								return
+							if notarizedBlocks.Add(1) == totalBlocks {
```
jonastheis: Does this work in the case where there's blocks without a future cone in a slot? ie when we revive the chain? I think we should force accept the blocks we didn't accept within `minCommittableAge` and then mark it as fully booked (or by force accepting those blocks it'd become fully booked automatically I guess).

karimodm: Blocks without a future cone can become naturally booked, and therefore the slot will become fully booked. This is because booked means that the past cone needs to exist, the future cone doesn't matter there.
On this specific line we check for all the blocks to be _notarized_, meaning accepted; and we force accept all the blocks we have received via warpsync on line 225.

pkg/protocol/commitment.go
```diff
@@ -261,9 +276,9 @@ func (c *Commitment) deriveRequestAttestations(chain *Chain, parent *Commitment)
 // deriveWarpSyncBlocks derives the WarpSyncBlocks flag of this Commitment which is true if our Chain is requesting
```
jonastheis, line 276: Comment should be adjusted

pkg/protocol/protocol_blocks.go
```diff
@@ -53,21 +53,12 @@ func newBlocksProtocol(protocol *Protocol) *BlocksProtocol {
...
-					return lo.Batch(
-						engine.Events.Scheduler.BlockScheduled.Hook(func(block *blocks.Block) { b.SendResponse(block.ModelBlock()) }).Unhook,
-						engine.Events.Scheduler.BlockSkipped.Hook(func(block *blocks.Block) { b.SendResponse(block.ModelBlock()) }).Unhook,
-					)
-				})
-			}
+		protocol.Chains.WithInitializedEngines(func(chain *Chain, engine *engine.Engine) (shutdown func()) {
+			return lo.Batch(
+				engine.Events.BlockRequester.Tick.Hook(b.SendRequest).Unhook,
+				engine.Events.Scheduler.BlockScheduled.Hook(func(block *blocks.Block) { b.SendResponse(block.ModelBlock()) }).Unhook,
```
jonastheis, line 59: This means we will also gossip blocks from our candidate engine, no? Is that something we want?

pkg/protocol/protocol_warp_sync.go
```diff
@@ -157,93 +168,133 @@ func (w *WarpSyncProtocol) ProcessResponse(commitmentID iotago.CommitmentID, blo
 			}
 
 			// make sure the engine is clean and requires a warp-sync before we start processing the blocks
```
jonastheis: Wrong comment

jonastheis (approved):

pkg/protocol/protocol_blocks.go
jonastheis: This can be addressed in #591

PR #600, opened 2023-12-05
Implement Committee Rotation Tests with docker-network

### LINKED ISSUE #559: Committee rotation integration tests
These scenarios should tests committee rotation, top stakers selection and all the involved mechanisms like candidacy announcement, committee reuse, etc in integration tests. We need to run a network to test the entire stack of protocol + inx plugins, specifically `inx-validator`.

### Setup

Relevant ProtocolParameters:
```
committeeTargetSize=4
slotsPerEpochExponent=4
minCommittableAge=2
maxCommittableAge=4
epochNearingThreshold=8
```

Validators: 5 validators, E0, E1, E2, E3, V (stake in order from high to low)

## 1. Staking and delegation lead to committee selection

1. Run (Docker) network
2. Do staking transaction (with higher stake than lowest existing committee member E3) for new validator V
3. Start inx-validator for V and issue CandidacyPayload
4. Check that V becomes part of committee in next or epoch after next
5. Issue delegation transaction so that stake of E3 > V
6. Check that E3 becomes part of committee again in favor of V

## 2. committee rotation to a smaller committee than targetCommitteeSize

1. Run (Docker) network
2. Shut down inx-validator of E1
3. Check that committee of size 3 is selected in next epoch

## 3. committee reuse due to no candidates announced but slot finalized at epochNearingThreshold

1. Run (Docker) network
2. Restart inx-validator with IssueCandidacyPayload=false
3. Check that committee is reused in next epoch due to no candidates but finalization

## 4. committee reuse due to no slot finalization at epochNearingThreshold and recovery after finalization comes back

1. Run (Docker) network
2. Shut down inx-validator of E1 and E2
3. Check that committee is reused in next epoch due to no finalization
4. Start inx-validator of E1 
5. Check that committee of size 3 is selected in next epoch and finalization occurs again from that epoch

jkrvivian:
* Implement DockerTestFramework to test scenarios in a docker network.
   * Enable creating snapshot with provided protocol parameters, snapshot options.
   * Enable stop/restart containers within the test
   * Dump container logs to files for each test, in `docker-network/tests/logs`
* Implement committee rotation tests from #559 

To run the test, add `dockertests` tag to command line:
e.g., `go test ./... -tags rocksdb,dockertests -v`


tools/docker-network/tests/dockerframework.go
```diff
@@ -0,0 +1,339 @@
...
+
+func (d *DockerTestFramework) GetRandomNode() *nodeclient.Client {
+	for _, client := range d.nodeClients {
+		return client
+	}
+
+	return nil
+}
+
+func (d *DockerTestFramework) AssertCommitteeSelection(expectedEpoch iotago.EpochIndex, expectedCommitteeSize int) {
```
jonastheis: Maybe rename to `AssertCommitteeSize`? And if we have the validators' accounts, we should probably better check for the exact committee to make sure the expected candidates have been selected

tools/docker-network/tests/committeerotation_test.go
```diff
@@ -0,0 +1,118 @@
...
+	require.NoError(t, err)
+
+	node := d.GetRandomNode()
+	status, err := getNodeStatus(node)
+	require.NoError(t, err)
+
+	currentEpoch := node.CommittedAPI().TimeProvider().EpochFromSlot(status.LatestAcceptedBlockSlot)
+
+	// stop validator 2
+	err = d.StopContainer("V2")
```
jonastheis: It's not really clear that this method stops the inx-container.

tools/docker-network/tests/committeerotation_test.go
```diff
@@ -0,0 +1,118 @@
...
+	require.NoError(t, err)
+
+	d.AssertCommitteeSelection(currentEpoch+2, 3)
```
jonastheis: As I commented on the function: maybe it would be better to assert the committee explicitly to make sure the members are in there.

tools/docker-network/tests/committeerotation_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"V4": "docker-network-inx-validator-4-1",
+}
+
+var ClientURLs = map[string]string{
+	"V1":    "http://localhost:8050",
+	"V2":    "http://localhost:8060",
+	"V3":    "http://localhost:8070",
+	"V4":    "http://localhost:8040",
+	"node5": "http://localhost:8090",
+}
```
jonastheis: Instead of having these separate we could create a struct `Node` that holds all this information. Again, this could be similar (but simplified) as in the testsuite. Then we could have some convenience functions that gives you the name of a node's validator container which you could then hand to the `StopContainer` method. 

Or another thing could be `d.AccountsFromNodes(d.Nodes("V1","V2"))` which we could then hand into `AssertCommitteeSelection`.

This should be a bit cleaner and easier to use.

jonastheis (approved):

PR #603, opened 2023-12-05
Add stake delegation test

PhilippGackstatter:
Add test with multiple delegation outputs in various states pointing to an account to ensure the staked and delegated amounts are correctly calculated.

- This TODO was removed since this case is now tested.

```
// TODO: do we have a testcase that checks transitioning a delegation output twice in the same slot?
```

- Additionally, the "the delegation output was created and destroyed in the same slot => do not track the delegation as newly created" case was removed, because we never enter this code path. `stateDiff.CreatedStates()` only returns outputs that were created and not consumed by the end of that slot. Hence if a Delegation Output is created and destroyed in the same slot, it will not be returned by `CreatedStates()`. 

Part of #466.


cyberphysic4l (approved):
lgtm

PR #620, opened 2023-12-07
Add Validator tests

### LINKED ISSUE #466: More accounts & validator tests

PhilippGackstatter:
Adds end-to-end validator tests implemented in the testsuite. This ensures we can issue blocks very precisely and test the protocol's behavior in a fine-grained way.

- Fixes an overallocation of memory in performance aggregation. The `make` call was missing a `0,` to only allocate the capacity of the slice, not also its length. It effectively ended up needing twice as much memory as would have been sufficient.
- The `ValidatorReward` function calculates the rewards of a validator. It has one issue:
    - It decays the rewards by 1 epoch even when claiming rewards that only became available in this epoch, which is odd. The current epoch is the first epoch to claim rewards for the last epoch so no "decay punishment" should be applied. Hence, one fewer epoch of decay is applied. Approved by @oliviasaa.
- A logger was added to the performance tracker and logs validators performance and rewards when an epoch ends.

Subsequently it would make sense to also setup validator reward tests with the setup from #600, in order to test inx-validator, but testing only with the setup from that PR would test too many parts at once, hence the implementation in the testsuite.


pkg/tests/validator_test.go
```diff
@@ -0,0 +1,381 @@
...
+	"github.com/iotaledger/hive.go/runtime/options"
+	"github.com/iotaledger/iota-core/pkg/testsuite"
+	"github.com/iotaledger/iota-core/pkg/testsuite/mock"
+	"github.com/stretchr/testify/require"
+
+	iotago "github.com/iotaledger/iota.go/v4"
+)
+
+// IOTA Mainnet Max Supply.
+const MAX_SUPPLY = iotago.BaseToken(4_600_000_000_000_000)
```
cyberphysic4l, line 20: I think it's better to use a local variable in the testsuite setup and then use the protocol params token supply.

pkg/tests/validator_test.go
```diff
@@ -0,0 +1,381 @@
...
+	ts.AddDefaultWallet(vnode1)
+
+	ts.Run(true)
+
+	// For debugging set the log level appropriately.
+	vnode1.Protocol.SetLogLevel(log.LevelInfo)
+	vnode2.Protocol.SetLogLevel(log.LevelError)
+	node3.Protocol.SetLogLevel(log.LevelError)
+
+	ts.SetCurrentSlot(1)
```
cyberphysic4l: Should we actually set the default CurrentSlot to be 1 when we create a new testsuite? Or rather set it to GenesisSlot + 1?

pkg/tests/validator_test.go
```diff
@@ -0,0 +1,381 @@
...
+
+func validatorTest(t *testing.T, test ValidatorTest) {
+	ts := test.ts
+
+	tip := ts.DefaultWallet().Node.Protocol.Engines.Main.Get().TipSelection.SelectTips(1)[iotago.StrongParentType][0]
+	startEpoch := iotago.EpochIndex(0)
+	endEpoch := iotago.EpochIndex(len(test.epochPerformanceFactors) - 1)
+	fmt.Printf("ValidatorTest: startEpoch=%d, endEpoch=%d\n", startEpoch, endEpoch)
+
+	subslotDuration := time.Duration(ts.API.ProtocolParameters().SlotDurationInSeconds()/test.validationBlocksPerSlot) * time.Second
```
cyberphysic4l: This is technically not the subSlotDuration because it is based on the test.validationBlocksPerSlot which is the value used by the validator rather than the protocol param. So this variable should be called issuancePeriod or something along those lines.

pkg/tests/validator_test.go
```diff
@@ -0,0 +1,381 @@
...
+	return ts
+}
+
+type EpochPerformanceMap = map[iotago.EpochIndex]uint64
+type ValidatorTest struct {
+	ts *testsuite.TestSuite
+
+	// How many validation blocks per slot the validator test nodes are supposed to issue.
+	// Can be set to a different value than the protocol parameter to model under- or overissuance.
+	validationBlocksPerSlot uint8
```
cyberphysic4l: perhaps name this actualValidationBlocksPerSlot to avoid confusion with the protocol parameter.

pkg/tests/validator_test.go
```diff
@@ -0,0 +1,381 @@
...
+
+		actualRewards[accountID] = rewardMana
+	}
+
+	for accountID, actualReward := range actualRewards {
+		lastRewardEpoch := iotago.EpochIndex(len(test.epochPerformanceFactors))
+		rewards := make([]reward, 0, lastRewardEpoch)
+		for epoch := iotago.EpochIndex(0); epoch < lastRewardEpoch; epoch++ {
+			epochPerformanceFactor := test.epochPerformanceFactors[epoch]
+			reward := calculateEpochReward(t, ts, accountID, epoch, epochPerformanceFactor, totalStake, totalValidatorStake)
```
cyberphysic4l: perhaps call this variable epochReward

pkg/tests/validator_test.go
```diff
@@ -0,0 +1,381 @@
...
+			rewards = append(rewards, reward)
+		}
+
+		expectedReward := calculateValidatorReward(t, ts, accountID, rewards, startEpoch, claimingEpoch)
+
+		require.Equal(t, expectedReward, actualReward, "expected reward for account %s to be %d, was %d", accountID, expectedReward, actualReward)
+	}
+}
+
+type reward struct {
```
cyberphysic4l: parhaps the type should also be called epochReward, and the slice used as input to the calculateValidatorRewards() should be called epochRewards

pkg/tests/validator_test.go
```diff
@@ -0,0 +1,381 @@
...
+	if err != nil || !exists {
+		t.Fatal(exists, err)
+	}
+
+	profitMarginExponent := ts.API.ProtocolParameters().RewardsParameters().ProfitMarginExponent
+	stakedAmount := accountData.ValidatorStake
+	fixedCost := accountData.FixedCost
+	poolStake := accountData.ValidatorStake + accountData.DelegationStake
+	rewardEpoch := startEpoch
+	decayedRewards := iotago.Mana(0)
```
cyberphysic4l, line 358:
```suggestion
	rewardEpoch := startEpoch
	decayedRewards := iotago.Mana(0)
```

pkg/tests/validator_test.go
```diff
@@ -0,0 +1,381 @@
...
+
+	for _, reward := range rewards {
```
cyberphysic4l:
```suggestion

	rewardEpoch := startEpoch
	decayedRewards := iotago.Mana(0)
	for _, reward := range rewards {
```

cyberphysic4l (commented):
Really like the setup of these tests, my only suggestions are some renames to improve clarity.

cyberphysic4l (approved):

PR #630, opened 2023-12-12
Feat: Add metastability breaking mechanism

### LINKED ISSUE #615: Metastability breaking on equal-weight forks
Currently if two chains have equal weight we cannot resolve it. We need some deterministic way to resolve those.

hmoog:
This PR adds a metastability breaking mechanism to the candidate selection that allows nodes to converge when chains have exactly the same weight, which works by comparing the IDs of the commitments where both chains diverged.


pkg/protocol/attestations.go
```diff
@@ -194,7 +194,7 @@ func (a *Attestations) processResponse(commitment *model.Commitment, attestation
 			}
 
 			return actualWeight
-		}) > 0 {
```
jonastheis, line 197: What exactly does this change do?

hmoog: Compute returns the previous value so currently we only set something to be attested after we received the 2nd attestations.

I rewrote the logic to make it more clear that we now set sth to be attested only the first time we receive attestations

jonastheis: Ok sounds good!

pkg/protocol/engine/blockdag/inmemoryblockdag/blockdag.go
```diff
@@ -179,6 +185,8 @@ func (b *BlockDAG) attach(data *model.Block) (block *blocks.Block, wasAttached b
 	}
 
 	if updated {
+		b.LogTrace("missing block attached", "block", block.ID())
```
piotrm50, line 188: Should we log it here or is it better to attach to the event and log upon the event like for the other events?

johanneskeim (approved):

jonastheis (approved):

pkg/protocol/attestations.go
```diff
@@ -78,8 +78,10 @@ func (a *Attestations) initRequester() (shutdown func()) {
 	unsubscribeFromTicker := lo.Batch(
 		a.protocol.Commitments.WithElements(func(commitment *Commitment) (shutdown func()) {
 			return commitment.RequestAttestations.WithNonEmptyValue(func(_ bool) (teardown func()) {
-				if commitment.CumulativeWeight() == 0 {
-					commitment.IsAttested.Set(true)
+				if commitment.CumulativeWeight.Get() == 0 {
+					a.workerPool.Submit(func() {
```
piotrm50, line 83: Add a comment

pkg/protocol/commitment_verifier.go
```diff
@@ -128,7 +128,7 @@ func (c *CommitmentVerifier) verifyCommitment(commitment *Commitment, attestatio
 	//    than it actually is. Nodes might consider to switch to this chain, even though it is invalid which will be discovered
 	//    before the candidate chain/engine is activated (it will never get heavier than the current chain).
 	if seatCount > commitment.Weight.Get() {
-		return nil, 0, ierrors.Errorf("invalid cumulative weight for commitment %s: expected %d, got %d", commitment.ID(), commitment.CumulativeWeight(), seatCount)
+		return nil, 0, ierrors.Errorf("invalid cumulative weight for commitment %s: expected %d, got %d", commitment.ID(), commitment.CumulativeWeight.Get(), seatCount)
```
piotrm50: Change error message and return the same value as is checked.

pkg/protocol/commitments.go
```diff
@@ -327,8 +327,16 @@ func (c *Commitments) processRequest(commitmentID iotago.CommitmentID, from peer
 // processResponse processes the given commitment response.
 func (c *Commitments) processResponse(commitment *model.Commitment, from peer.ID) {
 	c.workerPool.Submit(func() {
+		// make sure the main engine is available to process the response
+		mainEngine := c.protocol.Engines.Main.Get()
+		if mainEngine == nil {
+			c.LogDebug("main engine unavailable for response", "commitment", commitment.ID(), "fromPeer", from)
```
hmoog:
```suggestion
			c.LogError("main engine unavailable for response", "commitment", commitment.ID(), "fromPeer", from)
```

pkg/protocol/engine/blockdag/inmemoryblockdag/blockdag.go
hmoog: We should always log "in place" and maybe create a separate issue to move over the remaining logs to their corresponding container.

piotrm50 (approved):

PR #637, opened 2023-12-20
Forward blocks with semantically invalid txs

### LINKED ISSUE #580: Semantic Invalid TX block should be booked and forwarded
Currently, a **semantic** invalid block will not be booked, and hence it will not be forwarded and the block will never be accepted, so the issuer of the block will not pay any Mana. 

This should be changed to not book the transaction but book the block and forward it as normal.

A block containing an invalid transaction should be picked up as a Weak parent in tip selection as part of standard tip selection procedure.

PhilippGackstatter:
Forward blocks with semantically invalid txs.

This is fixed by also setting blocks up (`b.setupBlock`) when the transaction is invalid or `OnSignaturesInvalid`. The latter is necessary such that, for instance, TXs whose reward inputs point to incorrect output types are also forwarded, i.e. checks that are done in `ValidateSignatures`.


pkg/protocol/engine/booker/inmemorybooker/booker.go
```diff
@@ -113,6 +113,16 @@ func (b *Booker) Queue(block *blocks.Block) error {
...
+
+		transactionMetadata.OnInvalid(func(err error) {
+			// TODO: Do we need to set spender IDs?
+			b.setupBlock(block)
+		})
+	})
+
+	signedTransactionMetadata.OnSignaturesInvalid(func(err error) {
+		// TODO: Do we need to set spender IDs?
+		b.setupBlock(block)
```
cyberphysic4l: I think this is correct, and we do not need to set spenderIDs because the transactions is not booked.

pkg/tests/reward_test.go
```diff
@@ -246,9 +246,9 @@ func Test_RewardInputCannotPointToNFTOutput(t *testing.T) {
 
 	// TODO: Assertions do not pass for node2 because the block does not get forwarded from node1.
 	// node2 should be added in the assertion when issue iotaledger/iota-core#580 is fixed.
```
cyberphysic4l: Great that this works as expected. We can remove the TODO here of course.

pkg/tests/booker_test.go
```diff
@@ -830,3 +831,31 @@ func Test_RootBlockShallowLike(t *testing.T) {
...
+			// Set an issuer ID that is not unlocked in the TX which will cause the TX to be invalid.
+			nftBuilder.ImmutableIssuer(&iotago.Ed25519Address{})
+		},
+	)
+	ts.IssueBasicBlockWithOptions("block1", ts.DefaultWallet(), tx1)
+
+	ts.Wait(node1, node2)
+
+	ts.AssertTransactionsExist([]*iotago.Transaction{tx1.Transaction}, true, node1, node2)
+	ts.AssertTransactionFailure(lo.PanicOnErr(tx1.ID()), iotago.ErrIssuerFeatureNotUnlocked, node1, node2)
```
jonastheis: Would be good to assert here that the block becomes accepted/confirmed. Optional: if you continue issuing until the slot is committed you could check that the tx is not included in the commitment but the block is.

piotrm50 (approved):

pkg/tests/booker_test.go
```diff
@@ -846,16 +855,29 @@ func Test_BlockWithInvalidTransactionGetsBooked(t *testing.T) {
...
 		func(nftBuilder *builder.NFTOutputBuilder) {
 			// Set an issuer ID that is not unlocked in the TX which will cause the TX to be invalid.
 			nftBuilder.ImmutableIssuer(&iotago.Ed25519Address{})
 		},
 	)
-	ts.IssueBasicBlockWithOptions("block1", ts.DefaultWallet(), tx1)
+	block1 := ts.IssueBasicBlockWithOptions("block1", ts.DefaultWallet(), tx1)
 
-	ts.Wait(node1, node2)
+	vblock1 := ts.IssueValidationBlockWithHeaderOptions("vblock1", ts.DefaultWallet().Node, mock.WithWeakParents(block1.ID()), mock.WithStrongParents(tip))
```
jonastheis:
```suggestion
	vblock1 := ts.IssueValidationBlockWithHeaderOptions("vblock1", ts.DefaultWallet().Node, mock.WithWeakParents(block1.ID()), mock.WithStrongParents(ts.Block("Genesis").ID()))
```

pkg/tests/booker_test.go
```diff
@@ -846,16 +855,29 @@ func Test_BlockWithInvalidTransactionGetsBooked(t *testing.T) {
...
+	vblock2 := ts.IssueValidationBlockWithHeaderOptions("vblock2", ts.DefaultWallet().Node, mock.WithStrongParents(vblock1.ID()))
+	vblock3 := ts.IssueValidationBlockWithHeaderOptions("vblock3", ts.DefaultWallet().Node, mock.WithStrongParents(vblock2.ID()))
+
+	ts.AssertBlocksInCacheAccepted(ts.Blocks("block1"), true, ts.Nodes()...)
+	ts.AssertBlocksInCacheConfirmed(ts.Blocks("block1"), true, ts.Nodes()...)
 
 	ts.AssertTransactionsExist([]*iotago.Transaction{tx1.Transaction}, true, node1, node2)
 	ts.AssertTransactionFailure(lo.PanicOnErr(tx1.ID()), iotago.ErrIssuerFeatureNotUnlocked, node1, node2)
```
jonastheis:
```suggestion
	ts.AssertTransactionsExist([]*iotago.Transaction{tx1.Transaction}, true, ts.Nodes()...)
	ts.AssertTransactionFailure(lo.PanicOnErr(tx1.ID()), iotago.ErrIssuerFeatureNotUnlocked, ts.Nodes()...)
```

jonastheis (approved):
Looks good! I just marked some minor simplifications for the test

pkg/protocol/engine/mempool/v1/transaction_metadata.go
```diff
@@ -300,9 +300,12 @@ func (t *TransactionMetadata) setup() (self *TransactionMetadata) {
 		}
 	})
 
-	t.OnEarliestIncludedAttachmentUpdated(func(previousIndex iotago.BlockID, newIndex iotago.BlockID) {
-		if isIncluded, wasIncluded := newIndex.Slot() != 0, previousIndex.Slot() != 0; isIncluded != wasIncluded {
-			t.accepted.Set(isIncluded && t.AllInputsAccepted() && t.IsConflictAccepted())
+	t.OnEarliestIncludedAttachmentUpdated(func(previousBlockID iotago.BlockID, newBlockID iotago.BlockID) {
+		if previousBlockID.Empty() && !newBlockID.Empty() {
+			if t.invalid.Get() != nil || !t.IsConflictAccepted() || !t.AllInputsAccepted() {
```
piotrm50, line 305: I think `t.invalid.Get() != nil` should also be checked in other cases: `input.OnAccepted(func() {` (line 256) and `if t.AllInputsAccepted() && t.EarliestIncludedAttachment().Slot() != 0 {` (line 243)

piotrm50 (changes requested):

piotrm50 (approved):

Issue #617, opened 2023-12-07
Make `optsActivityWindow` a protocol parameter


piotrm50:
The minimum sane value should be MinCommittableAge. It's currently a parameter in the SeatManager implementation.

alexsporn:
This is a dup of #519

PR #640, opened 2023-12-21
Activity Window from Protocol Parameters

### LINKED ISSUE #519: Make activity window a protocol parameter
The `ActivityTracker` was recently changed to work based on protocol time (solid blocks seen, and filtered) instead of real time.

```go
e.Events.CommitmentFilter.BlockAllowed.Hook(func(block *blocks.Block) {
	// Only track identities that are part of the committee.
	committee, exists := s.CommitteeInSlot(block.ID().Slot())
	if !exists {
		panic(ierrors.Errorf("committee not selected for slot %d, but received block in that slot", block.ID().Slot()))
	}

	seat, exists := committee.GetSeat(block.ProtocolBlock().Header.IssuerID)
	if exists {
		s.activityTracker.MarkSeatActive(seat, block.ProtocolBlock().Header.IssuerID, block.IssuingTime())
	}

	s.events.BlockProcessed.Trigger(block)
})
```

Therefore, having it as a local node configuration does not make much sense as different activity windows would lead to a different perception of online committee, thus leading to different commitments. Instead, it should be part of the protocol parameters.

cyberphysic4l:
The activity window can no longer be a locally defined parameter as this would lead to different perceptions of the online committee. Instead, this PR uses an existing protocol parameter, MinCommittableAge, as the activity window. The reason for this choice is that we have established that activity window must be less than MaxCommittableAge to prevent forcing commits even when there are online validators, because the activity window does not adapt fast enough. It is best to have it as small as possible to prevent this occuring, and the smallest sane value is MinCommittableAge according to https://github.com/iotaledger/iota-core/issues/617 although I am not sure why exactly we have this constraint @piotrm50 .

closes #519

### DIFF
--- a/pkg/protocol/sybilprotection/seatmanager/seatmanager.go
+++ b/pkg/protocol/sybilprotection/seatmanager/seatmanager.go
@@ -15,7 +15,7 @@ type SeatManager interface {
 	// RotateCommittee rotates the committee evaluating the given set of candidates to produce the new committee.
 	RotateCommittee(epoch iotago.EpochIndex, candidates accounts.AccountsData) (*account.SeatedAccounts, error)
 
-	// ReuseCommittee reuses the committee from a previous epoch epochs.
+	// ReuseCommittee reuses the committee from a previous epoch.
 	ReuseCommittee(epoch iotago.EpochIndex, committee *account.Accounts) error
 
 	// InitializeCommittee initializes the committee for the current slot by marking whole or a subset of the committee as active.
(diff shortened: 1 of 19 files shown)


piotrm50:
Actually the minimum value would be the maximum network delay, but that value is unknown. The assumption is that network delay is smaller than `LivenessThreshold` (TSC) - so actually now that I think about it this is the lowest value. If we go below the network delay then some nodes might diverge and consider different committee members online.

Using MinCommittableAge is safe and in the worst case will slightly delay committment creation, but I think that's fine.

pkg/protocol/sybilprotection/seatmanager/seatmanager.go
```diff
@@ -15,9 +15,8 @@ type SeatManager interface {
 	// RotateCommittee rotates the committee evaluating the given set of candidates to produce the new committee.
 	RotateCommittee(epoch iotago.EpochIndex, candidates accounts.AccountsData) (*account.SeatedAccounts, error)
 
-	// SetCommittee sets the committee for a given slot.
-	// This is used when re-using the same committee for consecutive epochs.
-	SetCommittee(epoch iotago.EpochIndex, committee *account.Accounts) error
+	// ReuseCommittee reuses the committee from a previous epoch epochs.
```
piotrm50:
```suggestion
	// ReuseCommittee reuses the committee from a previous epoch.
```

piotrm50 (approved):

Issue #629, opened 2023-12-12
Add protocol parameter which exposes the reward retention period


PhilippGackstatter:
Rewards are kept for a limited amount of epochs. To make it transparent to API consumers how long that is, we should add a protocol parameter which exposes the `RewardsRetentionPeriod` (name proposal).

PhilippGackstatter:
This protocol parameter does not necessarily have to be the one used in pruning, but at the least used in calculating the rewards. That is, only rewards up until the retention period should be retrieved. Whether more are stored that could be pruned is not as important.

PR #646, opened 2024-01-09
Use Reward Retention Period as pruning delay

### LINKED ISSUE #638: Do not allow users to prune pool rewards and stats KV stores
We expose this `management/v1/database/prune` API and that ends up calling `Prunable.Prune`, which prunes the `poolRewards` and `poolStats` KV stores. It is wrong to allow users to prune those. If one node has rewards stored for the last 365 epochs (as is the default) and another node's admin pruned to only retain the last 100 epochs, this would result in nodes calculating different mana rewards if someone is validating/delegating since before those 100 epochs. So those stores shouldn't be prunable by the user. It should be pruned only by the protocol in a deterministic way, so that nodes agree exactly on which rewards are still retained and which are pruned.

It would make sense to tackle this in one go with #629.

Edit: It might be the case that the implementation does not allow users to prune data before a minimum pruning delay. We have to check if that's the case.

PhilippGackstatter:
Fixes iotaledger/iota-core#638 and iotaledger/iota-core#629.

- The hardcoded pruning delay is replaced by a function which captures an API provider to determine the retention period for the to-be-pruned epoch.
- The rewards calculation will now only fetch epochs which are retained according to the new protocol parameter. Without this, a node that has pruning disabled would retain rewards indefinitely and such a node would under certain circumstances disagree on the reward amount with a node that has automatic pruning enabled.
    - As long as we don't change the retention period parameter across protocol upgrades, this is fine and we can keep the user-managed prune action through the management API. However, if we were to increase the retention period, whether a node has automatic pruning or user-managed pruning would matter.
        - At some new epoch X, where the increased retention period becomes active through a new protocol version, an automatically-pruned node would have pruned older rewards already, while a user-managed node would potentially still have them and hence, nodes would disagree on the calculated rewards. At this point we would have to disable user-managed pruning to make all nodes agree. This can be left as a future problem or can be fixed by handing responsibility of the reward storage completely to the protocol.
- It's not entirely clear to me whether the approach in this PR would work when we would increase or decrease the retention period in a protocol upgrade. I'm not sure if we should take care of this now or if and when we want to upgrade this particular parameter.

### DIFF
--- a/pkg/protocol/sybilprotection/sybilprotection.go
+++ b/pkg/protocol/sybilprotection/sybilprotection.go
@@ -21,13 +21,13 @@ type SybilProtection interface {
 	// The first epoch in which rewards existed is returned (firstRewardEpoch).
 	// Since the validator may still be active and the EndEpoch might be in the future, the epoch until which rewards were calculated is returned in addition to the first epoch in which rewards existed (lastRewardEpoch).
 	// The rewards are decayed until claimingEpoch, which should be set to the epoch in which the rewards would be claimed.
-	ValidatorReward(validatorID iotago.AccountID, stakingFeature *iotago.StakingFeature, claimingEpoch iotago.EpochIndex, retentionPeriod iotago.EpochIndex) (validatorReward iotago.Mana, firstRewardEpoch iotago.EpochIndex, lastRewardEpoch iotago.EpochIndex, err error)
+	ValidatorReward(validatorID iotago.AccountID, stakingFeature *iotago.StakingFeature, claimingEpoch iotago.EpochIndex) (validatorReward iotago.Mana, firstRewardEpoch iotago.EpochIndex, lastRewardEpoch iotago.EpochIndex, err error)
 	// DelegatorReward returns the amount of mana that a delegator has earned in a given epoch range.
 	//
 	// The first epoch in which rewards existed is returned (firstRewardEpoch).
 	// Since the Delegation Output's EndEpoch might be unset due to an ongoing delegation, the epoch until which rewards were calculated is also returned (lastRewardEpoch).
 	// The rewards are decayed until claimingEpoch, which should be set to the epoch in which the rewards would be claimed.
-	DelegatorReward(validatorID iotago.AccountID, delegatedAmount iotago.BaseToken, epochStart iotago.EpochIndex, epochEnd iotago.EpochIndex, claimingEpoch iotago.EpochIndex, retentionPeriod iotago.EpochIndex) (delegatorReward iotago.Mana, firstRewardEpoch iotago.EpochIndex, lastRewardEpoch iotago.EpochIndex, err error)
+	DelegatorReward(validatorID iotago.AccountID, delegatedAmount iotago.BaseToken, epochStart iotago.EpochIndex, epochEnd iotago.EpochIndex, claimingEpoch iotago.EpochIndex) (delegatorReward iotago.Mana, firstRewardEpoch iotago.EpochIndex, lastRewardEpoch iotago.EpochIndex, err error)
 	// PoolRewardsForAccount returns the pool rewards of the account identified by accountID in the last committed epoch.
 	PoolRewardsForAccount(accountID iotago.AccountID) (
 		poolRewardsForAccount iotago.Mana, exists bool, err error)
--- a/pkg/protocol/sybilprotection/sybilprotection.go
+++ b/pkg/protocol/sybilprotection/sybilprotection.go
@@ -21,7 +21,7 @@ type SybilProtection interface {
 	// The first epoch in which rewards existed is returned (firstRewardEpoch).
 	// Since the validator may still be active and the EndEpoch might be in the future, the epoch until which rewards were calculated is returned in addition to the first epoch in which rewards existed (lastRewardEpoch).
 	// The rewards are decayed until claimingEpoch, which should be set to the epoch in which the rewards would be claimed.
-	ValidatorReward(validatorID iotago.AccountID, stakingFeature *iotago.StakingFeature, claimingEpoch iotago.EpochIndex) (validatorReward iotago.Mana, firstRewardEpoch iotago.EpochIndex, lastRewardEpoch iotago.EpochIndex, err error)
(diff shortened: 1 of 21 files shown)


pkg/storage/prunable/prunable_epoch.go
```diff
@@ -17,9 +17,6 @@ const (
 
 const (
 	pruningDelayDecidedUpgradeSignals = 7
```
jonastheis, line 19: We should get rid of this hardcoded parameter as well. Probably in a separate PR. With the changes made in this PR it should be straightforward.

pkg/storage/prunable/epochstore/epoch_kv.go
```diff
@@ -75,13 +75,15 @@ func (e *EpochKVStore) DeleteEpoch(epoch iotago.EpochIndex) error {
 }
 
 func (e *EpochKVStore) Prune(epoch iotago.EpochIndex, defaultPruningDelay iotago.EpochIndex) error {
+	minPruningDelay := e.prunginDelayFunc(epoch)
+
 	// The epoch we're trying to prune already takes into account the defaultPruningDelay.
 	// Therefore, we don't need to do anything if it is greater equal e.pruningDelay and take the difference otherwise.
```
jonastheis: adjust comment

pkg/storage/prunable/epochstore/store.go
```diff
@@ -122,13 +122,15 @@ func (s *Store[V]) DeleteEpoch(epoch iotago.EpochIndex) error {
 }
 
 func (s *Store[V]) Prune(epoch iotago.EpochIndex, defaultPruningDelay iotago.EpochIndex) error {
+	minPruningDelay := s.prunginDelayFunc(epoch)
+
 	// The epoch we're trying to prune already takes into account the defaultPruningDelay.
 	// Therefore, we don't need to do anything if it is greater equal s.pruningDelay and take the difference otherwise.
```
jonastheis: adjust comment

pkg/protocol/sybilprotection/sybilprotection.go
```diff
@@ -21,13 +21,13 @@ type SybilProtection interface {
 	// The first epoch in which rewards existed is returned (firstRewardEpoch).
 	// Since the validator may still be active and the EndEpoch might be in the future, the epoch until which rewards were calculated is returned in addition to the first epoch in which rewards existed (lastRewardEpoch).
 	// The rewards are decayed until claimingEpoch, which should be set to the epoch in which the rewards would be claimed.
-	ValidatorReward(validatorID iotago.AccountID, stakingFeature *iotago.StakingFeature, claimingEpoch iotago.EpochIndex) (validatorReward iotago.Mana, firstRewardEpoch iotago.EpochIndex, lastRewardEpoch iotago.EpochIndex, err error)
+	ValidatorReward(validatorID iotago.AccountID, stakingFeature *iotago.StakingFeature, claimingEpoch iotago.EpochIndex, retentionPeriod iotago.EpochIndex) (validatorReward iotago.Mana, firstRewardEpoch iotago.EpochIndex, lastRewardEpoch iotago.EpochIndex, err error)
```
jonastheis: Would be good to explain the behavior/effects of `retentionPeriod` in the comment. 

Another thought: why is this part of the interface and putting the "burden" on the caller. This increases the surface for errors when calling this method as the caller needs to specify the retention period. However, I think the Sybilprotection implementation should be aware of the protocol parameters already. Is there a way we can abstract it away?

pkg/protocol/sybilprotection/sybilprotection.go
```diff
@@ -21,13 +21,13 @@ type SybilProtection interface {
...
 	// DelegatorReward returns the amount of mana that a delegator has earned in a given epoch range.
 	//
 	// The first epoch in which rewards existed is returned (firstRewardEpoch).
 	// Since the Delegation Output's EndEpoch might be unset due to an ongoing delegation, the epoch until which rewards were calculated is also returned (lastRewardEpoch).
 	// The rewards are decayed until claimingEpoch, which should be set to the epoch in which the rewards would be claimed.
-	DelegatorReward(validatorID iotago.AccountID, delegatedAmount iotago.BaseToken, epochStart iotago.EpochIndex, epochEnd iotago.EpochIndex, claimingEpoch iotago.EpochIndex) (delegatorReward iotago.Mana, firstRewardEpoch iotago.EpochIndex, lastRewardEpoch iotago.EpochIndex, err error)
+	DelegatorReward(validatorID iotago.AccountID, delegatedAmount iotago.BaseToken, epochStart iotago.EpochIndex, epochEnd iotago.EpochIndex, claimingEpoch iotago.EpochIndex, retentionPeriod iotago.EpochIndex) (delegatorReward iotago.Mana, firstRewardEpoch iotago.EpochIndex, lastRewardEpoch iotago.EpochIndex, err error)
```
jonastheis: Would be good to explain the behavior/effects of `retentionPeriod` in the comment. 

Another thought: why is this part of the interface and putting the "burden" on the caller. This increases the surface for errors when calling this method as the caller needs to specify the retention period. However, I think the Sybilprotection implementation should be aware of the protocol parameters already. Is there a way we can abstract it away?

pkg/protocol/sybilprotection/sybilprotectionv1/performance/rewards.go
```diff
@@ -42,6 +42,13 @@ func (t *Tracker) ValidatorReward(validatorID iotago.AccountID, stakingFeature *
 
 	decayEndEpoch := t.decayEndEpoch(claimingEpoch, lastRewardEpoch)
 
+	// Determine the earliest epoch for which rewards are retained and only attempt to fetch those.
+	var earliestRetainedRewardEpoch iotago.EpochIndex
+	if retentionPeriod < claimingEpoch {
+		earliestRetainedRewardEpoch = claimingEpoch - retentionPeriod
+	}
+	firstRewardEpoch = lo.Max(earliestRetainedRewardEpoch, firstRewardEpoch)
```
jonastheis: Maybe deduplicate this code and create a little convenience function?

jonastheis:
> It's not entirely clear to me whether the approach in this PR would work when we would increase or decrease the retention period in a protocol upgrade. I'm not sure if we should take care of this now or if and when we want to upgrade this particular parameter.

I think decreasing should be ok as all nodes have the required information. Increasing would be tricky but I think we should worry about that if and when we want to do that.

jonastheis (approved):

PR #651, opened 2024-01-15
Do not propagate confirmation ratifiers over epoch boundaries

### LINKED ISSUE #650: Only count confirmation votes if the votes are from the same epoch as the block they are voting for.
To avoid any issues with changing committees and counting previous and newer votes at the epoch boundary, we should add a condition to not count votes if they are not from the same epoch the block was issued in. This might delay confirmation for the last 2 slots of an epoch but will make sure that confirmation is safe in cases where the minority of voters got different seats for the next epoch.

jonastheis:
Fixes #650


alexsporn (approved):

alexsporn (approved):
Nice test 👍

piotrm50 (approved):

PR #652, opened 2024-01-15
Add Blockretainer

### LINKED ISSUE #588: Retainer needs to update the transaction metadata for all attachments
The transaction metadata should always be the same, no matter which block you are asking for, so we need to update the transaction metadata for all attachments.
Should we change the retainer storage for that, so we don't have to write the block metadata again?

Is "updating all attachments" actually an amplification attack vector?

### LINKED ISSUE #701: Cleanup internal node events and hook correctly with API
The `inx-faucet` sometimes sends several payouts, because it adds the user requests back to the queue in case of an error.

```
docker-network-inx-faucet-1        | 2024-01-25T18:23:21Z	WARNING	Faucet	checkPendingTransactionState failed: failed to fetch metadata of the pending transaction, blockID: BlockID(0x3792a2572f63c9ddd330aed36303026ea062531d82dba7a512d24db010a0239374000000:116), txID: TransactionID(0xfa4955eeaa7824ef62849f8c61954743340deb3affe94c5ac43ef0dd80b886136a000000:106) 
```

It seems to me that the block is not known to the node, even though it accepted the block and returned "OK" via inx.
We should wait in the node before returning, until it is 100% certain that the block will be sent to the network.

- maybe we can return earlier in the post block endpoint in case of error
- we also need to make sure that the retainer saw the block/tx before we return from the post block endpoint in the other cases

### LINKED ISSUE #756: Create BlockRetainer component
The BlockRetainer is a compinent which stores metadata about the blocks and is based on the current design with the prunable storage.
It needs to handle Pruning and Chain Switching (Reset method) correctly. We should also add the optional error message mentioned in https://github.com/iotaledger/iota-core/issues/697 while we are at it.

We should do test driven development for these components.
The tests should include at least the following scenarios:

   - First set object as pending
   - Overwrite with new state, check if potential former fields are correctly set/unset.
   - Check if values can be retrieved after setting them.
   - Reset at forking point and check if state is correctly set (newer objects deleted, old still exist)
   - Check if pruning works correctly

Follow up to #588

daria305:
Fixes issue: #588 

The retainer was split into a `BlockRetainer` and a `TransactionRetainer` component.
See #769 as well.


pkg/retainer/retainer/retainer.go
```diff
@@ -84,51 +85,37 @@ func NewProvider() module.Provider[*engine.Engine, retainer.Retainer] {
...
 		e.Events.Scheduler.BlockDropped.Hook(func(b *blocks.Block, err error) {
-			r.RetainBlockFailure(b.ID(), api.BlockFailureDroppedDueToCongestion)
+			r.RetainBlockFailure(b, api.BlockFailureDroppedDueToCongestion)
 		})
 
 		e.Initialized.OnTrigger(func() {
 			e.Ledger.MemPool().OnSignedTransactionAttached(func(signedTransactionMetadata mempool.SignedTransactionMetadata) {
-				attachment := signedTransactionMetadata.Attachments()[0]
-
+				//attachment := signedTransactionMetadata.Attachments()[0]
```
jkrvivian: Remove this line.

pkg/retainer/retainer.go
```diff
@@ -13,8 +15,9 @@ type Retainer interface {
 	RegisteredValidatorsCache(uint32) ([]*api.ValidatorResponse, bool)
 	RetainRegisteredValidatorsCache(uint32, []*api.ValidatorResponse)
 
-	RetainBlockFailure(iotago.BlockID, api.BlockFailureReason)
-	RetainTransactionFailure(iotago.BlockID, error)
+	RetainBlockFailure(*blocks.Block, api.BlockFailureReason)
+	RetainModelBlockFailure(*model.Block, api.BlockFailureReason)
```
jkrvivian: We can get `model.Block` from `blocks.Block`, so only 1 RetainBlockFailure is needed, with the signature: `RetainBlockFailure(*model.Block, api.BlockFailureReason)`

pkg/retainer/retainer/retainer.go
```diff
@@ -151,7 +138,7 @@ func (r *Retainer) Shutdown() {
 }
 
 func (r *Retainer) BlockMetadata(blockID iotago.BlockID) (*retainer.BlockMetadata, error) {
-	blockStatus, blockFailureReason := r.blockStatus(blockID)
+	blockStatus, blockFailureReason, _ := r.blockStatus(blockID)
```
jkrvivian: Get the transactionID here and make `transactionStatus` to take a transactionID instead of BlockID, that way we don't need to `getBlockData` for 2 times for the transactionID.

pkg/retainer/retainer/store.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	if currentSlot < latestAttachmentSlot {
+		moved, err := m.moveTransactionData(txID, currentSlot, latestAttachmentSlot)
+		if moved {
+			m.transactionLatestAttachmentSlot.Set(txID, latestAttachmentSlot)
+		}
+		if err != nil {
+			return err
+		}
```
jkrvivian: Check the error first.

pkg/retainer/retainer/store.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/iotaledger/iota-core/pkg/storage/prunable/slotstore"
+	iotago "github.com/iotaledger/iota.go/v4"
+	"github.com/iotaledger/iota.go/v4/api"
+)
+
+type metadataStore struct {
+	store StoreFunc
+
+	// transaction metadata is kept in one place and stored under the latest attachment slot to align block and transaction pruning
+	transactionLatestAttachmentSlot *shrinkingmap.ShrinkingMap[iotago.TransactionID, iotago.SlotIndex]
```
jkrvivian: Does the shrinkingMap also prune the pruned slots?

pkg/storage/prunable/slotstore/retainer.go
```diff
@@ -41,59 +46,69 @@ func BlockRetainerDataFromBytes(bytes []byte) (*BlockRetainerData, int, error) {
...
 type TransactionRetainerData struct {
-	TransactionID iotago.TransactionID
 	State         api.TransactionState
 	FailureReason api.TransactionFailureReason
+	// needed for a finalization status evaluation
+	ConfirmedAttachmentSlot iotago.SlotIndex
 }
 
 func (t *TransactionRetainerData) Bytes() ([]byte, error) {
 	byteBuffer := stream.NewByteBuffer(2)
```
jkrvivian: The size here doesn't seem correct.

jkrvivian (changes requested):

pkg/protocol/engine/ledger/ledger/ledger.go
```diff
@@ -131,7 +131,13 @@ func (l *Ledger) AttachTransaction(block *blocks.Block) (attachedTransaction mem
 	if signedTransaction, hasTransaction := block.SignedTransaction(); hasTransaction {
 		signedTransactionMetadata, err := l.memPool.AttachSignedTransaction(signedTransaction, signedTransaction.Transaction, block.ID())
 		if err != nil {
-			l.retainTransactionFailure(block.ID(), err)
+			transactionID, txIDErr := signedTransaction.Transaction.ID()
+			if txIDErr != nil {
+				l.errorHandler(err)
```
PhilippGackstatter:
```suggestion
				l.errorHandler(txIDErr)
```

Shouldn't we pass the txIDErr? Or can you document in a comment why you pass `err` instead of `txIDErr`?

pkg/retainer/retainer/retainer.go
```diff
@@ -98,51 +99,36 @@ func NewProvider() module.Provider[*engine.Engine, retainer.Retainer] {
...
-					r.RetainTransactionFailure(attachment, err)
+					r.RetainTransactionFailure(txID, err)
 				})
 
 				signedTransactionMetadata.OnSignaturesValid(func() {
 					transactionMetadata := signedTransactionMetadata.TransactionMetadata()
 
 					// transaction is not included yet, thus EarliestIncludedAttachment is not set.
-					if err := r.onTransactionAttached(attachment); err != nil {
+					if err := r.onTransactionAttached(txID); err != nil {
```
PhilippGackstatter: Just double-checking: Is this comment still relevant after your change?

pkg/retainer/retainer/retainer.go
```diff
@@ -229,121 +208,85 @@ func (r *Retainer) RetainRegisteredValidatorsCache(index uint32, resp []*api.Val
...
 	if err != nil {
-		r.errorHandler(ierrors.Wrapf(err, "could not get retainer store for slot %d", blockID.Slot()))
-		return api.BlockStateUnknown, api.BlockFailureNone
+		r.errorHandler(ierrors.Wrap(err, "failed to get txID from attached block"))
 	}
 
-	blockData, exists := store.GetBlock(blockID)
-	if !exists {
-		return api.BlockStateUnknown, api.BlockFailureNone
+	return txID
```
PhilippGackstatter: If there is an error we should not touch `txID` imo, because we should not assume anything about its value (could be empty TX ID or nil). We should return an explicit value like `iotago.EmptyTransactionID`.
On the other hand, if we're sure the operation cannot ever fail, then we should use `lo.PanicOnErr` with a comment explaining why it's okay to use. The latter would be preferable here imo, since the computation should always succeed unless the serialization is broken in which case we would see the panic during tests.

pkg/retainer/retainer/store.go
```diff
@@ -0,0 +1,222 @@
...
+	return nil
+}
+
+func (m *metadataStore) moveTransactionMetadataToTheLatestAttachmentSlot(txID iotago.TransactionID, latestAttachmentSlot iotago.SlotIndex) error {
+	m.attachmentSlotMutex.Lock()
+	defer m.attachmentSlotMutex.Unlock()
+
+	if txID == iotago.EmptyTransactionID {
+		return nil
+	}
```
PhilippGackstatter: Move the lock/unlock after the empty tx ID check.

pkg/retainer/retainer/retainer.go
```diff
@@ -187,37 +173,30 @@ func (r *Retainer) BlockMetadata(blockID iotago.BlockID) (*retainer.BlockMetadat
...
-	store, storeErr := r.store(blockID.Slot())
-	if storeErr != nil {
-		r.errorHandler(ierrors.Wrapf(storeErr, "could not get retainer store for slot %d", blockID.Slot()))
-		return
-	}
-
-	if err := store.StoreTransactionFailure(blockID, determineTxFailureReason(err)); err != nil {
+// RetainTransactionFailure stores the transaction failure in the retainer.
+func (r *Retainer) RetainTransactionFailure(txID iotago.TransactionID, reason error) {
+	err := r.store.setTransactionFailure(txID, determineTxFailureReason(reason))
```
PhilippGackstatter: This now needs to be `api.DetermineTxFailureReason(reason)` since we moved that function to iota.go.

PhilippGackstatter (commented):
Some preliminary comments while I try to understand the rest.

pkg/retainer/retainer/retainer.go
daria305: looks like it is not

pkg/retainer/retainer/retainer.go
daria305: I have returned an `EmptyTransactionID` as in vm for example we return error, and just to avoid panic

pkg/storage/prunable/slotstore/retainer.go
```diff
@@ -41,59 +53,69 @@ func BlockRetainerDataFromBytes(bytes []byte) (*BlockRetainerData, int, error) {
...
+
 	return t, byteReader.BytesRead(), nil
 }
 
 type Retainer struct {
 	slot       iotago.SlotIndex
 	blockStore *kvstore.TypedStore[iotago.BlockID, *BlockRetainerData]
 	// we store transaction metadata per blockID as in API requests we always request by blockID
-	transactionStore *kvstore.TypedStore[iotago.BlockID, *TransactionRetainerData]
+	transactionStore *kvstore.TypedStore[iotago.TransactionID, *TransactionRetainerData]
```
PhilippGackstatter: Update the comment btw.

muXxer (approved):

PR #666, opened 2024-01-16
Re-elected committee member retains its SeatIndex

### LINKED ISSUE #648: Try to assign the same seats to committee members that get re-selected for the next epoch

piotrm50:


pkg/protocol/sybilprotection/seatmanager/topstakers/topstakers.go
```diff
@@ -86,21 +86,45 @@ func (s *SeatManager) RotateCommittee(epoch iotago.EpochIndex, candidates accoun
...
+
+	prevCommittee, err := s.committeeStore.Load(epoch - 1)
 	if err != nil {
-		return nil, ierrors.Wrapf(err, "error while getting committeeAccounts for newly selected committee for epoch %d", epoch)
+		return nil, err
+	}
+
+	// If there is no previous committee, then we can assign seats by default.
+	if prevCommittee == nil {
+		newCommittee := committeeAccounts.SeatedAccounts(committeeAccounts.IDs()...)
```
alexsporn: this is kind of the same as `if epoch == 0`

muXxer: grafik
just saying..... :smiling_imp:

hmoog (approved):
LGTM

pkg/testsuite/sybilprotection.go
```diff
@@ -41,6 +41,39 @@ func (t *TestSuite) AssertSybilProtectionCommittee(epoch iotago.EpochIndex, expe
...
+			}
+
+			committeeInPrevEpochAccounts, err := committeeInPrevEpoch.Accounts()
+			if err != nil {
+				return ierrors.Errorf("AssertReelectedCommitteeSeatIndices: %s: failed to get accounts in committee in previous epoch %d: %w", node.Name, prevEpoch, err)
+			}
+
+			committeeInPrevEpochAccounts.ForEach(func(id iotago.AccountID, _ *account.Pool) bool {
+				if seatIndex, memberExists := committeeInNewEpoch.GetSeat(id); memberExists {
+					require.Equalf(t.Testing, seatIndex, lo.Return1(committeeInPrevEpoch.GetSeat(id)), "account %s must have the same SeatIndex in the previous and new epoch committee", id)
```
jonastheis: This doesn't really work with the way we use `t.Eventually`. I think the correct way would be to assign an inner error and check for it after the loop. Or maybe you can get all the accounts as a slice?

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
```diff
@@ -226,17 +225,10 @@ func (o *SybilProtection) committeeRoot(targetCommitteeEpoch iotago.EpochIndex)
 		iotago.AccountIDFromBytes,
 	)
 
-	var innerErr error
-	committee.ForEach(func(accountID iotago.AccountID, _ *account.Pool) bool {
-		if err := committeeTree.Add(accountID); err != nil {
-			innerErr = ierrors.Wrapf(err, "failed to add account %s to committee tree", accountID)
-			return false
+	for _, accountID := range committee.IDs() {
+		if err = committeeTree.Add(accountID); err != nil {
```
jonastheis, line 229: Since we export seat numbers for a snapshot (via `o.performanceTracker.Export`) we should really make sure that they are the same across nodes as otherwise snapshots at the same epoch can't be directly compared (eg via hashes). Would it make sense to commit to the seat here in a `ads.Map`?

pkg/core/account/accounts_test.go
```diff
@@ -124,14 +119,78 @@ func TestAccounts(t *testing.T) {
...
+
+		return true
+	})
+
+	// Make sure that size of the committee and SeatIndex values are sane.
+	require.Equal(t, newCommittee.SeatCount(), committeeSize)
+
+	newCommitteeAccounts.ForEach(func(id iotago.AccountID, _ *account.Pool) bool {
+		require.Less(t, int(lo.Return1(newCommittee.GetSeat(id))), committeeSize)
+		require.GreaterOrEqual(t, int(lo.Return1(newCommittee.GetSeat(id))), 0)
```
jonastheis: What exactly are you checking here? Not easy to read

pkg/core/account/accounts.go
```diff
@@ -125,9 +118,54 @@ func (a *Accounts) ForEach(callback func(id iotago.AccountID, pool *Pool) bool)
...
+
+	// If previous committee exists the seats need to be assigned and re-elected committee members must retain their SeatIndex.
+	// Keep track of re-elected members' seat indices.
+	takenSeatIndices := ds.NewSet[SeatIndex]()
+
+	// Assign re-elected members' SeatIndex.
+	a.ForEach(func(id iotago.AccountID, _ *Pool) bool {
+		if seatIndex, seatExists := prevCommittee.GetSeat(id); seatExists {
+			newCommittee.Set(seatIndex, id)
+			takenSeatIndices.Add(seatIndex)
```
jonastheis, line 140: How exactly does this behave if `prevCommittee` is bigger than `newCommittee`? In the test the committees are equal size iiuc. Would be good to have a test case for that.

pkg/core/account/accounts.go
```diff
@@ -125,9 +118,54 @@ func (a *Accounts) ForEach(callback func(id iotago.AccountID, pool *Pool) bool)
...
+		}
+
+		return true
+	})
+
+	// Sort members lexicographically.
+	committeeAccountIDs := a.IDs()
+	sort.Slice(committeeAccountIDs, func(i int, j int) bool {
```
jonastheis, line 148: Is this guaranteed to **always** result in the same seats? eg nodes starting from genesis vs starting from snapshot

jonastheis (changes requested):

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
piotrm50: Damn, this is something I completely missed. There is actually no reason why nodes would be required to assign the same seats - this should be completely subjective. Maybe instead we should just export accounts to the snapshot so that it can be hashed and compared, and then nodes initialize their seats when loading the snapshot? Then each node could assign seats differently without causing much problems.

pkg/core/account/accounts.go
piotrm50: No, but this should be completely subjective. What is important that a node maintains the same SeatIndex if a committee member is reelected. We should take care of that snapshot issue you mentioned though.

pkg/core/account/accounts_test.go
piotrm50: Here I check that SeatIndices are within <0; committeeSize) range. Added a comment to explain.

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
jonastheis: Yep that sounds good. Just need to make sure to document this properly so that it is clear that `Export` and `SaveToDisk` (or whatever we'll call it) are different in this way.

pkg/core/account/accounts.go
piotrm50: I extended the test to cover going to a smaller and bigger committee.

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
piotrm50: I changed it and now we only export Accounts to the snapshot along with the "reused" flag.
When loading the snapshot, we don't maintain seats for past slots as I think this is only required for the current epochs to track seat activity correctly - or maybe I'm missing something? I provided extensive comments describing the reasoning.

jonastheis (approved):

PR #673, opened 2024-01-17
Feat: Add validator tip set

### LINKED ISSUE #618: Introduce a special tip pool to track ValidatorBlocks and their future cones.
Currently acceptance stops if tip pool size explodes, which is bad for the node. 
Introduce a tip pool that will keep track of latest validator blocks and their future cone for each validator. We're only interested in a single blocks from the future cone of validator block, so size of this tip pool will be limited and it will allow us to continue progressing acceptance time regardless of the tip pool size. This needs to be discussed in more detail to make sure that the game-theoretical properties of the tip selection are not changed significantly.

hmoog:


pkg/protocol/engine/tipmanager/v1/tipmanager.go
```diff
@@ -150,6 +199,40 @@ func (t *TipManager) setupBlockMetadata(tipMetadata *TipMetadata) {
 	t.blockAdded.Trigger(tipMetadata)
 }
 
+// trackLatestValidationBlock tracks the latest validator block and takes care of marking the corresponding TipMetadata.
+func (t *TipManager) trackLatestValidationBlock(tipMetadata *TipMetadata) (teardown func()) {
+	if _, isValidationBlock := tipMetadata.Block().ValidationBlock(); !isValidationBlock {
+		return
+	}
+
+	committee, exists := t.retrieveCommitteeInSlot(tipMetadata.Block().ID().Slot())
```
piotrm50, line 208: This is a hot path that reads data from the committee KVStore for each block - we should add a cache at some level. Maybe we should have a committee cache in SeatManager and clear it whenever SybilProtection.Reset() is called? Not necessarily as part of this PR, but then we should create an issue for that.

piotrm50 (approved):

PR #681, opened 2024-01-19
Fix race condition when stopping publishing candidate payloads.

piotrm50:
Fix race condition when stopping publishing candidate payloads at network startup.
Previously the network started and sometimes candidate payloads were issued before it was stopped in a test, causing committee checks to fail due to selecting different committee than expected.

### DIFF
--- a/tools/docker-network/tests/committeerotation_test.go
+++ b/tools/docker-network/tests/committeerotation_test.go
@@ -152,7 +152,7 @@ func Test_NoCandidacyPayload(t *testing.T) {
 	d.AddValidatorNode("V4", "docker-network-inx-validator-4-1", "http://localhost:8040", "rms1pr8cxs3dzu9xh4cduff4dd4cxdthpjkpwmz2244f75m0urslrsvtsshrrjw")
 	d.AddNode("node5", "docker-network-node-5-1", "http://localhost:8090")
 
-	err := d.Run()
+	err := d.Run(map[string]bool{"V1": false, "V2": false, "V3": false, "V4": false})
 	require.NoError(t, err)
 
 	err = d.WaitUntilSync()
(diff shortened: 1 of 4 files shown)


tools/docker-network/tests/committeerotation_test.go
```diff
@@ -152,7 +152,7 @@ func Test_NoCandidacyPayload(t *testing.T) {
 	d.AddValidatorNode("V4", "docker-network-inx-validator-4-1", "http://localhost:8040", "rms1pr8cxs3dzu9xh4cduff4dd4cxdthpjkpwmz2244f75m0urslrsvtsshrrjw")
 	d.AddNode("node5", "docker-network-node-5-1", "http://localhost:8090")
 
-	err := d.Run()
+	err := d.Run(map[string]bool{"V1": false, "V2": false, "V3": false, "V4": false})
```
jonastheis: Maybe this could be integrated into the `Node` struct? Then you can configure a validator to be issuing or not when adding the validator node. Otherwise it's quite complex to configure this correctly

tools/docker-network/tests/dockerframework.go
```diff
@@ -284,23 +297,36 @@ func (d *DockerTestFramework) AccountsFromNodes(nodes ...*Node) []string {
 	return accounts
 }
 
-func (d *DockerTestFramework) SetIssueCandidacyPayload(issueCandidacyPayloadMap map[string]bool) {
-	cmd := exec.Command("docker", "compose", "up", "-d")
+func (d *DockerTestFramework) StartIssueCandidacyPayload(nodeNames ...string) {
```
jonastheis: I think it's better design to pass a slice/variadic of nodes instead of the string/alias. That way you can also pass in a variable of a node and don't need to handle the string. When calling it with the aliases you just need to wrap it in `d.Nodes("V1", "V2", "V3")`.

tools/docker-network/tests/dockerframework.go
```diff
@@ -284,23 +297,36 @@ func (d *DockerTestFramework) AccountsFromNodes(nodes ...*Node) []string {
...
+			node.IssueCandidacyPayload = true
+		}
+	}
+
+	d.DockerComposeUp(true)
+}
 
-	for nodeName, issueCandidacyPayload := range issueCandidacyPayloadMap {
-		cmd.Env = append(cmd.Env, fmt.Sprintf("ISSUE_CANDIDACY_PAYLOAD_%s=%t", nodeName, issueCandidacyPayload))
+func (d *DockerTestFramework) StopIssueCandidacyPayload(nodeNames ...string) {
```
jonastheis: Same as above

jonastheis (approved):

tools/docker-network/tests/dockerframework.go
piotrm50: Done

PR #719, opened 2024-02-05
New WorkScoreParameters

cyberphysic4l:
This PR includes:
- benchmarking tool and regression analysis used to calculate new values for the WorkScoreParameters
- changes to the tests to work more robustly with any set of valid WorkScoreParameters

In summary, the regression analysis works by benchmarking how long different types of blocks take to be processed by a mock node in our existing test suite, from issuance to scheduling. This is used as an indicator of how computationally expensive the block is to process. By varying the number and types of inputs and outputs in different transactions, we end up with a data set which we then use to train a linear model by regression, i.e., we find the WorkScoreParameters that give the best fit to the data we generated. 

The regression analysis itself gives us the ratios of the parameters to one another, and these values can then be scaled as we like. In this case we decide the scale such that a "standard block" costs roughly 500,000 Mana (we will call this 0.5 Mana to users and the "code Mana" would be called microMana). The standard block we choose is a block containing a transaction that creates an account output with a block issuer feature from a basic output. 

Another thing that the regression analysis does not give us is the `DataByte` factor of the WorkScoreParameters because this is related to bandwidth resources which are orthogonal to computation resources. To decide a `DataByte` factor, we just need to specify what importance we place on bandwidth versus computation by specifying a ratio. For example, the chosen parameters use a ratio 0.5, which means that bandwidth is 0.5 times as important as computation. The result is that we choose our data byte factor such that the data byte part of a standard block's WorkScore is 0.5 times that of the other components of that same block's WorkScore.


pkg/testsuite/mock/wallet.go
```diff
@@ -121,26 +121,26 @@ func (w *Wallet) TransactionID(alias string) iotago.TransactionID {
...
-	address := w.keyManager.Address(iotago.AddressImplicitAccountCreation)
+func (w *Wallet) ImplicitAccountCreationAddress(index ...uint32) *iotago.ImplicitAccountCreationAddress {
+	address := w.keyManager.Address(iotago.AddressImplicitAccountCreation, index...)
 	//nolint:forcetypeassert
 	return address.(*iotago.ImplicitAccountCreationAddress)
 }
 
-func (w *Wallet) HasAddress(address iotago.Address) bool {
-	return address.Equal(w.Address()) || address.Equal(w.ImplicitAccountCreationAddress())
+func (w *Wallet) HasAddress(address iotago.Address, index uint32) bool {
```
muXxer: wouldn't it make sense to also use index as variadic parameter here?

pkg/testsuite/testsuite.go
```diff
@@ -99,7 +99,7 @@ func NewTestSuite(testingT *testing.T, opts ...options.Option[TestSuite]) *TestS
 		optsTick:    durationFromEnvOrDefault(2*time.Millisecond, "CI_UNIT_TESTS_TICK"),
 		optsLogger:  loggerFromEnvOrDefault("CI_UNIT_TESTS_NO_LOG", "CI_UNIT_TESTS_LOG_LEVEL"),
 	}, opts, func(t *TestSuite) {
-		fmt.Println("Setup TestSuite -", testingT.Name(), " @ ", time.Now())
```
muXxer, line 102: was this on purpose? maybe remove the whole line if we don't need it anymore.

pkg/testsuite/mock/node.go
```diff
@@ -99,10 +99,11 @@ func NewNode(t *testing.T, parentLogger log.Logger, net *Network, partition stri
 	} else {
 		validationBlockIssuer = nil
 	}
-
+	logger := parentLogger.NewChildLogger(name)
+	//logger.SetLogLevel(log.LevelTrace)
 	return &Node{
 		Testing: t,
-		logger:  parentLogger.NewChildLogger(name),
+		logger:  logger,
```
muXxer: I think those changes can be reverted?

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -664,15 +1010,19 @@ func (w *Wallet) createSignedTransactionWithOptions(transactionName string, opts
...
 
 	// register the outputs in the wallet
-	w.registerOutputs(transactionName, signedTransaction.Transaction)
+	w.registerOutputs(transactionName, signedTransaction.Transaction, addressIndexes...)
 
 	return signedTransaction
 }
 
-func (w *Wallet) registerOutputs(transactionName string, transaction *iotago.Transaction) {
+func (w *Wallet) registerOutputs(transactionName string, transaction *iotago.Transaction, indexes ...uint32) {
```
muXxer: I would call it `addressIndexes` here as well.

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -664,15 +1010,19 @@ func (w *Wallet) createSignedTransactionWithOptions(transactionName string, opts
...
+	if indexes == nil {
+		indexes = []uint32{0}
+	}
```
muXxer:
```suggestion
	if len(indexes) == 0 {
		indexes = []uint32{0}
	}
```

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -301,6 +380,169 @@ func (w *Wallet) TransitionImplicitAccountToAccountOutput(transactionName string
...
+			},
+			Features: iotago.BasicOutputFeatures{
+				nativeTokenFeature,
+			},
+		})
+	}
+
+	accountOutput, isAccount := outputStates[1].(*iotago.AccountOutput)
+	if !isAccount {
+		panic("output with alias accountName is not *iotago.AccountOutput")
```
muXxer: Print the actual accountName in the panic message.

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -301,6 +380,169 @@ func (w *Wallet) TransitionImplicitAccountToAccountOutput(transactionName string
...
+	if !isSimple {
+		panic("only simple token schemes supported")
+	}
+	tokenScheme := &iotago.SimpleTokenScheme{
+		MaximumSupply: previousTokenScheme.MaximumSupply,
+		MeltedTokens:  previousTokenScheme.MeltedTokens,
+		MintedTokens:  previousTokenScheme.MintedTokens.Add(previousTokenScheme.MintedTokens, big.NewInt(1)),
+	}
+
+	outputFoundry := builder.NewFoundryOutputBuilderFromPrevious(inputFoundry).
```
muXxer, line 476: Should we check if the previous foundry already had a token feature? Otherwise we would burn those tokens and the native token sum wouldn't be balanced in the tx.

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -301,6 +380,169 @@ func (w *Wallet) TransitionImplicitAccountToAccountOutput(transactionName string
...
+	manaDecayProvider := apiForSlot.ManaDecayProvider()
+	storageScoreStructure := apiForSlot.StorageScoreStructure()
+
+	totalInputMana := lo.PanicOnErr(vm.TotalManaIn(manaDecayProvider, storageScoreStructure, w.currentSlot, vm.InputSet{input.OutputID(): input.Output()}, vm.RewardsInputSet{}))
+
+	var allotments iotago.Allotments
+	for _, accountID := range accountIDs {
+		allotments = append(allotments, &iotago.Allotment{
+			AccountID: accountID,
+			Mana:      totalInputMana / iotago.Mana(len(accountIDs)),
```
muXxer: what about a potential remainder here?

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -301,6 +380,169 @@ func (w *Wallet) TransitionImplicitAccountToAccountOutput(transactionName string
...
+			CommitmentID: w.Node.Protocol.Engines.Main.Get().SyncManager.LatestCommitment().Commitment().MustID(),
+		}),
+	)
+
+	return signedTransaction
+}
+
+func (w *Wallet) AllotManaFromInput(transactionName string, inputName string, accountIDs ...iotago.AccountID) *iotago.SignedTransaction {
+	input := w.Output(inputName)
+	output := &iotago.BasicOutput{
```
muXxer, line 512: Should we only allow basic outputs as input, or do we burn account outputs etc?

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -338,13 +580,108 @@ func (w *Wallet) CreateBasicOutputsEquallyFromInput(transactionName string, outp
...
+
+	manaAmount := totalInputMana / iotago.Mana(outputCount)
+	remainderMana := totalInputMana
+
+	tokenAmount := inputAmount / iotago.BaseToken(outputCount)
+	remainderFunds := inputAmount
+
+	outputStates := make(iotago.Outputs[iotago.Output], 0, outputCount)
+	for i, index := range addressIndexes {
+		if i+1 == outputCount {
```
muXxer, line 613: This works, but I liked the way you did it in another function.
There you calculated the remainder outside of the loop and simply added it to the last. That was easier to read.

```
remainderBaseToken := input.BaseTokenAmount() - outputBaseToken*iotago.BaseToken(outputCount)
remainderMana := input.StoredMana() - outputMana*iotago.Mana(outputCount)

```

pkg/testsuite/benchmarks_test.go
```diff
@@ -0,0 +1,578 @@
...
+	standardBlockCost := iotago.Mana(500_000)
+	// dataByteRatio is the ratio of the workScore due to the dataByte factor to that due to the rest of the factors.
+	dataByteRatio := 0.5
+	// calculate the workScoreParameters from the coefficients based on a standardBlockCost of 500,000 Mana and dataByteRatio of 0.5
+	workScoreParams := workScoreParamsFromCoefficients(coeffs, standardBlock, standardBlockCost, dataByteRatio)
+	ts := NewTestSuite(t, WithProtocolParametersOptions(iotago.WithWorkScoreOptions(
+		workScoreParams.DataByte, workScoreParams.Block, workScoreParams.Input, workScoreParams.ContextInput,
+		workScoreParams.Output, workScoreParams.NativeToken, workScoreParams.Staking, workScoreParams.BlockIssuer,
+		workScoreParams.Allotment, workScoreParams.SignatureEd25519,
+	)))
```
muXxer:
```suggestion
	ts := NewTestSuite(t, WithProtocolParametersOptions(iotago.WithWorkScoreOptions(
		workScoreParams.DataByte,
		workScoreParams.Block,
		workScoreParams.Input,
		workScoreParams.ContextInput,
		workScoreParams.Output,
		workScoreParams.NativeToken,
		workScoreParams.Staking,
		workScoreParams.BlockIssuer,
		workScoreParams.Allotment,
		workScoreParams.SignatureEd25519,
	)))
```

pkg/testsuite/benchmarks_test.go
```diff
@@ -0,0 +1,578 @@
...
+			b.StopTimer()
+			ts := NewTestSuite(t)
+			node := ts.AddValidatorNode("node1")
+			ts.AddDefaultWallet(node)
+			ts.Run(true)
+			var addressIndexes []uint32
+			for i := 0; i < numIn; i++ {
+				addressIndexes = append(addressIndexes, uint32(i))
+			}
+			// First, create 128 outputs
```
muXxer: the comment seems wrong

pkg/testsuite/benchmarks_test.go
```diff
@@ -0,0 +1,578 @@
...
+				blockchan <- block
+			})
+			node.Protocol.IssueBlock(modelBlock)
+			<-blockchan
+
+			inputNames := make([]string, numIn)
+			for i := 0; i < numIn; i++ {
+				inputNames[i] = fmt.Sprintf("tx1:%d", i)
+			}
+			// Then, create a transaction with 128 inputs
```
muXxer: here as well

pkg/testsuite/benchmarks_test.go
```diff
@@ -0,0 +1,578 @@
...
+	block := ts.IssueBasicBlockWithOptions("block", ts.DefaultWallet(), tx, mock.WithSlotCommitment(genesisCommitment))
+
+	return block.ProtocolBlock()
+}
+
+func basicInBasicOut(t *testing.T, numIn int, numOut int, signatures bool) (float64, []float64) {
+	blockchan := make(chan *blocks.Block, 1)
+	var block *iotago.Block
+
+	// basic block with one input and one output
```
muXxer: move those comments above the function signature for all of these functions.

pkg/testsuite/benchmarks_test.go
```diff
@@ -0,0 +1,578 @@
...
+		}
+	}
+	// get the ns/op of processing the block
+	nsPerBlock := float64(testing.Benchmark(fn).NsPerOp())
+	fmt.Printf("One input, %d allotments: %f ns/op\n", numAllotments, nsPerBlock)
+	// get the regressors
+	regressors := getBlockWorkScoreRegressors(block)
+	printRegressors(regressors)
+
+	return nsPerBlock, regressors
```
muXxer: A lot of the code repeats itself for all the different scenarios. Maybe we can abstract that into one function, that does all the "stop timer, init test suite, call func to prepare genesis, hook to the block event, call func for the specific test case, start timer and issue block, etc..."

pkg/testsuite/benchmarks_test.go
```diff
@@ -0,0 +1,578 @@
...
+	basicBlockBody, isBasic := block.Body.(*iotago.BasicBlockBody)
+	if !isBasic {
+		panic("block body is not a basic block body")
+	}
+	signedTx, isSignedTx := basicBlockBody.Payload.(*iotago.SignedTransaction)
+	if !isSignedTx {
+		panic("block payload is not a signed transaction")
+	}
+	// add one to the Input regressor for each input
+	regressors[0] += float64(len(signedTx.Transaction.TransactionEssence.Inputs))
```
muXxer: Maybe we should define an enum for those regressor indexes to avoid mistakes in the future.

pkg/testsuite/benchmarks_test.go
```diff
@@ -0,0 +1,578 @@
...
+// The parameters are scaled such that the the standardBlock has workScore of standardBlockMinCost. This represents the Mana cost
+// of the "standard block" when the reference Mana cost (RMC) is at its minimum value of 1.
+// The dataByteRatio is the ratio of the part of the WorkScore due to the DataBytes factor to that due to the rest of the factors.
+func workScoreParamsFromCoefficients(coeffs []float64, standardBlock *iotago.Block, standardBlockMinCost iotago.Mana, dataByteRatio float64) iotago.WorkScoreParameters {
+	standarBlockRegressors := getBlockWorkScoreRegressors(standardBlock)
+	fmt.Printf("Standard block regressors: %+v\n", standarBlockRegressors)
+	standardBlockWorkScore := coeffs[0]
+	for i, regressor := range standarBlockRegressors {
+		standardBlockWorkScore += regressor * coeffs[i+1]
+	}
```
muXxer: Can you add some comments there about the indexes? I don't get why it is "i+1" on the first view.

muXxer (changes requested):

pkg/testsuite/mock/wallet_transactions.go
cyberphysic4l: We are actually already using the nativeTokenAmount from an existing foundry output and increasing it by one. But what we should check here is that the max supply is not already minted as this would cause the tx to be invalid.

muXxer (approved):

PR #729, opened 2024-02-12
Fix eviction 

### LINKED ISSUE #714: WarpSync failed to verify mutation proof
During testing of #673 in Docker Network, when running guava spam, acceptance stopped due to a blowball and large number of tips. Nodes force committed and then when trying to agree on a common chain, non-validator nodes reported the error below, which prevented them from successfully switching the chain.

piotrm50:
This PR fixes the issues with atomically advancing internal state and eviction described in more detail in https://github.com/iotaledger/iota-core/issues/714#issuecomment-1938606407. 

Additionally, following minor changes:
- Change Docker network liveness options presets to `minCommittableAge=3`, `maxCommittableAge=6`, `epoch length=128 slots`
- Add metric collection for accepted transactions within a commitment
- Add panel to `Commitments Overview` Grafana dashboard to display accepted block and transaction count for each commitment (per node)
- fix commitment debug API to correctly show chains (even if a commitment's children are missing)


pkg/protocol/engine/ledger/ledger/ledger.go
```diff
@@ -192,24 +191,30 @@ func (l *Ledger) CommitSlot(slot iotago.SlotIndex) (stateRoot iotago.Identifier,
...
 		tx.Commit()
+
+		//nolint:forcetypeassert // we only handle a single type of transaction currently
+		mutations = append(mutations, tx.Transaction().(*iotago.Transaction))
+
 		return true
 	})
 
-	return l.utxoLedger.StateTreeRoot(), stateDiff.Mutations().Root(), l.accountsLedger.AccountsTreeRoot(), outputs, spenders, nil
+	l.memPool.Evict(slot)
```
jonastheis, line 218: We should add a comment here why the eviction is called separately at the end

jonastheis (approved):

PR #737, opened 2024-02-14
Fix chain managment logic

### LINKED ISSUE #634: Memory Usage
We have the problem that memory fills up slowly but steady:

image

Here is a pprof profile from one node:

[iota-core.zip](https://github.com/iotaledger/iota-core/files/13691689/iota-core.zip)

A perhaps helpful additional information is that after a restart it consumes about the same amount of memory after about 1-2 minutes, so it doesn't seem to be a leak but really be used by something.

Unfortunately the only option we have right now to workaround this is to completly wipe the nodes and start with a fresh network.

edit:
after restarting it consumes even more:
image

and golang heap:
image

A second profile after restarting:
[iota-core.zip](https://github.com/iotaledger/iota-core/files/13691720/iota-core.zip)

hmoog:
This PR fixes a memleak in the Protocol, so that we correctly evict old commitments and improves the way we manage chains and commitments in the Protocol.

This also extends the engine switching test to assert that the state of the commitments and chains is as it should be. To make that possible, we also refactored the engine switching tests to be predictable.


pkg/protocol/chain.go
```diff
@@ -188,11 +192,35 @@ func (c *Chain) initLogger() (shutdown func()) {
 
 // initDerivedProperties initializes the behavior of this chain by setting up the relations between its properties.
 func (c *Chain) initDerivedProperties() (shutdown func()) {
+	markChainEvicted := func() {
```
jonastheis: Let's put this into a worker

pkg/protocol/commitment.go
```diff
@@ -309,7 +301,7 @@ func (c *Commitment) deriveChain(parent *Commitment) func() {
 
 		// if we are not the main child of our parent, we spawn a new chain
 		if c != mainChild {
-			if currentChain == nil {
```
jonastheis, line 312: Can we add some more comments on this function? This is critical to decide how and when we spawn new chains if new commitments are received and when we evict after advancing the root commitment.

pkg/tests/protocol_engine_switching_test.go
```diff
@@ -376,24 +382,54 @@ func TestProtocol_EngineSwitching(t *testing.T) {
...
+
+		ts.AssertMainChain(ts.CommitmentOfMainEngine(nodesP1[0], oldestNonEvictedCommitment).ID(), ts.Nodes()...)
+		ts.AssertUniqueCommitmentChain(ts.Nodes()...)
+		ts.AssertLatestEngineCommitmentOnMainChain(ts.Nodes()...)
+		ts.AssertCommitmentsOnEvictedChain(ultimateCommitmentsP2, true, ts.Nodes()...)
+
+		ts.AssertCommitmentsOnEvictedChain(commitmentsMainChain, false, ts.Nodes()...)
+
+		ts.AssertCommitmentsOnChain(commitmentsMainChain, ts.CommitmentOfMainEngine(nodesP1[0], oldestNonEvictedCommitment).ID(), ts.Nodes()...)
+		// TODO: check that its not on the main chain
```
jonastheis: Let's resolve this TODO

pkg/tests/protocol_engine_switching_test.go
```diff
@@ -650,6 +692,25 @@ func TestProtocol_EngineSwitching_CommitteeRotation(t *testing.T) {
...
+	})
+
+	// Assert Protocol.Chains and Protocol.Commitments state.
+	{
+		ts.AssertUniqueCommitmentChain(ts.Nodes()...)
+		ts.AssertLatestEngineCommitmentOnMainChain(ts.Nodes()...)
+		ts.AssertCommitmentsOnEvictedChain(ultimateCommitmentsP2, true, ts.Nodes()...)
+		ts.AssertCommitmentsOnEvictedChain(commitmentsMainChain, false, ts.Nodes()...)
+		ts.AssertCommitmentsOnChain(commitmentsMainChain, ts.CommitmentOfMainEngine(nodesP1[0], oldestNonEvictedCommitment).ID(), ts.Nodes()...)
+		// TODO: check that its not on the main chain
```
jonastheis: Let's resolve this TODO

pkg/tests/protocol_engine_switching_test.go
```diff
@@ -938,29 +1011,101 @@ func TestProtocol_EngineSwitching_Tie(t *testing.T) {
...
+		// P3 commitments on P2 node should be on separate chain.
+		ts.AssertCommitmentsOnChain(ultimateCommitmentsP3, ultimateCommitmentsP3[0].ID(), otherPartitions[1])
+
+		ts.AssertCommitmentsAndChainsEvicted(5, ts.Nodes()...)
+	}
+
+	fmt.Println("---------------------")
+	// Finalize further slot and make sure the nodes have the same state of chains.
+	{
+		//ts.CommitUntilSlot(forkingSlot+maxCommittableAge+2, ts.BlockIDsWithPrefix(fmt.Sprintf("P%d-merge:", mainPartitionIndex))...)
```
jonastheis: remove

piotrm50 (approved):

PR #753, opened 2024-02-19
Allot more mana at account creation in docker test

jkrvivian:
`Test_Delegation` and `Test_AccountTransitions` failed because the account ran out of BIC, which became negative after issuing account transition transaction.

The solution is to request faucet funds and then include it in the account transition transaction, which gives more mana to allot to the account.

`IncreaseBIC` is added to the docker TestFramework, it is used to gain more BIC from faucet to an account. Could be useful for further testing.


muXxer:
Please do not merge this PR yet. The problem seems to be somewhere else. Even if I increase the faucet payouts, it seems to send "mana 0" from time to time. That is the underlying problem.

muXxer:
I reverted the fix, and only left the new `IncreaseBIC` function for future usage, because the underlying problem was fixed in #763

muXxer (approved):

PR #758, opened 2024-02-20
Autopeering improvements

alexsporn:
- Use a more explicit namespace for DHT discovery.
- Use a shrinking map for holding neighbors to avoid long-term leaks.
- Use a distinct network-name everytime the docker or feature network snapshots are created to avoid collisions in autopeering.
- Use autopeering bootstrap peers in the docker network to speed up peer discovery.
- Let nodes wait until node-1 is healthy before starting up. This, together with the autopeering improvements speeds up the network bootstrapping.
- Introduced a max number of peers for autopeering. If set to `0` autopeering won't be started.
- Stop advertising for autopeering if the max number of peers is reached.


muXxer (approved):

pkg/network/p2p/manager.go
```diff
@@ -197,31 +224,41 @@ func (m *Manager) DropNeighbor(id peer.ID) error {
...
 	for _, nbr := range neighbors {
 		nbr.Enqueue(packet, protocolID)
 	}
 }
 
-// AllNeighbors returns all the neighbors that are currently connected.
-func (m *Manager) AllNeighbors() []*Neighbor {
+func (m *Manager) AllNeighbors() []network.Neighbor {
+	neighbors := m.allNeighbors()
+	result := make([]network.Neighbor, len(neighbors))
```
muXxer, line 238: I think you could also use `lo.Values` here.

muXxer (approved):

pkg/network/p2p/manager.go
alexsporn: No because I need to cast from `[]*neighbor` to `[]network.Neighbor`

muXxer (approved):

pkg/network/errors.go
```diff
@@ -0,0 +1,18 @@
...
+	// ErrNotRunning is returned when a peer is added to a stopped or not yet started network manager.
+	ErrNotRunning = ierrors.New("manager not running")
+	// ErrUnknownPeer is returned when the specified peer is not known to the network manager.
+	ErrUnknownPeer = ierrors.New("unknown neighbor")
+	// ErrLoopbackPeer is returned when the own peer is added.
+	ErrLoopbackPeer = ierrors.New("loopback connection not allowed")
+	// ErrDuplicatePeer is returned when the same peer is added more than once.
+	ErrDuplicatePeer = ierrors.New("already connected")
+	// ErrFirstPacketNotReceived is returned when the first packet from a peer is not received.
+	ErrFistPacketNotReceived = ierrors.New("first packet not received")
```
MiKeF05: Small typo in `ErrFistPacketNotReceived `

Issue #752, opened 2024-02-19
Create API Cache package


muXxer:
The API Cache package should cache results of certain API calls in memory, e.g. the `validators` endpoint.

It is important that this component does not create wrong results due to chain switching.
Test cases for this should be added. The component should also have a limited size for the cached elements, otherwise the node could go OOM. It should not create wrong cached results. For example the validators for the current epoch can change with every new slot commitment, so we need to invalidate the last cached result with every slot. For past epochs the values can't change, so we can keep those results in the cache.

To limit the max size, we could use https://github.com/VictoriaMetrics/fastcache.

I went through the API, for now it only seems to make sense to cache the `validators` endpoint.

muXxer:
We should also add a mapping here between internal errors and the HTTP error codes or RPC error codes that should be returned.
For example here https://github.com/iotaledger/iota-core/pull/766#discussion_r1497969669

PR #766, opened 2024-02-21
Hook up retainer events

cyberphysic4l:
Change event links for retainer and api so that blocks are only retained once they are booked, and api only returns when blocks are retained.

to be merged alongside https://github.com/iotaledger/iota.go/pull/696


components/restapi/core/blocks.go
```diff
@@ -121,16 +120,7 @@ func sendBlock(c echo.Context) (*api.BlockCreatedResponse, error) {
...
-		case ierrors.Is(err, blockhandler.ErrBlockAttacherInvalidBlock):
-			return nil, ierrors.Wrapf(httpserver.ErrInvalidParameter, "failed to attach block: %w", err)
-
-		case ierrors.Is(err, blockhandler.ErrBlockAttacherAttachingNotPossible):
-			return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to attach block: %w", err)
-
-		default:
-			return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to attach block: %w", err)
-		}
+		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to attach block: %w", err)
```
muXxer, line 109: Should really every error that could potentially happen be returned as 500 internal server error?
I think some of the possible errors should still be user errors, which should return `http.StatusBadRequest` 400 error, for example like the `ErrInvalidBlockVersion`, `ErrBlockTimeTooFarAheadInFuture`, `ErrValidatorNotInCommittee`, `ErrBlockParentNotFound`, `ErrBlockIssuingTimeNonMonotonic` etc...

components/inx/server_blocks.go
```diff
@@ -190,16 +189,7 @@ func (s *Server) attachBlock(ctx context.Context, block *iotago.Block) (*inx.Blo
...
-		case ierrors.Is(err, blockhandler.ErrBlockAttacherInvalidBlock):
-			return nil, status.Errorf(codes.InvalidArgument, "failed to attach block: %s", err.Error())
-
-		case ierrors.Is(err, blockhandler.ErrBlockAttacherAttachingNotPossible):
-			return nil, status.Errorf(codes.Internal, "failed to attach block: %s", err.Error())
-
-		default:
-			return nil, status.Errorf(codes.Internal, "failed to attach block: %s", err.Error())
-		}
+		return nil, status.Errorf(codes.Internal, "failed to attach block: %s", err.Error())
```
muXxer, line 191: Same here as mentioned in the core api.

pkg/retainer/events.go
```diff
@@ -0,0 +1,21 @@
+package retainer
+
+import (
+	"github.com/iotaledger/hive.go/runtime/event"
+	"github.com/iotaledger/iota-core/pkg/protocol/engine/blocks"
+)
+
+// Events is a collection of Tangle related Events.
```
muXxer:
```suggestion
// Events is a collection of Retainer related Events.
```

components/restapi/core/blocks.go
muXxer: We should do this in another PR maybe in the API package that also contains the caches. I will create an issue.

I updated [#752](https://github.com/iotaledger/iota-core/issues/752) to add this.

PR #769, opened 2024-02-22
Add transaction retainer

### LINKED ISSUE #697: Store block and transaction failure error message in retainer
To improve our developer experience, we should allow a node to be configured, such that the full error message string of a block or transaction failure (perhaps transactions only is sufficient) is stored in the retainer and returned in the `/api/core/v3/blocks/{blockId}/metadata` and/or `/api/core/v3/transactions/{transactionId}/metadata` APIs.

That way, developers have much more information for debugging.

By default, the option should be disabled but all our test networks as well as the docker-network should have that option enabled.

## Example

Instead of just getting `TxFailureDelegationStartEpochInvalid`, the full error message also includes the context of which delegation output was problematic and what start epoch was expected:

```
failed to execute transaction: invalid chain transition; inner err: invalid delegation output transition: invalid start epoch (is 5, expected 1); Delegation 0x0000000000000000000000000000000000000000000000000000000000000000: new chain 0x8b0056a3897eb4e0e7b66641e566c800287bed155499d1c8052704b9332cf013 (*iotago.DelegationOutput) state transition failed
```

### LINKED ISSUE #751: Create TransactionRetainer component
The TransactionRetainer is a component which stores the state and failure reasons for transactions in a relational database (SQLite) because that is a better fit for this usecase than a key value store.

The component needs to handle Pruning and Chain Switching (Reset method) correctly. We should also add the optional error message mentioned in https://github.com/iotaledger/iota-core/issues/697 while we are at it.

We should do test driven development for these components.
The tests should include at least the following scenarios:

    - First set object as pending
    - Overwrite with new state, check if potential former fields are correctly set/unset.
    - Check if values can be retrieved after setting them.
    - Reset at forking point and check if state is correctly set (newer objects deleted, old still exist)
    - Check if pruning works correctly
    - Check if transaction with invalid signatures do not overwrite the state of transactions with valid signatures.

### LINKED ISSUE #800: Add Rollback to Block/TransactionRetainer
We need to add a functionality to `Rollback` and `Reset` the `TransactionRetainer` component.

- Use another Table that logs the changes?
- Use an AuditLog in gorm? (https://aayushacharya.com.np/blog/audit-log-gorm/)
- Should we add a cache to the `TransactionRetainer`?

muXxer:
The Transaction Retainer introduces an SQLite database to store all transaction related metadata.
It caches all updates in memory, until the slot of the earliest attachment of a transaction is smaller or equal to the latest committed slot. If that happens, all the updates of that slot are written in an atomic transaction to the database.


components/prometheus/metrics_db.go
```diff
@@ -7,9 +7,9 @@ import (
 const (
 	dbNamespace = "db"
 
-	sizeBytesPermanent = "size_bytes_permanent"
-	sizeBytesPrunable  = "size_bytes_prunable"
-	sizeBytesRetainer  = "size_bytes_retainer"
+	sizeBytesPermanent   = "size_bytes_permanent"
+	sizeBytesPrunable    = "size_bytes_prunable"
```
PhilippGackstatter: Should we rename this to `sizeBytesPrunableKV` or something like that, to make clear that this is distinct from `sizeBytesPrunableSQL` and does not include it?

pkg/protocol/engine/mempool/errors.go
```diff
@@ -1,5 +1,10 @@
...
 
-import "github.com/iotaledger/hive.go/ierrors"
+import (
+	"github.com/iotaledger/hive.go/ierrors"
+)
 
-var ErrStateNotFound = ierrors.New("state not found")
+var (
+	ErrStateNotFound = ierrors.New("state not found")
+	ErrRequestFailed = ierrors.New("request failed")
```
PhilippGackstatter: Being more specific here would be great, as "request failed" is very generic. Since we use it only in one place where the error is about solidification of inputs, perhaps:

```suggestion
	ErrInputSolidificationRequestFailed = ierrors.New("UTXO input solidification failed")
```

pkg/retainer/txretainer/database.go
```diff
@@ -0,0 +1,182 @@
...
+	return r.handleError(err, "failed to close database")
+}
+
+func (r *TransactionRetainer) updateTxMetadata(txID iotago.TransactionID, newTxMeta *txMetadata) {
+	err := r.dbExecFunc(func(db *gorm.DB) error {
+		return db.Transaction(func(tx *gorm.DB) error {
+			txMeta := &txMetadata{}
+
+			// init the txMeta if it does not exist yet, or get the existing one from the database
+			tx.FirstOrInit(txMeta, &txMetadata{TransactionID: txID[:]})
```
PhilippGackstatter:
```suggestion
			dbTx.FirstOrInit(txMeta, &txMetadata{TransactionID: txID[:]})
```

Calling this `dbTx` makes this slightly more understandable with the mix of tx terminology from the db and ours.

pkg/retainer/txretainer/testsuite_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	latestCommitedSlot iotago.SlotIndex
+	finalizedSlot      iotago.SlotIndex
+}
+
+func newTestSuite(t *testing.T) *txRetainerTestsuite {
+	dbParams := sql.DatabaseParameters{
+		Engine:   db.EngineSQLite,
+		Path:     t.TempDir(),
+		Filename: "test.db",
```
PhilippGackstatter:
```suggestion
		Filename: "txRetainerTestSQLite.db",
```

In case we ever want to find the test file locally, it's better to have a more descriptive name. For this to make any sense, my assumption is that the file is not immediately deleted after the test.

pkg/retainer/txretainer/transaction_retainer.go
```diff
@@ -0,0 +1,315 @@
...
+	"github.com/iotaledger/hive.go/runtime/workerpool"
+	"github.com/iotaledger/iota-core/pkg/protocol/engine"
+	"github.com/iotaledger/iota-core/pkg/protocol/engine/mempool"
+	"github.com/iotaledger/iota-core/pkg/retainer"
+	"github.com/iotaledger/iota-core/pkg/storage"
+	iotago "github.com/iotaledger/iota.go/v4"
+	"github.com/iotaledger/iota.go/v4/api"
+)
+
+type Status struct {
```
PhilippGackstatter: A comment/description for what this represents and what it's used for would be nice.

pkg/retainer/txretainer/transaction_retainer.go
```diff
@@ -0,0 +1,315 @@
...
+	_ retainer.TransactionRetainer = &TransactionRetainer{}
+)
+
+type (
+	DatabaseFunc            func(func(*gorm.DB) error) error
+	LatestCommittedSlotFunc func() iotago.SlotIndex
+	FinalizedSlotFunc       func() iotago.SlotIndex
+)
+
+// TransactionRetainer keeps and resolves all the information needed in the API and INX.
```
PhilippGackstatter:
```suggestion
// TransactionRetainer keeps and resolves all the transaction-related data needed in the API and INX.
```

pkg/retainer/txretainer/transaction_retainer.go
```diff
@@ -0,0 +1,315 @@
...
+					State:                  byte(api.TransactionStatePending),
+					FailureReason:          byte(api.TxFailureNone),
+					ErrorMsg:               nil,
+				})
+
+				/*
+					confirmed and finalized logic need to be done in the getter
+					with the slot of the earliest included attachment
+						first check if the earliest included attachment slot is finalized
+						if not, check if the attachment itself was confirmed
```
PhilippGackstatter: This needs to be added to `func (r *TransactionRetainer) TransactionMetadata`.

pkg/retainer/txretainer/transaction_retainer.go
```diff
@@ -0,0 +1,315 @@
...
+					/*
+						// TransactionStateFailed indicates that the transaction has not been executed by the node due to a failure during processing.
+						TransactionStateFailed
+				*/
```
PhilippGackstatter: Leftover commented code?

pkg/retainer/txretainer/transaction_retainer.go
```diff
@@ -0,0 +1,315 @@
...
+						FailureReason:          0,
+						ErrorMsg:               nil,
+					})
+				})
+
+				// a conflict that lost => double spent
+				transactionMetadata.OnRejected(func() {
+					e.LogTrace("TxRetainer.TransactionRejected", "tx", transactionMetadata.ID())
+
+					err := iotago.ErrInputAlreadySpent
```
PhilippGackstatter: I think this should be the error:

```suggestion
					err := iotago.ErrTxConflictRejected
```

pkg/retainer/txretainer/transaction_retainer.go
```diff
@@ -0,0 +1,315 @@
...
+					})
+				})
+
+				// on invalid is when the transaction execution failed
+				transactionMetadata.OnInvalid(func(err error) {
+					e.LogTrace("TxRetainer.TransactionInvalid", "tx", transactionMetadata.ID(), "err", err)
+
+					// if it's not the ErrRequestFailed, the error comes from the transaction execution
+					// and therefore must have had a valid signature. If it's ErrRequestFailed, the signature
+					// check is not done yet, but we are fine storing the error reason as invalid signature.
```
PhilippGackstatter: If you update the error name as I suggested elsewhere, also update this comment.

pkg/retainer/txretainer/transaction_retainer.go
```diff
@@ -0,0 +1,315 @@
...
+						TransactionID:          txID[:],
+						ValidSignature:         !isMempoolRequestFailedError,
+						EarliestAttachmentSlot: transactionMetadata.EarliestIncludedAttachment().Slot(),
+						State:                  byte(api.TransactionStateFailed),
+						FailureReason:          byte(api.DetermineTransactionFailureReason(err)),
+						ErrorMsg:               getTxErrMsg(err),
+					})
+				})
+
+				// 1. attached to the wrong place in the tangle, was not accepted within minCommitmentAge
```
PhilippGackstatter:
```suggestion
				// 1. attached to the wrong place in the tangle, was not accepted within minCommittableAge
```

pkg/retainer/txretainer/transaction_retainer.go
```diff
@@ -0,0 +1,315 @@
...
+				// 2. one of the inputs is orphaned, so the other transaction that produced that input got orphaned
+				// 3. a conflicting tx was commited before this one, so this one gets orphaned
+				transactionMetadata.OnOrphanedSlotUpdated(func(slot iotago.SlotIndex) {
+					e.LogTrace("TxRetainer.TransactionOrphanedSlotUpdated", "tx", transactionMetadata.ID(), "slot", slot)
+
+					//validSignature is unclear
+					// TODO: wrong error reason
+					err := iotago.ErrInputAlreadySpent
```
PhilippGackstatter: Wouldn't this warrant a new `TransactionFailureReason`, e.g. `TransactionFailureOrphaned`, and constructing a custom `ErrorMsg` ala `("transaction orphaned in slot %d", slot)`? We can keep `TransactionStateFailed` in that case I think.

pkg/retainer/txretainer/transaction_retainer.go
```diff
@@ -0,0 +1,315 @@
...
+						EarliestAttachmentSlot: getEarliestIncludedAttachmentSlot(signedTransactionMetadata),
+						State:                  byte(api.TransactionStateFailed),
+						FailureReason:          byte(api.DetermineTransactionFailureReason(err)),
+						ErrorMsg:               getTxErrMsg(err),
+					})
+				})
+			})
+		})
+
+		//e.Events.Retainer.LinkTo(r.events)
```
PhilippGackstatter: Leftover?

pkg/retainer/txretainer/transaction_retainer.go
```diff
@@ -0,0 +1,315 @@
...
+
+				// attachment with invalid signature detected
+				// store that there is an attachment with an invalid signature.
+				signedTransactionMetadata.OnSignaturesInvalid(func(err error) {
+					txID := signedTransactionMetadata.TransactionMetadata().ID()
+
+					r.updateTxMetadata(txID, &txMetadata{
+						TransactionID:          txID[:],
+						ValidSignature:         false,
+						EarliestAttachmentSlot: getEarliestIncludedAttachmentSlot(signedTransactionMetadata),
```
PhilippGackstatter: The `getEarliestIncludedAttachmentSlot` is only used here, perhaps it's better to inline it?

pkg/retainer/txretainer/transaction_retainer.go
```diff
@@ -0,0 +1,315 @@
...
+		r.errorHandler(ierrors.Wrap(err, "failed to reset transaction retainer"))
+	}
+}
+
+// Prune prunes the component state as if the last pruned slot was targetSlot.
+func (r *TransactionRetainer) Prune(targetSlot iotago.SlotIndex) error {
+	return r.deleteTxMetadataWithSlotSmallerOrEqualThan(targetSlot)
+}
+
+func (r *TransactionRetainer) UpdateTransactionMetadata(txID iotago.TransactionID, validSignature bool, earliestAttachmentSlot iotago.SlotIndex, state api.TransactionState, txErr error) {
```
PhilippGackstatter: This function is only used in tests, but it is part of the `TransactionRetainer` interface - is that intended? If the retainer is informed about all relevant tx updates via events, then we can remove that function from the interface at least, so that no one accidentally updates the retainer directly. Then we could consider renaming the function too, to indicate it's test-only nature, or at least adding a comment to it, that it is only used in tests.

pkg/retainer/txretainer/transaction_retainer_test.go
```diff
@@ -0,0 +1,418 @@
...
+				}
+			},
+			testFunc: nil,
+			resultFunc: func() *api.TransactionMetadataResponse {
+				return &api.TransactionMetadataResponse{
+					TransactionID:             txID,
+					TransactionState:          api.TransactionStateFailed,
+					EarliestAttachmentSlot:    0,
+					TransactionFailureReason:  api.DetermineTransactionFailureReason(iotago.ErrInputAlreadySpent),
+					TransactionFailureDetails: "input already spent",
```
PhilippGackstatter: I think it's better to insert a reference rather than a copy here. If we ever update the error message in iota.go we would also have to touch iota-core with the current code.

```suggestion
					TransactionFailureDetails: iotago.ErrInputAlreadySpent.Error(),
```

Ideally do the same for the rest of the file as well.

pkg/retainer/txretainer/transaction_retainer_test.go
```diff
@@ -0,0 +1,418 @@
...
+					State:                  api.TransactionStatePending,
+					TransactionError:       nil,
+				}
+			},
+			testFunc: nil,
+			resultFunc: func() *api.TransactionMetadataResponse {
+				return &api.TransactionMetadataResponse{
+					TransactionID:             txID,
+					TransactionState:          api.TransactionStatePending,
+					EarliestAttachmentSlot:    10,
```
PhilippGackstatter: The test name says the attachment slot is lower, but it's the same here?

pkg/retainer/txretainer/transaction_retainer_test.go
```diff
@@ -0,0 +1,418 @@
...
+	ts := newTestSuite(t)
+	defer ts.Close()
+
+	tr := ts.TxRetainer
+
+	txID := tpkg.RandTransactionID()
+
+	// IMPORTANT: all those tests modify the same transaction metadata and depend on each other
+	for _, test := range []*test{
+		// ok - prune below the earliest attachment slot
```
PhilippGackstatter: Nit: Some functions here expect an error but are named `"ok - ..."`.

pkg/retainer/txretainer/transaction_retainer_test.go
```diff
@@ -0,0 +1,418 @@
...
+				return &api.TransactionMetadataResponse{
+					TransactionID:             txID,
+					TransactionState:          api.TransactionStatePending,
+					EarliestAttachmentSlot:    10,
+					TransactionFailureReason:  api.TxFailureNone,
+					TransactionFailureDetails: "",
+				}
+			},
+			targetErr: nil,
+		},
```
PhilippGackstatter: Consider adding another test case where the new metadata has `ValidSignature: false` and then verify that nothing is modified?

pkg/storage/prunable/slotstore/block_metadata.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/iotaledger/hive.go/ierrors"
+	"github.com/iotaledger/hive.go/kvstore"
+	"github.com/iotaledger/hive.go/serializer/v2"
+	"github.com/iotaledger/hive.go/serializer/v2/stream"
+	iotago "github.com/iotaledger/iota.go/v4"
+	"github.com/iotaledger/iota.go/v4/api"
+)
+
+const (
+	// api.BlockState.
```
PhilippGackstatter, line 13:
```suggestion
```

pkg/storage/prunable/slotstore/block_metadata.go
```diff
@@ -0,0 +1,107 @@
...
+
+	var err error
+	blockMetadata := new(BlockMetadata)
+
+	if blockMetadata.State, err = stream.Read[api.BlockState](byteReader); err != nil {
+		return nil, 0, ierrors.Wrap(err, "failed to read block state")
+	}
+
+	return blockMetadata, byteReader.BytesRead(), nil
+}
```
PhilippGackstatter, line 42: Why are we not using serix for this?

pkg/storage/prunable/slotstore/block_metadata.go
```diff
@@ -0,0 +1,107 @@
...
+
+type BlockMetadataStore struct {
+	slot               iotago.SlotIndex
```
PhilippGackstatter, line 46:
```suggestion
    // The slot for which this store holds the metadata from all blocks with the corresponding slot.
	slot               iotago.SlotIndex
```

pkg/storage/prunablesql/prunable_sql.go
```diff
@@ -0,0 +1,84 @@
...
+
+func New(parentLogger log.Logger, directory string, errorHandler func(error)) *PrunableSQL {
+	baseDir := utils.NewDirectory(directory, true)
+
+	sqlDatabase, _, err := sql.New(
+		parentLogger.NewChildLogger("sql"),
+		sql.DatabaseParameters{
+			Engine:   db.EngineSQLite,
+			Path:     baseDir.Path(),
+			Filename: "tx_retainer.db",
```
PhilippGackstatter: It seems a bit inconsistent to name the logger "sql" but the database file specific to the tx retainer. Maybe we can use `sqlite.db` as the name instead? I do think that it would be nice to have more specific names for the db files, but this seems like a hack. If we ever reuse the prunable sql for something else, it would be confusing.

pkg/storage/prunablesql/prunable_sql.go
```diff
@@ -0,0 +1,84 @@
...
+		return nil, ierrors.Wrap(err, "failed to copy prunable storage directory to new storage path")
+	}
+
+	return New(parentLogger, directory, errorHandler), nil
+}
+
+//func (p *PrunableSQL) RestoreFromDisk() (lastPrunedEpoch iotago.EpochIndex) {
+//
+//	return
+//}
```
PhilippGackstatter:
```suggestion
```

pkg/storage/storage.go
```diff
@@ -122,25 +134,48 @@ func (s *Storage) Directory() string {
...
 func (s *Storage) Size() int64 {
-	return s.PermanentDatabaseSize() + s.PrunableDatabaseSize()
+	return s.PermanentDatabaseSize() + s.PrunableDatabaseSize() + s.PrunableSQLDatabaseSize()
+}
+
+func (s *Storage) RestoreFromDisk() {
+	s.pruningLock.Lock()
+	defer s.pruningLock.Unlock()
+
+	lastPrunedEpoch := s.prunable.RestoreFromDisk()
```
PhilippGackstatter, line 164: So this means we have the lastPrunedEpoch for the prunable and prunable sql store stored in the prunable store. We have to make sure that prunable SQL cannot be in an inconsistent state (i.e. that it retains data that is older than lastPrunedEpoch) when we hook that up with the pruning. Prune on the TransactionRetainer is currently not called outside of tests, so I guess that is yet to be done.

pkg/storage/testframework_test.go
```diff
@@ -194,14 +199,14 @@ func uint64ToBytes[V ~uint64](v V) []byte {
...
-func (t *TestFramework) assertPrunableSizeGreater(expected int64) {
-	require.GreaterOrEqual(t.t, float64(t.Instance.PrunableDatabaseSize()), float64(expected)*0.8)
-}
-
 func (t *TestFramework) assertPermanentSizeGreater(expected int64) {
 	require.GreaterOrEqual(t.t, float64(t.Instance.PermanentDatabaseSize()), float64(expected)*0.8)
 }
 
+func (t *TestFramework) assertPrunableSizeGreater(expected int64) {
+	require.GreaterOrEqual(t.t, float64(t.Instance.PrunableDatabaseSize()), float64(expected)*0.8)
```
PhilippGackstatter, line 207: Please add a description for the function or a comment to explain the magic number.

PhilippGackstatter (commented):
I really like the approach with the events hooked up in the retainer, rather than various components calling the retainer - much cleaner and easier to understand.

Adding test within the testsuite or docker-framework test would be best to make sure than the retainer works as intended with the "ValidSignature-based" logic.

pkg/retainer/txretainer/transaction_retainer.go
muXxer: Status was removed because it wasn't used :+1:

pkg/retainer/txretainer/transaction_retainer.go
muXxer: Oh, I didn't want to push my TODO comments :D

pkg/retainer/txretainer/transaction_retainer.go
muXxer: Was removed. Everything is done by EarliestIncludedAttachment info in the `transactionMetadata` from the mempool

pkg/retainer/txretainer/transaction_retainer.go
muXxer: I removed it, this was a leftover from first tries, where the transaction retainer was all over the codebase, before I changed to the event driven approach.

pkg/storage/prunable/slotstore/block_metadata.go
muXxer: Thats not an error, it's a comment to describe the length :trollface: 
We always do that this way :sweat_smile:

pkg/storage/prunablesql/prunable_sql.go
muXxer: renamed it to `tx-retainer-db`

pkg/storage/storage.go
muXxer: I hooked to the `Pruned` event of the storage now, to trigger pruning inside the txRetainer.
There can be some inconsistencies, in case the pruning of the storage stops mid-process, but I think that's fine for now. It will fix itself with the next successful pruning. So basically it's the same inconsistency as if you call the different API endpoints during an ongoing pruning process.

pkg/storage/testframework_test.go
muXxer: I did not even modify that code, I only moved it :D

pkg/storage/prunable/slotstore/block_metadata.go
muXxer: I didn't want to touch that logic tbh, which serix API should we use and where do we get it from in that part of the storage? CommitedAPI from the protocol? It's a bit ugly as well I guess...

pkg/retainer/txretainer/transaction_retainer.go
muXxer: Added :+1:

pkg/retainer/txretainer/transaction_retainer_test.go
muXxer: Like the first comment before the test definitions says:
`// IMPORTANT: all those tests modify the same transaction metadata and depend on each other`
So, before slot 15 was set, and now it's slot 10, so it's lower.

pkg/retainer/txretainer/transaction_retainer_test.go
muXxer: Done :+1:

pkg/requesthandler/accounts.go
```diff
@@ -35,13 +35,19 @@ func (r *RequestHandler) Validators(slotRange, cursorIndex, pageSize uint32) (*a
...
+	//	registeredValidators, err := r.protocol.Engines.Main.Get().SybilProtection.OrderedRegisteredCandidateValidatorsList(latestEpoch)
+	//	if err != nil {
+	//		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to get ordered registered validators list for epoch %d : %s", latestEpoch, err)
+	//	}
+	//	r.protocol.Engines.Main.Get().BlockRetainer.RetainRegisteredValidatorsCache(slotRange, registeredValidators)
+	//}
+
+	registeredValidators, err := r.protocol.Engines.Main.Get().SybilProtection.OrderedRegisteredCandidateValidatorsList(latestEpoch)
+	if err != nil {
+		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to get ordered registered validators list for epoch %d : %s", latestEpoch, err)
```
PhilippGackstatter: - Should be `WithMessagef` I think?
- Is `%s` intended? Shouldn't it be `%w`?

pkg/retainer/retainer.go
```diff
@@ -32,13 +17,16 @@ type BlockRetainer interface {
...
+	// TransactionMetadata returns the metadata of a transaction.
 	TransactionMetadata(txID iotago.TransactionID) (*api.TransactionMetadataResponse, error)
-	RetainTransactionFailure(iotago.TransactionID, error)
 
 	// Reset resets the component to a clean state as if it was created at the last commitment.
-	Reset()
+	Reset(targetSlot iotago.SlotIndex)
+
+	// Prune prunes the component state as if the last pruned slot was targetSlot.
+	Prune(targetSlot iotago.SlotIndex) error
```
PhilippGackstatter: This is only called from tests, so it shouldn't need to be part of the interface I think?

pkg/retainer/txretainer/tx_retainer.go
```diff
@@ -0,0 +1,380 @@
...
+	"github.com/iotaledger/iota-core/pkg/storage"
+	iotago "github.com/iotaledger/iota.go/v4"
+	"github.com/iotaledger/iota.go/v4/api"
+)
+
+var (
+	// ErrEntryNotFound is returned when a transaction metadata entry is not found.
+	ErrEntryNotFound = ierrors.New("entry not found")
+
+	_ retainer.TransactionRetainer = &TransactionRetainer{}
```
PhilippGackstatter, line 27: What is this used for?

pkg/retainer/txretainer/tx_retainer.go
```diff
@@ -0,0 +1,380 @@
...
+			}, asyncOpt)
+
+			// this event is fired when the storage successfully pruned an epoch
+			e.Storage.Pruned.Hook(func(epoch iotago.EpochIndex) {
+				epochStartSlot := e.CommittedAPI().TimeProvider().EpochStart(epoch)
+				if epochStartSlot == 0 {
+					return
+				}
+
+				// pruning should be done until and including the last slot of the previous epoch
```
PhilippGackstatter: I can guess why this is the logic, but can you add a comment describing the reasoning for this?

pkg/retainer/txretainer/tx_retainer_cache.go
```diff
@@ -0,0 +1,221 @@
...
+
+	txMetadataUpdateFunc txMetadataUpdateFunc
+
+	uncommittedTxMetadataChangesByID *shrinkingmap.ShrinkingMap[iotago.TransactionID, *TransactionMetadata]
+	uncommittedTxIDsBySlot           *shrinkingmap.ShrinkingMap[iotago.SlotIndex, map[iotago.TransactionID]struct{}]
+}
+
+// defaultTxMetadataUpdateFunc is the default update function for transaction metadata in the transaction retainer cache.
+// HINT: do not modify the contents of the old transaction metadata in the update function, as it is a reference to the cache.
+func defaulTxMetadataUpdateFunc(oldTxMeta *TransactionMetadata, newTxMeta *TransactionMetadata) (*TransactionMetadata, bool, error) {
```
PhilippGackstatter: So both the `oldTxMeta` and `newTxMeta` will always come from the retainer *cache*, where no `ID` was assigned yet by the DB, correct? If so, then do we need any of the `newTxMeta.ID = oldTxMeta.ID` logic? Isn't it just overwriting `0` with `0`? I don't think you need to care about `ID` in the cache at all iiuc.

pkg/retainer/txretainer/tx_retainer_database.go
```diff
@@ -0,0 +1,177 @@
...
+var (
+	dbTables = []interface{}{
+		&TransactionMetadata{},
+	}
+)
+
+// transactionRetainerDatabase is a wrapper around the gorm database to store transaction metadata.
+type transactionRetainerDatabase struct {
+	dbExecFunc storage.SQLDatabaseExecFunc
+}
```
PhilippGackstatter, line 23: Documenting here what the storage approach is would be helpful, i.e. that the table can contain multiple metadata entries for the same transaction ID.

pkg/retainer/txretainer/tx_retainer_cache_test.go
```diff
@@ -0,0 +1,162 @@
...
+	require.Len(t, txIDs, 0)
+
+	// check that the transaction is not part of slot 20
+	txIDs = txRetainerCache.DeleteAndReturnTxMetadataChangesBySlot(20)
+	require.Len(t, txIDs, 0)
+
+	// check that the transaction metadata doesn't exist anymore
+	txMeta, exists = txRetainerCache.TransactionMetadataByID(txID1)
+	require.False(t, exists)
+	require.Nil(t, txMeta)
```
PhilippGackstatter: This is the exact same check as a couple lines above. Copy-paste error or intended :thinking:?

pkg/storage/testframework_test.go
PhilippGackstatter: :crying_cat_face:

PhilippGackstatter (approved):
Looks good to me. Perhaps the consensus team can comment on the outstanding TODOs in the event hook logic.

pkg/protocol/engine/mempool/transaction.go
```diff
@@ -7,9 +7,13 @@ import (
...
+	// MustID works like ID but panics if the SignedTransactionID can't be computed.
+	MustID() iotago.SignedTransactionID
 }
 
 type Transaction interface {
 	// ID returns the identifier of the Transaction.
 	ID() (iotago.TransactionID, error)
+	// MustID works like ID but panics if the TransactionID can't be computed.
+	MustID() iotago.TransactionID
 }
```
jkrvivian, line 19: It seems more reasonable to put these interfaces in iota.go.  :thinking: 
Did Transaction and SignedTransaction structure in iota.go implement these interfaces? If yes, then we also need to implement MustID() for each.

pkg/retainer/txretainer/testsuite_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/iotaledger/iota-core/pkg/storage/clonablesql"
+	iotago "github.com/iotaledger/iota.go/v4"
+)
+
+type txRetainerTestsuite struct {
+	T          *testing.T
+	database   *clonablesql.ClonableSQLiteDatabase
+	TxRetainer *txretainer.TransactionRetainer
+
+	latestCommitedSlot iotago.SlotIndex
```
jkrvivian:
```suggestion
	latestCommittedSlot iotago.SlotIndex
```

jkrvivian (changes requested):

pkg/protocol/engine/mempool/transaction.go
muXxer: The `MustID()` was added to iota.go in https://github.com/iotaledger/iota.go/pull/696 so this is fine.
Those interfaces are only used in the mempool though, so I don't think we should move them to iota.go itself.

pkg/retainer/txretainer/tx_retainer_cache.go
muXxer: good catch. Before I used the same logic in the database, and that's why I confused these things. It is not needed here :+1:

pkg/retainer/txretainer/tx_retainer.go
muXxer: Oh, this was wrong actually, I fixed it.

jkrvivian (approved):

pkg/retainer/txretainer/tx_retainer_cache.go
```diff
@@ -0,0 +1,216 @@
...
+
+// DeleteAndReturnTxMetadataChangesBySlot deletes all the transaction metadata changes of a certain slot and returns them.
+func (c *transactionRetainerCache) DeleteAndReturnTxMetadataChangesBySlot(targetSlot iotago.SlotIndex) map[iotago.TransactionID]*TransactionMetadata {
+	c.accessMutex.Lock()
+	defer c.accessMutex.Unlock()
+
+	var slots []iotago.SlotIndex
+
+	// iterate over all slots, to ensure that all slots below the given slot are deleted from the cache as well
+	c.uncommittedTxIDsBySlot.ForEachKey(func(slot iotago.SlotIndex) bool {
```
cyberphysic4l, line 180: Is there some way to iterate in order perhaps so we can continue when we reach the targetslot?

pkg/retainer/txretainer/tx_retainer.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+
+	if err := r.txRetainerCache.UpdateTxMetadata(txMeta); err != nil {
+		r.errorHandler(err)
+	}
+}
+
+// TransactionMetadata returns the metadata of a transaction.
+func (r *TransactionRetainer) TransactionMetadata(txID iotago.TransactionID) (*api.TransactionMetadataResponse, error) {
+	// first check the cache, if there is an entry and it is accepted, we can return it without checking the database
```
cyberphysic4l, line 327: Could you add a comment here about why it needs to be accepted?

pkg/retainer/txretainer/tx_retainer_cache.go
```diff
@@ -0,0 +1,216 @@
...
+
+	txMetadataUpdateFunc txMetadataUpdateFunc
+
+	uncommittedTxMetadataChangesByID *shrinkingmap.ShrinkingMap[iotago.TransactionID, *TransactionMetadata]
+	uncommittedTxIDsBySlot           *shrinkingmap.ShrinkingMap[iotago.SlotIndex, map[iotago.TransactionID]struct{}]
+}
+
+// defaultTxMetadataUpdateFunc is the default update function for transaction metadata in the transaction retainer cache.
+// HINT: do not modify the contents of the old transaction metadata in the update function, as it is used as a reference
+// to the earliest attachment slot in the cache.
```
cyberphysic4l, line 25: Type here in the name defaul instead of default

pkg/retainer/txretainer/tx_retainer_cache.go
```diff
@@ -0,0 +1,216 @@
...
+		return newTxMeta, true, nil
+	}
+
+	if oldTxMeta.Equal(newTxMeta) {
+		// the transaction metadata has not changed, return the old one
+		return oldTxMeta, false, nil
+	}
+
+	if oldTxMeta.ValidSignature && !newTxMeta.ValidSignature {
+		// do not update the entry in the cache if the signature was valid before and is invalid now
```
cyberphysic4l: In what situation would this happen?

cyberphysic4l: And the next case where the signature goes from invalid to valid?

cyberphysic4l (approved):

pkg/retainer/txretainer/tx_retainer_cache.go
muXxer: Maps are unsorted unfortunately, but it shouldn't be much entries anyway.

pkg/retainer/txretainer/tx_retainer.go
muXxer: The more detailed explanation is some lines below. I don't think it makes sense to be more verbose here.

pkg/retainer/txretainer/tx_retainer_cache.go
muXxer: For a different attachment.

pkg/retainer/txretainer/tx_retainer.go
```diff
@@ -0,0 +1,387 @@
...
+
+// NewProvider creates a new TransactionRetainer provider.
+func NewProvider(opts ...options.Option[TransactionRetainer]) module.Provider[*engine.Engine, retainer.TransactionRetainer] {
+	return module.Provide(func(e *engine.Engine) retainer.TransactionRetainer {
+		r := New(e.Workers.CreateGroup("TransactionRetainer"),
+			e.Storage.TransactionRetainerDatabaseExecFunc(),
+			func() iotago.SlotIndex {
+				// use settings in case SyncManager is not constructed yet.
+				if e.SyncManager == nil {
+					return e.Storage.Settings().LatestCommitment().Slot()
```
alexsporn: why is this needed? just change the order in the engine providing part

pkg/retainer/txretainer/tx_retainer_cache.go
```diff
@@ -0,0 +1,217 @@
...
+)
+
+// transactionRetainerCache is a cache for transaction metadata changes that are not yet committed to the database.
+type transactionRetainerCache struct {
+	accessMutex sync.RWMutex
+
+	txMetadataUpdateFunc txMetadataUpdateFunc
+
+	uncommittedTxMetadataChangesByID *shrinkingmap.ShrinkingMap[iotago.TransactionID, *TransactionMetadata]
+	uncommittedTxIDsBySlot           *shrinkingmap.ShrinkingMap[iotago.SlotIndex, map[iotago.TransactionID]struct{}]
```
alexsporn, line 20: use `*memstorage.IndexedStorage[iotago.SlotIndex, iotago.TransactionID, struct{}]` instead

pkg/retainer/txretainer/tx_retainer.go
muXxer: I stole that part from the block retainer :D
I changed the order now and removed it in both retainers :+1:

pkg/retainer/txretainer/tx_retainer_cache.go
muXxer: Good idea, but I think it doesn't fit my use case exactly.

alexsporn (approved):

PR #770, opened 2024-02-23
Recreate mempool breaking with invalid signature

### LINKED ISSUE #765: Crash: there is a gap in the ledgerstate
All nodes on our network died at the same time with the panic:

```
panic: there is a gap in the ledgerstate 18074 vs 18073
```

[iota-core.log](https://github.com/iotaledger/iota-core/files/14359554/iota-core.log)

cyberphysic4l:
See trace level logs in failing test at pkg/tests/mempool_invalid_signatures_test.go

Same sequence of mempool events occur as note in https://github.com/iotaledger/iota-core/issues/765#issuecomment-1959674942

Mempool registers the signatures as invalid but still accepts the transaction, which somehow results in the transaction inputs being consumed, but no new outputs created. In the case of implicit account transition with invalid signatures, this results in a failure to commit the mana manager state because the consumed output can't be retrieved, and the node subsequently crashes on the next slot due to a gap in the ledger state. 

Mempool behaviour should be fixed, but engine should also panic when failing to commit rather than waiting until next commitment to realise there is a gap in ledger state.

### DIFF
--- a/pkg/testsuite/mock/wallet_transactions.go
+++ b/pkg/testsuite/mock/wallet_transactions.go
@@ -388,6 +388,7 @@ func (w *Wallet) CreateImplicitAccountAndBasicOutputFromInput(transactionName st
 
 func (w *Wallet) TransitionImplicitAccountToAccountOutput(transactionName string, inputNames []string, opts ...options.Option[builder.AccountOutputBuilder]) *iotago.SignedTransaction {
 	var implicitAccountOutput *utxoledger.Output
+	var baseTokenAmount iotago.BaseToken
 	inputs := make(utxoledger.Outputs, 0, len(inputNames))
 	for _, inputName := range inputNames {
 		input := w.Output(inputName)
@@ -399,13 +400,14 @@ func (w *Wallet) TransitionImplicitAccountToAccountOutput(transactionName string
 			implicitAccountOutput = input
 		}
 		inputs = append(inputs, input)
+		baseTokenAmount += input.BaseTokenAmount()
 	}
 	if implicitAccountOutput == nil {
 		panic("no implicit account output found")
 	}
 	implicitAccountID := iotago.AccountIDFromOutputID(implicitAccountOutput.OutputID())
 
-	accountOutput := options.Apply(builder.NewAccountOutputBuilder(w.Address(), MinIssuerAccountAmount(w.Node.Protocol.CommittedAPI().ProtocolParameters())).
+	accountOutput := options.Apply(builder.NewAccountOutputBuilder(w.Address(), baseTokenAmount).
 		AccountID(implicitAccountID),
 		opts).MustBuild()
 
--- a/pkg/testsuite/mock/wallet_transactions.go
+++ b/pkg/testsuite/mock/wallet_transactions.go
@@ -347,7 +347,7 @@ func (w *Wallet) CreateImplicitAccountFromInput(transactionName string, inputNam
 	return signedTransaction
 }
 
-// CreateImplicitAccountFromInput creates an implicit account output.
+// CreateImplicitAccountAndBasicOutputFromInput creates an implicit account output and a remainder basic output from a basic output.
 func (w *Wallet) CreateImplicitAccountAndBasicOutputFromInput(transactionName string, inputName string, recipientWallet *Wallet) *iotago.SignedTransaction {
 	input := w.Output(inputName)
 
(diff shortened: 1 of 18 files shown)


tools/docker-network/docker-compose.yml
```diff
@@ -297,11 +297,6 @@ services:
       --inx.address=node-1-validator:9029
       --faucet.bindAddress=inx-faucet:8091
       --faucet.rateLimit.enabled=false
-      --faucet.baseTokenAmount=1000000000
```
muXxer: Can you undo this change please? We need the higher values, not the default ones.

pkg/testsuite/testsuite.go
```diff
@@ -97,7 +97,7 @@ func NewTestSuite(testingT *testing.T, opts ...options.Option[TestSuite]) *TestS
 
 		optsWaitFor: durationFromEnvOrDefault(5*time.Second, "CI_UNIT_TESTS_WAIT_FOR"),
 		optsTick:    durationFromEnvOrDefault(2*time.Millisecond, "CI_UNIT_TESTS_TICK"),
-		optsLogger:  loggerFromEnvOrDefault("CI_UNIT_TESTS_NO_LOG", "CI_UNIT_TESTS_LOG_LEVEL"),
+		optsLogger:  log.NewLogger(log.WithLevel(log.LevelTrace)), //loggerFromEnvOrDefault("CI_UNIT_TESTS_NO_LOG", "CI_UNIT_TESTS_LOG_LEVEL"),
```
muXxer: Remove that before merging of the PR.

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -347,20 +347,66 @@ func (w *Wallet) CreateImplicitAccountFromInput(transactionName string, inputNam
...
+	var implicitAccountOutput *utxoledger.Output
+	inputs := make(utxoledger.Outputs, 0, len(inputNames))
+	for _, inputName := range inputNames {
+		input := w.Output(inputName)
+		basicOutput, isBasic := input.Output().(*iotago.BasicOutput)
+		if !isBasic {
+			panic(fmt.Sprintf("output with alias %s is not *iotago.BasicOutput", inputName))
+		}
+		if basicOutput.UnlockConditionSet().Address().Address.Type() == iotago.AddressImplicitAccountCreation {
+			implicitAccountOutput = input
```
muXxer, line 403: You should check if `implicitAccountOutput` was already set, and if so, you should panic because there was already another implicit account.

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -347,20 +347,66 @@ func (w *Wallet) CreateImplicitAccountFromInput(transactionName string, inputNam
 	return signedTransaction
 }
 
-func (w *Wallet) TransitionImplicitAccountToAccountOutput(transactionName string, inputName string, opts ...options.Option[builder.AccountOutputBuilder]) *iotago.SignedTransaction {
+// CreateImplicitAccountFromInput creates an implicit account output.
```
muXxer: Wrong comment.

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -347,20 +347,66 @@ func (w *Wallet) CreateImplicitAccountFromInput(transactionName string, inputNam
...
+			&iotago.AddressUnlockCondition{Address: recipientWallet.ImplicitAccountCreationAddress()},
+		},
+		Features: iotago.BasicOutputFeatures{},
 	}
-	if basicOutput.UnlockConditionSet().Address().Address.Type() != iotago.AddressImplicitAccountCreation {
-		panic(fmt.Sprintf("output with alias %s is not an implicit account", inputName))
+
+	remainderBasicOutput := &iotago.BasicOutput{
+		Amount: input.BaseTokenAmount() - MinIssuerAccountAmount(w.Node.Protocol.CommittedAPI().ProtocolParameters()),
+		Mana:   input.StoredMana() - AccountConversionManaCost(w.Node.Protocol.CommittedAPI().ProtocolParameters()),
```
muXxer: I wonder if it would be easier to read if we simply subtract the balance of the implicitAccountOutput instead if recalculating the Amount and ManaCost here.

muXxer (approved):

muXxer:
We should extend the test with issuance of a valid attachment after the behavior with the invalid attachment was fully checked in the tests. After that valid attachment, we should also issue another invalid attachment again.

PR #771, opened 2024-02-23
Fix bug with implicit account and basic output signatures

### LINKED ISSUE #765: Crash: there is a gap in the ledgerstate

cyberphysic4l:
To be merged along with companion PR https://github.com/iotaledger/iota.go/pull/697

This fixes a bug where a transaction containing both an implicit account and a basic output with the same underlying pubkey hash would register as invalid in the vm. A small change to address keys in the companion PR mentioned above means that a signature from the same pubkey will be valid for both the implicit account address and ed25519 address. 

The PR also includes a test for this where we transition an implicit account alongside a basic output with the same underlying public key. 

This will fix the problems for implicit account creation noted in this issue https://github.com/iotaledger/iota-core/issues/765

Certain transactions with invalid signatures may still crash the network due to mempool issues which will be addressed in https://github.com/iotaledger/iota-core/pull/770


muXxer:
Converted to a draft to discuss the changes in https://github.com/iotaledger/iota.go/pull/697, I would rather like to go with a solution like https://github.com/iotaledger/iota.go/pull/698.

muXxer (approved):

PR #772, opened 2024-02-23
Fix active flag in validators endpoint

muXxer:


piotrm50 (approved):

components/restapi/core/accounts.go
```diff
@@ -142,7 +142,8 @@ func validatorByAccountAddress(c echo.Context) (*api.ValidatorResponse, error) {
 
 	nextEpoch := deps.Protocol.APIForSlot(latestCommittedSlot).TimeProvider().EpochFromSlot(latestCommittedSlot) + 1
 
-	active, err := deps.Protocol.Engines.Main.Get().SybilProtection.IsCandidateActive(accountID, nextEpoch)
+	// HINT: the validators for an epoch are stored in the previous epoch
+	active, err := deps.Protocol.Engines.Main.Get().SybilProtection.IsCandidateActive(accountID, nextEpoch-1)
```
alexsporn: when why do a +1 before?

muXxer: It's the same as in the validators endpoint :trollface:

PR #775, opened 2024-02-26
Add event API docker tests

### LINKED ISSUE #724: Add tests for the Event API in the docker test framework
We should subscribe to all existing events and check if they are triggered correctly.

jkrvivian:
* Add assertion methods of event API topics to DockerTestFramework.
* Add methods: `CreateBlock` to create block only, and `SubmitBlock` to send it to nodes.
   Block creation and issuance are done in one method before, with separating creation and issuance, we can get the prepared block ID, etc to start listening on event API and then check if they are all received.
* Add `Wallet` to handle addresses, keys, and account information.
* Refactor methods of DockerTestFramework to use wallet properly.
* Move assertion methods from dockerframework.go to utils.go

Close #724


tools/docker-network/tests/wallet.go
```diff
@@ -0,0 +1,454 @@
...
+type Output struct {
+	// ID is the unique identifier of the output.
+	ID iotago.OutputID
+	// Output is the iotago Output.
+	Output iotago.Output
+	// Address is the address of the output.
+	Address iotago.Address
+	// AddressIndex is the index of the address in the keyManager.
+	AddressIndex uint32
+}
```
cyberphysic4l: We now have at least three Wallets, each with different structures for storing the output data. This one, the mock wallet and the evil-tools wallet. I really think we need to unify these wallets even though they work differently, we should make the mock wallet more generic to suit each.

tools/docker-network/tests/dockerframework.go
```diff
@@ -368,435 +352,304 @@ func (d *DockerTestFramework) StopIssueCandidacyPayload(nodes ...*Node) {
...
-
-	return updatedAccount
-}
-
-func (d *DockerTestFramework) CheckAccountStatus(ctx context.Context, blkID iotago.BlockID, txID iotago.TransactionID, creationOutputID iotago.OutputID, accountAddress *iotago.AccountAddress, checkIndexer ...bool) {
-	// request by blockID if provided, otherwise use txID
-	// we take the slot from the blockID in case the tx is created earlier than the block.
-	clt := d.Node("V1").Client
-	slot := blkID.Slot()
+	if congestionResp.BlockIssuanceCredits <= 500 {
```
cyberphysic4l: We should use RMC from the congestion response and the block to calculate this exactly.

cyberphysic4l: We should use the user the same flow as the block issuer in the mock package https://github.com/iotaledger/iota-core/blob/28a4e44108558839f4238ebbd2d1676cefdbf7bf/pkg/testsuite/mock/blockissuer.go#L214-L288

cyberphysic4l: FOr now, I would actually just remove the skipCheck option and the IncreaseBIC function as a way of increasing BIC because it is not how a user would do it. The additional required BIC should be allotted in the same transaction by providing an addiitonal input if it is required to increase BIC, not separate transaction for the sole purpose of increasing BIC.

tools/docker-network/tests/dockerframework.go
```diff
@@ -368,435 +352,304 @@ func (d *DockerTestFramework) StopIssueCandidacyPayload(nodes ...*Node) {
...
 	}
 
-	d.AwaitTransactionPayloadAccepted(ctx, blkID)
+	return issuerResp, congestionResp
+}
 
-	// wait for the account to be committed
-	d.AwaitCommitment(slot)
+// IncreaseBIC requests faucet funds then uses it to allots mana to an account.
+func (d *DockerTestFramework) IncreaseBIC(ctx context.Context, toId iotago.AccountID) {
```
cyberphysic4l: I think we should remove this function as it is not the way a user would/should increase their BIC. We should use the transaction builder utilities which allow additional Mana to be allotted as needed for the block requirements.

tools/docker-network/tests/dockerframework.go
```diff
@@ -368,435 +352,304 @@ func (d *DockerTestFramework) StopIssueCandidacyPayload(nodes ...*Node) {
...
+	signedTx := d.wallet.AllotManaFromAccount(fromId, toId, manaToAllot, fundsOutputID, issuerResp)
+	blkID := d.SubmitPayload(ctx, signedTx, from.ID, congestionResp, issuerResp)
 
-		require.EqualValues(d.Testing, account.OutputID, *outputID)
-		require.EqualValues(d.Testing, account.AccountOutput, output)
+	fmt.Println("Allot mana transaction sent, blkID:", blkID.ToHex(), ", txID:", lo.PanicOnErr(signedTx.Transaction.ID()).ToHex(), ", slot:", blkID.Slot())
 
-		return nil
-	})
+	d.AwaitTransactionPayloadAccepted(ctx, blkID)
```
cyberphysic4l, line 547: Should also await commitment here because that is when the allotment takes effect.

cyberphysic4l: Should we also assert that the account state was updated correctly in the node here?

cyberphysic4l (changes requested):
Just some things I would like to remove about allotting Mana as mentioned in comments. Other than that I think it all looks good. A few things will change of course when we refactor the testsuite wallet and merge it with docker wallet.

tools/docker-network/tests/wallet.go
jkrvivian: This will be resolved in another PR.

cyberphysic4l (approved):
Lgtm

Issue #776, opened 2024-02-26
Crash: "failed to calculate pool rewards" in 2 node setup on both nodes


HM4704:
The following call stack is identical on both nodes:

```
panic: failed to calculate pool rewards for account 0x907c02e9302e0f0571f10f885594e56d8c54ff0708ab7a39bc1b74d396b93b12: failed to calculate pool reward without fixed costs due to overflow for slot 8196: 2523412566627373056 * 9: integer under- or overflow

goroutine 218 [running]:
github.com/iotaledger/iota-core/pkg/protocol/sybilprotection/sybilprotectionv1/performance.(*Tracker).ApplyEpoch.func1({0x90, 0x7c, 0x2, 0xe9, 0x30, 0x2e, 0xf, 0x5, 0x71, 0xf1, ...}, ...)
	/home/pi/Projects/crypto/iota/iota-core/pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go:266 +0x488
github.com/iotaledger/hive.go/ds/shrinkingmap.(*ShrinkingMap[...]).ForEach(0x2ca0c40, 0x4015a555a0)
	/home/pi/go/pkg/mod/github.com/iotaledger/hive.go/ds@v0.0.0-20240216135101-261e99d9d84a/shrinkingmap/shrinkingmap.go:174 +0x258
github.com/iotaledger/iota-core/pkg/core/account.(*Accounts).ForEach(...)
	/home/pi/Projects/crypto/iota/iota-core/pkg/core/account/accounts.go:118
github.com/iotaledger/iota-core/pkg/protocol/sybilprotection/sybilprotectionv1/performance.(*Tracker).ApplyEpoch(0x40003e5c20, 0x0, 0x40147e4228)
	/home/pi/Projects/crypto/iota/iota-core/pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go:224 +0x270
github.com/iotaledger/iota-core/pkg/protocol/sybilprotection/sybilprotectionv1.(*SybilProtection).CommitSlot(0x400034ac60, 0x2004)
	/home/pi/Projects/crypto/iota/iota-core/pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go:189 +0x2a0
github.com/iotaledger/iota-core/pkg/protocol/engine/notarization/slotnotarization.(*Manager).createCommitment(0x40004281e0, 0x2004)
[...]

PR #786, opened 2024-02-29
change order of operations for a quick fix on the overflow

oliviasaa:


PhilippGackstatter (approved):
The panic that prompted this PR no longer occurs with this PR, so I'll approve (see https://github.com/iotaledger/iota-core/issues/776#issuecomment-1972344158 for how this was tested).

muXxer (approved):

PR #803, opened 2024-03-04
Add optional check when starting engine about correctness of last commitment and ledger state

### LINKED ISSUE #707: Add optional ledger state verification on startup
Add a flag to verify the integrity of the ledger state and compare the calculated merkle root with the one from the latest commitment

### LINKED ISSUE #708: Add option to validate snapshot after download
Validate the snapshot and compare the calculated merkle roots to the latest commitment.
This is related to #707

polinikita:
This PR is supposed to cover two issues: #707 and #708 

Upon starting an engine, and loading data into storage, it optionally triggers the check of the commitment and the ledger state. Once triggered, it then verifies the ledger state against the commitment and recomputes the latest commitment to ensure that the commitment fields correspond to the commitment ID. 

It requires accessing the roots of the latest commitment, so when creating a snapshot for a certain slot, we additionally store all the roots of the respective commitment (unless the genesis slot). When loading a snapshot, we additionally import the roots (unless the genesis slot).

One suite test is included to check the correctness of the whole logic.


pkg/protocol/protocol.go
```diff
@@ -143,6 +143,7 @@ func (p *Protocol) LatestAPI() iotago.API {
 }
 
 // initSubcomponents initializes the subcomponents of the protocol and returns a function that shuts them down.
+
```
piotrm50:
```suggestion
```

pkg/tests/engine_check_ledger_state_commitment_test.go
```diff
@@ -0,0 +1,91 @@
...
+	// Issue up to slot 10, committing slot 8.
+	{
+		ts.IssueBlocksAtSlots("", []iotago.SlotIndex{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}, 3, "Genesis", ts.Nodes(), true, false)
+		ts.AssertNodeState(ts.Nodes(),
+			testsuite.WithLatestFinalizedSlot(7),
+			testsuite.WithLatestCommitmentSlotIndex(8),
+			testsuite.WithEqualStoredCommitmentAtIndex(8),
+		)
+	}
+	// Create snapshot from node0 and start node 2 from it. Stop node 0
```
piotrm50: Is it like that though?

```suggestion
	// Create snapshot from node0 and start node 2 from it. During startup, it will correctly validate the ledger state for the last committed slot against the latest commitment.
	// Stop node 0.
```

pkg/protocol/engine/engine.go
```diff
@@ -127,8 +128,9 @@ func New(
 			LatestCommitment: reactive.NewVariable[*model.Commitment](),
 			Workers:          workers,
 
-			optsSnapshotPath:  "snapshot.bin",
-			optsSnapshotDepth: 5,
+			optsSnapshotPath:    "snapshot.bin",
+			optsSnapshotDepth:   5,
+			optsCheckCommitment: true,
```
piotrm50, line 135: Add an option so that this can be modified. Take snapshot path as a reference.

piotrm50 (changes requested):

pkg/protocol/engine/engine.go
polinikita: Made this

pkg/tests/engine_check_ledger_state_commitment_test.go
polinikita: yes

pkg/protocol/options.go
```diff
@@ -54,6 +54,9 @@ type Options struct {
 	// SnapshotPath is the path to the snapshot file that should be used to initialize the protocol.
 	SnapshotPath string
 
+	// CommitmentCheck is an opt flag that allows enginem upon startup, check correctness of commitment and ledger state
```
piotrm50:
```suggestion
	// CommitmentCheck is an opt flag that allows engines check correctness of commitment and ledger state upon startup.
```

piotrm50 (approved):

PR #807, opened 2024-03-05
Refactor REST api request handling

cyberphysic4l:
Moves api request handling to the requesthandler package so it can be used externally by the testsuite rather than copy and pasting the request handling functionality in the testsuite.

This also merges the block handler package into this request handler which handles all requests rather than just submitted blocks.


components/restapi/component.go
```diff
@@ -100,8 +100,8 @@ func provide(c *dig.Container) error {
 		Component.LogPanic(err.Error())
 	}
 
-	if err := c.Provide(func(deps dependencies) *blockhandler.BlockHandler {
-		return blockhandler.New(deps.Protocol)
+	if err := c.Provide(func(deps dependencies) *requesthandler.RequestHandler {
```
muXxer: That looks weird. Why do we use the `dependencies` of the component here, instead of an own `requestHandlerDeps struct` than only contains the `Protocol *protocolpkg.Protocol`.

components/restapi/core/accounts.go
```diff
@@ -32,16 +29,16 @@ func congestionByAccountAddress(c echo.Context) (*api.CongestionResponse, error)
 		workScores = append(workScores, workScore)
 	}
 
-	commitment := deps.Protocol.Engines.Main.Get().SyncManager.LatestCommitment()
+	commitment := deps.RequestHandler.GetLatestCommitment()
```
muXxer: Maybe we can avoid the call to the request handler, if we know that the commitmentID is not empty.

pkg/requesthandler/node.go
```diff
@@ -0,0 +1,19 @@
+package requesthandler
+
+import (
+	"github.com/iotaledger/iota-core/pkg/protocol/engine/clock"
+	"github.com/iotaledger/iota-core/pkg/protocol/engine/syncmanager"
+	iotago "github.com/iotaledger/iota.go/v4"
+)
+
+func (r *RequestHandler) APIProvider() *iotago.EpochBasedProvider {
```
muXxer: wasn't there an APIProvider interface as well?

components/restapi/core/transaction.go
```diff
@@ -5,57 +5,40 @@ import (
...
 	if err != nil {
 		return nil, ierrors.Wrapf(echo.ErrBadRequest, "failed to get block ID by transaction ID: %s", err)
 	}
 
-	block, exists := deps.Protocol.Engines.Main.Get().Block(blockID)
-	if !exists {
-		return nil, ierrors.Wrapf(echo.ErrNotFound, "block not found: %s", blockID.ToHex())
+	block, err := deps.RequestHandler.BlockByID(blockID)
+	if err != nil {
+		return nil, ierrors.Wrapf(echo.ErrNotFound, "block not found: %s", err)
```
muXxer: We already added the "block not found" message in `BlockByID`

components/restapi/core/component.go
```diff
@@ -40,7 +40,7 @@ type dependencies struct {
 	AppInfo          *app.Info
 	RestRouteManager *restapipkg.RestRouteManager
 	Protocol         *protocolpkg.Protocol
```
muXxer: Do we still need the Protocol dep?

components/restapi/core/accounts.go
```diff
@@ -164,143 +97,35 @@ func rewardsByOutputID(c echo.Context) (*api.ManaRewardsResponse, error) {
...
-	timeProvider := deps.Protocol.CommittedAPI().TimeProvider()
-
-	var slot iotago.SlotIndex
+	timeProvider := deps.RequestHandler.CommittedAPI().TimeProvider()
 
 	epoch, err := httpserver.ParseEpochQueryParam(c, api.ParameterEpoch)
 	if err != nil {
 		// by default we return current epoch
-		slot = timeProvider.SlotFromTime(time.Now())
+		slot := timeProvider.SlotFromTime(time.Now())
```
muXxer: Maybe we should add `CurrentSlot` and `CurrentEpoch` to the timeProvider?

muXxer (changes requested):

components/restapi/core/accounts.go
cyberphysic4l: moved the check for empty commitment ID to withing the getter.

pkg/requesthandler/node.go
cyberphysic4l: We do, but in this case we specifically need an epoch based provider as is supplied by the node settings. Just trying to keep things unchanged as much as possible for this PR.

components/restapi/core/accounts.go
cyberphysic4l: That would seem like a nice addition, but I wouldn't add it in this PR.

components/restapi/core/component.go
cyberphysic4l: We can do without it actually by just adding one more method for IsNodeSynced, so I have done that.

components/restapi/core/transaction.go
cyberphysic4l: ok, better to simply return the error without wrapping or to add something else as a wrapper?

muXxer: Since we already added the `echo.ErrNotFound` in `BlockByID`, we can simply return the error here without further wrapping.

pkg/requesthandler/node.go
cyberphysic4l: Now modified to use the APIProvider interface instead after refactor of info endpoint.

muXxer (approved):

PR #814, opened 2024-03-07
Implement `epochstore.CachedStore` and use it for committee store to avoid disk access when propagating weight

### LINKED ISSUE #713: Add cache to avoid loading committee from KVStore for every block
This is a hot path that reads data from the committee KVStore for each block - we should add a cache at some level. Maybe we should have a committee cache in SeatManager and clear it whenever SybilProtection.Reset() is called? Not necessarily as part of this PR, but then we should create an issue for that.

_Originally posted by @piotrm50 in https://github.com/iotaledger/iota-core/pull/673#discussion_r1461859333_
            

Might also be an issue in other places

jonastheis:
This PR implements an `epochstore.CachedStore` that provides a convenient in-memory wrapper around `epochstore.Store`. The cache works according to a simple FIFO queue, based on when an element was added.

This PR uses this new `epochstore.CachedStore` as an in-place replacement for `epochstore.Store` (now called `epochstore.BaseStore`) for the committee to fix #713 

Without the cache:
image

With the cache:
image

How to run this benchmark is described [here](https://github.com/iotaledger/iota-core/issues/713#issuecomment-1975987871)


pkg/storage/prunable/epochstore/store.go
```diff
@@ -136,40 +153,188 @@ func (s *Store[V]) Prune(epoch iotago.EpochIndex, defaultPruningDelay iotago.Epo
...
+
+		prunedEpochs = append(prunedEpochs, i)
 	}
 
-	return nil
+	return prunedEpochs, nil
 }
 
-func (s *Store[V]) RollbackEpochs(epoch iotago.EpochIndex) (lastPrunedEpoch iotago.EpochIndex, err error) {
+func (s *BaseStore[V]) RollbackEpochs(epoch iotago.EpochIndex) (lastPrunedEpoch iotago.EpochIndex, rollbackEpochs []iotago.EpochIndex, err error) {
```
alexsporn: Maybe rename the return param to `rolledbackEpochs`

pkg/storage/prunable/epochstore/store.go
```diff
@@ -136,40 +153,188 @@ func (s *Store[V]) Prune(epoch iotago.EpochIndex, defaultPruningDelay iotago.Epo
...
 }
 
-func (s *Store[V]) isTooOld(epoch iotago.EpochIndex) bool {
+func (s *BaseStore[V]) isTooOld(epoch iotago.EpochIndex) bool {
 	prunedEpoch, hasPruned := s.lastPrunedEpoch.Index()
 
 	return hasPruned && epoch <= prunedEpoch
 }
+
+type CachedStore[V any] struct {
```
alexsporn: Split interface and implementations into separate files for readability

pkg/storage/prunable/epochstore/store.go
```diff
@@ -136,40 +153,188 @@ func (s *Store[V]) Prune(epoch iotago.EpochIndex, defaultPruningDelay iotago.Epo
...
+		cache:      shrinkingmap.New[iotago.EpochIndex, V](),
+	}
+}
+
+func (c *CachedStore[V]) addToCache(epoch iotago.EpochIndex, value V) {
+	if !c.addedQueue.Offer(epoch) {
+		removedEpoch, _ := c.addedQueue.Poll()
+		c.cache.Delete(removedEpoch)
+
+		c.addedQueue.Offer(epoch)
```
alexsporn: Maybe we should add this as an atomic operation to the queue

pkg/storage/prunable/epochstore/store.go
```diff
@@ -136,40 +153,188 @@ func (s *Store[V]) Prune(epoch iotago.EpochIndex, defaultPruningDelay iotago.Epo
...
+
+func (c *CachedStore[V]) LastPrunedEpoch() (iotago.EpochIndex, bool) {
+	return c.store.LastPrunedEpoch()
+}
+
+func (c *CachedStore[V]) Load(epoch iotago.EpochIndex) (V, error) {
+	var zeroValue V
+
+	c.mutex.RLock()
+	if value, exists := c.cache.Get(epoch); exists {
```
alexsporn: This should be using Compute and loading it from store inside the func instead of having an extra mutex

jonastheis: I'd avoid using compute as this will use a write lock in this hot path. It is ugly but imo worth the performance we get form it. I added some comments and an additional check whether it has been added in the meantime to the write locked part.

alexsporn (approved):

PR #817, opened 2024-03-07
Test REST API with the Docker test framework

### LINKED ISSUE #722: Add tests for the REST API in the docker test framework
- We want to test every single endpoint in the rest API if it works as expected, especially regarding race conditions with other endpoint calls.

### LINKED ISSUE #792: Add status codes map to API pkg
Currently we always return 503 internal server error, even if we should return 404 for "block not found" etc.
We should check all endpoints and return the correct http status codes, as well as the correct RPC codes in the inx server methods.
Maybe it would make sense to add a map that resolves the error to the correct http error code.

daria305:
Corresponding iota.go [PR](https://github.com/iotaledger/iota.go/pull/707)

- add tests for REST API Fixes #722 
- correct http error codes, e.g. use not Found instead of internal server error. Fixes #792 
- add tests for error codes requests


components/restapi/core/component.go
```diff
@@ -152,7 +152,7 @@ func configure() error {
 		}
 
 		return responseByHeader(c, resp)
-	})
+	}, checkNodeSynced())
```
muXxer: Not all endpoints need the node to be synced.
Values that never change after they are created can safely returned by an unsync node.
For example "UTXO changes by slot" are no problem. Either the unsynced node has them or not. No need to be synced. But for transaction metadata or block metadata endpoints, the provided information could be outdated if the node is unsync.

pkg/requesthandler/accounts.go
```diff
@@ -99,7 +99,7 @@ func (r *RequestHandler) ValidatorByAccountAddress(accountAddress *iotago.Accoun
 func (r *RequestHandler) RewardsByOutputID(outputID iotago.OutputID, slot iotago.SlotIndex) (*api.ManaRewardsResponse, error) {
 	utxoOutput, err := r.protocol.Engines.Main.Get().Ledger.Output(outputID)
 	if err != nil {
-		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to get output %s from ledger: %s", outputID.ToHex(), err)
+		return nil, ierrors.Wrapf(echo.ErrNotFound, "output %s not found in the Ledger: %s", outputID.ToHex(), err)
```
muXxer: This would return `ErrNotFound` in every possible error case.
It should only be `ErrNotFound` if the database entry was not found, otherwise it should be `ErrInternalServerError`.

pkg/requesthandler/commitments.go
```diff
@@ -38,7 +38,7 @@ func (r *RequestHandler) GetCommitmentByID(commitmentID iotago.CommitmentID) (*m
 
 	commitment, err := r.protocol.Engines.Main.Get().Storage.Commitments().Load(commitmentID.Slot())
 	if err != nil {
-		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to load commitment, commitmentID: %s, slot: %d, error: %w", commitmentID, commitmentID.Slot(), err)
+		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to load commitment, commitmentID: %s, slot: %d, error: %v", commitmentID, commitmentID.Slot(), err)
```
muXxer: I don't understand the change from `%w` to `%v`.
With that change, the `ierrors` lib is unable to properly wrap and chain the errors.

Can you revert these changes?

In this special case you even could do something like this to be more clean:
```suggestion
		return nil, ierrors.Join(echo.ErrInternalServerError, ierrors.Wrapf(err, "failed to load commitment, commitmentID: %s, slot: %d", commitmentID, commitmentID.Slot()))
```

pkg/requesthandler/transaction.go
```diff
@@ -16,7 +16,7 @@ func (r *RequestHandler) BlockIDFromTransactionID(transactionID iotago.Transacti
 
 	output, spent, err := r.protocol.Engines.Main.Get().Ledger.OutputOrSpent(outputID)
 	if err != nil {
-		return iotago.EmptyBlockID, ierrors.Wrapf(echo.ErrInternalServerError, "failed to get output %s: %s", outputID.ToHex(), err)
+		return iotago.EmptyBlockID, ierrors.Wrapf(echo.ErrNotFound, "output %s of transaction %s not found: %s", transactionID.ToHex(), outputID.ToHex(), err)
```
muXxer: same here

pkg/requesthandler/utxo.go
```diff
@@ -12,7 +12,7 @@ import (
 func (r *RequestHandler) OutputFromOutputID(outputID iotago.OutputID) (*api.OutputResponse, error) {
 	output, err := r.protocol.Engines.Main.Get().Ledger.Output(outputID)
 	if err != nil {
-		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to get output %s from the Ledger: %s", outputID.ToHex(), err)
+		return nil, ierrors.Wrapf(echo.ErrNotFound, "output %s not found in the Ledger: %s", outputID.ToHex(), err)
```
muXxer: same here

pkg/requesthandler/utxo.go
```diff
@@ -24,7 +24,7 @@ func (r *RequestHandler) OutputFromOutputID(outputID iotago.OutputID) (*api.Outp
 func (r *RequestHandler) OutputMetadataFromOutputID(outputID iotago.OutputID) (*api.OutputMetadata, error) {
 	output, spent, err := r.protocol.Engines.Main.Get().Ledger.OutputOrSpent(outputID)
 	if err != nil {
-		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to get output %s from the Ledger: %s", outputID.ToHex(), err)
+		return nil, ierrors.Wrapf(echo.ErrNotFound, "output %s not found in the ledger: %s", outputID.ToHex(), err)
```
muXxer: same here

pkg/requesthandler/utxo.go
```diff
@@ -37,7 +37,7 @@ func (r *RequestHandler) OutputMetadataFromOutputID(outputID iotago.OutputID) (*
 func (r *RequestHandler) OutputWithMetadataFromOutputID(outputID iotago.OutputID) (*api.OutputWithMetadataResponse, error) {
 	output, spent, err := r.protocol.Engines.Main.Get().Ledger.OutputOrSpent(outputID)
 	if err != nil {
-		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to get output %s from the Ledger: %s", outputID.ToHex(), err)
+		return nil, ierrors.Wrapf(echo.ErrNotFound, "output %s not found in the Ledger: %s", outputID.ToHex(), err)
```
muXxer: same here

pkg/requesthandler/blocks.go
```diff
@@ -23,7 +23,7 @@ func (r *RequestHandler) BlockByID(blockID iotago.BlockID) (*iotago.Block, error
 func (r *RequestHandler) BlockMetadataByBlockID(blockID iotago.BlockID) (*api.BlockMetadataResponse, error) {
 	blockMetadata, err := r.protocol.Engines.Main.Get().BlockRetainer.BlockMetadata(blockID)
 	if err != nil {
-		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to get block metadata %s: %s", blockID.ToHex(), err)
+		return nil, ierrors.Wrapf(echo.ErrNotFound, "metadata not found for block %s: %s", blockID.ToHex(), err)
```
muXxer: same here

muXxer (changes requested):

components/restapi/core/component.go
daria305: I see, I left the ones added for OutputMetadata

tools/docker-network/tests/eventapi_test.go
```diff
@@ -413,11 +413,8 @@ func test_NFTTransactionBlocks(t *testing.T, d *DockerTestFramework) {
 			assertion()
 		}
 
-		// d.AssertTransactionMetadataByTransactionID(ctx, eventClt, outputId.TransactionID(), finish)
-		// d.AssertTransactionMetadataIncludedBlocks(ctx, eventClt, outputId.TransactionID(), finish)
```
jkrvivian, line 417: Can you add a TODO comment to this file that reminds us to test transactionMetadata topics?

tools/docker-network/tests/dockerframework.go
```diff
@@ -349,105 +359,136 @@ func (d *DockerTestFramework) StopIssueCandidacyPayload(nodes ...*Node) {
...
 	issuerResp, congestionResp := d.PrepareBlockIssuance(ctx, clt, issuer.Address)
 
-	return d.CreateBlock(ctx, &iotago.TaggedData{
+	return d.CreateBlock(&iotago.TaggedData{
 		Tag: tag,
-	}, issuerId, congestionResp, issuerResp)
+	}, issuerID, congestionResp, issuerResp)
+}
+
+func (d *DockerTestFramework) CreateValueBlock(issuerAccountID iotago.AccountID) (*iotago.Block, *iotago.SignedTransaction, *OutputData, iotago.Output) {
```
jkrvivian: Rename it to `CreateBasicOutputBlock`.

tools/docker-network/tests/dockerframework.go
```diff
@@ -349,105 +359,136 @@ func (d *DockerTestFramework) StopIssueCandidacyPayload(nodes ...*Node) {
...
+		AddInput(&builder.TxInput{
+			UnlockTarget: input.Address,
+			InputID:      input.ID,
+			Input:        input.Output,
+		}).
+		AddOutput(basicOutput).
+		SetCreationSlot(currentSlot).
+		AllotAllMana(currentSlot, issuerAccountID, 0).
+		Build()
+	require.NoError(d.wallet.Testing, err)
```
jkrvivian: Let's create a `CreateBasicOutputFromInput` method in wallet and move the tx building logic there, so it's aligned to other methods.

tools/docker-network/tests/coreapi_test.go
```diff
@@ -0,0 +1,542 @@
...
+				require.Nil(t, respBlock)
+			},
+		},
+		{
+			name: "Test_BlockMetadataByBlockID_Failure",
+			testFunc: func(t *testing.T, nodeAlias string) {
+				blockID := tpkg.RandBlockID()
+				resp, err := d.wallet.Clients[nodeAlias].BlockMetadataByBlockID(context.Background(), blockID)
+				require.Error(t, err)
+				require.True(t, isStatusCode(err, http.StatusInternalServerError))
```
jkrvivian: This should return NotFound?

tools/docker-network/tests/coreapi_test.go
```diff
@@ -0,0 +1,542 @@
...
+					require.NoError(t, err)
+					commitment, err := d.wallet.Clients[nodeAlias].CommitmentBySlot(context.Background(), infoRes.Status.LatestCommitmentID.Slot())
+					require.NoError(t, err)
+
+					resp, err = d.wallet.Clients[nodeAlias].Congestion(context.Background(), accountAddress, 0, commitment.MustID())
+					require.NoError(t, err)
+					require.NotNil(t, resp)
+					// later we check if all nodes have returned the same BIC value for this account
+					bicPerNoode[nodeAlias] = resp.BlockIssuanceCredits
+				})
```
jkrvivian, line 277: We can call assertBICs here, no?

components/restapi/core/accounts.go
```diff
@@ -27,7 +27,7 @@ func congestionByAccountAddress(c echo.Context) (*api.CongestionResponse, error)
 		workScores = append(workScores, workScore)
 	}
 
-	commitment, err := deps.RequestHandler.GetCommitmentByID(commitmentID)
+	commitment, err := deps.RequestHandler.GetCommitmentByID(queryCommitmentID)
```
jkrvivian: `queryCommitment` then? :grimacing:

components/restapi/core/accounts.go
```diff
@@ -43,6 +43,16 @@ func congestionByAccountAddress(c echo.Context) (*api.CongestionResponse, error)
...
+	maxCommittableAge := deps.RequestHandler.CommittedAPI().ProtocolParameters().MaxCommittableAge()
+	latestCommittedSlot := deps.RequestHandler.GetLatestCommitment().Slot()
+	if latestCommittedSlot >= maxCommittableAge && commitment.Slot()+maxCommittableAge < latestCommittedSlot && queryCommitmentID != iotago.EmptyCommitmentID {
+		return nil, ierrors.Wrapf(echo.ErrBadRequest, "invalid commitmentID, target slot index older than allowed (%d<%d)", commitment.Slot(), latestCommittedSlot-maxCommittableAge)
+	}
+
+	if commitment.Slot() > latestCommittedSlot && queryCommitmentID != iotago.EmptyCommitmentID {
+		return nil, ierrors.Wrapf(echo.ErrBadRequest, "invalid commitmentID, slot %d is not committed yet, latest committed slot: %d", commitment.Slot(), latestCommittedSlot)
+	}
+
```
jkrvivian: Move this check right after we parse queryCommitmentID, so we can return this error asap.

```go
if queryCommitmentID != iotago.EmptyCommitmentID {
    if commitment.Slot() > latestCommittedSlot {...}

    if latestCommittedSlot >= maxCommittableAge && commitment.Slot()+maxCommittableAge < latestCommittedSlot {...}

}
```

jkrvivian (changes requested):
Thanks for fixing all linter errors!

components/restapi/core/accounts.go
daria305: but that's `commitmentID`

tools/docker-network/tests/coreapi_test.go
daria305: This check if all nodes returned the same and iterates over `bicPerNoode map`, so it should be called afterwards

tools/docker-network/tests/coreapi_test.go
daria305: It was like that, but this still can return error for getting the store, so it's better to use InternalServerErr, see the above MuXxer comment

components/restapi/core/accounts.go
jkrvivian: I meant the returned commitment:
```suggestion
	queryCommitment, err := deps.RequestHandler.GetCommitmentByID(queryCommitmentID)
```

jkrvivian (approved):

muXxer (approved):

PR #825, opened 2024-03-11
Add cache for core API `validators/`

### LINKED ISSUE #752: Create API Cache package

jkrvivian:
Close #752 

* Request validators by epoch instead of slot
* Remove granularity in validators API
* Add option to set the maximum size of cache, default to 50MB
* In validators API:
   * Always request the node for validators if the `requested_Epoch == current_Epoch` and the cache result doesn't exist with key (`EpochIndex` + `SlotIndex`)
   * Keep validatorsResponse in cache `requested_Epoch < current_Epoch`
* Add unit test and docker test for validators API


pkg/requesthandler/accounts.go
```diff
@@ -32,23 +35,50 @@ func (r *RequestHandler) CongestionByAccountAddress(accountAddress *iotago.Accou
...
+		registeredValidators, err = r.protocol.Engines.Main.Get().SybilProtection.OrderedRegisteredCandidateValidatorsList(epochIndex)
+	} else {
+		registeredValidators, err = r.registeredValidatorsFromCache(epochIndex)
+	}
 	if err != nil {
-		return nil, ierrors.Join(echo.ErrInternalServerError, ierrors.Wrapf(err, "failed to get ordered registered validators list for epoch %d", latestEpoch))
+		return nil, ierrors.Wrapf(err, "failed to get registered validators for epoch %d", epochIndex)
 	}
 
 	page := registeredValidators[cursorIndex:lo.Min(cursorIndex+pageSize, uint32(len(registeredValidators)))]
```
alexsporn: we need some checks here since cursorIndex can be passed from the outside

pkg/requesthandler/accounts.go
```diff
@@ -60,7 +97,7 @@ func (r *RequestHandler) Validators(slotRange, cursorIndex, pageSize uint32) (*a
 	if int(cursorIndex+pageSize) > len(registeredValidators) {
 		resp.Cursor = ""
```
daria305: We could have single `if` here as Cursor is already empty
```
// not the last page
if int(cursorIndex+pageSize) <= len(registeredValidators) {
resp.Cursor = fmt.Sprintf("%d,%d", epochIndex, cursorIndex+pageSize)
}
```

pkg/requesthandler/accounts.go
```diff
@@ -32,26 +35,60 @@ func (r *RequestHandler) CongestionByAccountAddress(accountAddress *iotago.Accou
...
+		registeredValidators, err = r.registeredValidatorsFromCache(epochIndex, key)
+	} else {
+		registeredValidators, err = r.registeredValidatorsFromCache(epochIndex, epochIndex.MustBytes())
+	}
+	if err != nil {
+		return nil, ierrors.Wrapf(err, "failed to get registered validators for epoch %d", epochIndex)
+	}
+
+	pageEndIndex := lo.Min(cursorIndex+pageSize, uint32(len(registeredValidators)))
+	if cursorIndex >= pageEndIndex {
```
daria305: In simpler terms this could be 
`int(cursorIndex) >= len(registeredValidators)`
and we can check is `pageSize==0` at the beginning

pkg/requesthandler/requesthandler.go
```diff
@@ -1,26 +1,40 @@
...
+	})
 }
 
 // Shutdown shuts down the block issuer.
 func (r *RequestHandler) Shutdown() {
 	r.workerPool.Shutdown()
 	r.workerPool.ShutdownComplete.Wait()
 }
+
+func WithCacheMaxSizeOptions(size int) options.Option[RequestHandler] {
```
daria305, line 36: We should use it, or remove it

pkg/requesthandler/requesthandler.go
```diff
@@ -1,26 +1,40 @@
...
-func New(p *protocol.Protocol) *RequestHandler {
-	return &RequestHandler{
-		workerPool: p.Workers.CreatePool("BlockHandler"),
-		protocol:   p,
-	}
+func New(p *protocol.Protocol, opts ...options.Option[RequestHandler]) *RequestHandler {
+	return options.Apply(&RequestHandler{
+		workerPool:       p.Workers.CreatePool("BlockHandler"),
+		protocol:         p,
+		optsCacheMaxSize: 2 << 20, // 2MB
```
daria305: What about adding it to parameters next to `restapi.ParamsRestAPI.MaxPageSize`?

pkg/requesthandler/accounts.go
```diff
@@ -32,26 +35,60 @@ func (r *RequestHandler) CongestionByAccountAddress(accountAddress *iotago.Accou
...
-	//if !exists {
-	//	registeredValidators, err := r.protocol.Engines.Main.Get().SybilProtection.OrderedRegisteredCandidateValidatorsList(latestEpoch)
-	//	if err != nil {
-	//		return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to get ordered registered validators list for epoch %d : %s", latestEpoch, err)
-	//	}
-	//	r.protocol.Engines.Main.Get().BlockRetainer.RetainRegisteredValidatorsCache(slotRange, registeredValidators)
-	//}
-
-	registeredValidators, err := r.protocol.Engines.Main.Get().SybilProtection.OrderedRegisteredCandidateValidatorsList(latestEpoch)
+func (r *RequestHandler) registeredValidatorsFromCache(epochIndex iotago.EpochIndex, key []byte) ([]*api.ValidatorResponse, error) {
```
daria305: It could be nicer if the cache mechanics would be more separated, like this would be the method in `cache/cache.go`
Or at least wrap the second of creating the cache in some `Store` func
```go
if registeredValidatorsBytes == nil {
// get the ordered registered validators list from engine.
registeredValidators, err := r.protocol.Engines.Main.Get().SybilProtection.OrderedRegisteredCandidateValidatorsList(epochIndex)
if err != nil {...}
err := r.registeredValidatorsCache.Store(key, registeredValidators)
```

pkg/requesthandler/accounts.go
```diff
@@ -32,26 +35,60 @@ func (r *RequestHandler) CongestionByAccountAddress(accountAddress *iotago.Accou
...
+		registeredValidatorsBytes, err := apiForEpoch.Encode(registeredValidators, serix.WithTypeSettings(validatorResponsesTypeSettings))
+		if err != nil {
+			return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to encode ordered registered validators list for epoch %d : %s", epochIndex, err)
+		}
+		r.registeredValidatorsCache.Set(key, registeredValidatorsBytes)
+
+		return registeredValidators, nil
+	}
+
+	validatorResp := make([]*api.ValidatorResponse, 0)
```
daria305: Following the above suggestion we could do decoding in `Cache` logic and already get `[]*api.ValidatorResponse`

daria305 (changes requested):
Really nice test, if you could also remove 'Test_Validators' with its todos from coreapi_test.go after merging develop

daria305 (approved):

PR #831, opened 2024-03-13
Update rewards.go

oliviasaa:
change order of operations to prevent overflow


oliviasaa:
this is just a quick fix so it stops overflowing.
It decreases the precision a lot, so I have to work more on this later

PhilippGackstatter (approved):

PR #832, opened 2024-03-13
Fix wrong public route of committee endpoint in core API

jkrvivian:
* Fix the public route of committee endpoint so the query parameters are allowed.
* Add committee endpoint docker tests (with query parameters)


daria305 (approved):

muXxer:
Please run the gedoc script before merging!

PR #834, opened 2024-03-13
Fix: Orphanage

### LINKED ISSUE #784: Revisit transaction orphanage
> Blocks are marked as invalid if their transaction is orphaned and the orphaned slot is below the block's slotcommitment slot.
> 
> Is that correct though? If transaction is invalid we still book the block carrying that transaction. Why should it be otherwise if the transaction is orphaned when we try to book the block? Shouldn't we treat it the same as an invalid transaction in this case? 
> 
> A transaction is orphaned when: 
> * all valid attachments are evicted (then well, the block itself is evicted)
> * another spender of a transaction input is committed (the transaction is a losing doublespend). 
> * one of the inputs is marked as orphaned (which means that the transaction that created it is orphaned as well).
> 
> This also seems prone to race conditions, because if one node orphans the transaction, evicts it and then processes a new attachment of that transaction, it would not mark the block as invalid and would successfully attach the transaction to the mempool (as it already forgot about the orphaned version).
> However if a node evicts slightly later and sees the same block, it would not add a fresh attachments to the mempool (because it still has the orphaned version), and would mark the block (and it's whole future cone) as invalid. 
> 
>  

 _Originally posted by @piotrm50 in [#90](https://github.com/iotaledger/iota-core/issues/90#issuecomment-1968874212)_

hmoog:


pkg/protocol/engine/blocks/block.go
```diff
@@ -12,10 +13,15 @@ import (
 	"github.com/iotaledger/hive.go/stringify"
 	"github.com/iotaledger/iota-core/pkg/core/account"
 	"github.com/iotaledger/iota-core/pkg/model"
+	"github.com/iotaledger/iota-core/pkg/protocol/engine/mempool"
 	iotago "github.com/iotaledger/iota.go/v4"
 )
 
 type Block struct {
+	AllParentsBooked          reactive.Event
+	AllDependenciesReady      reactive.Event
```
piotrm50: Add a comment what these means exactly.

pkg/protocol/engine/booker/inmemorybooker/booker.go
```diff
@@ -94,26 +94,19 @@ func (b *Booker) Init(ledger ledger.Ledger, loadBlockFromStorage func(iotago.Blo
...
+	} else if signedTransactionMetadata == nil {
 		return ierrors.Errorf("transaction in block %s was not attached", block.ID())
 	}
+	block.SignedTransactionMetadata.Set(signedTransactionMetadata)
 
 	// Based on the assumption that we always fork and the UTXO and Tangle past cones are always fully known.
 	signedTransactionMetadata.OnSignaturesValid(func() {
 		transactionMetadata := signedTransactionMetadata.TransactionMetadata()
 
-		if orphanedSlot, isOrphaned := transactionMetadata.OrphanedSlot(); isOrphaned && orphanedSlot <= block.SlotCommitmentID().Slot() {
```
piotrm50: Why is this not the case anymore? The block would never become booked, but if we know that it's objectively invalid we might as well mark it as such. Wdyt?

pkg/tests/booker_test.go
```diff
@@ -212,6 +228,8 @@ func Test_DoubleSpend(t *testing.T) {
 }
 
 func Test_MultipleAttachments(t *testing.T) {
+	t.Skip("This test is currently disabled because it is not yet clear how to handle this case.")
```
piotrm50: We need to figure out those skipped tests.

piotrm50 (commented):
Check the fruit spams in the Docker network as well.

pkg/protocol/engine/mempool/state.go
```diff
@@ -15,6 +15,9 @@ type State interface {
 
 	// Whether the state is read only.
 	IsReadOnly() bool
+
+	// SlotBooked returns the slot index of the state if it is booked, otherwise -1.
```
jonastheis: `SlotIndex` is a `uint32`. How can it return `-1`?

piotrm50 (approved):

PR #835, opened 2024-03-14
Fix error wrapping order

### LINKED ISSUE #798: Replace errors with correct error level and Wrap/WithMessage
1. Check if `ierrors` was used everywhere (not `fmt.Errorf` etc).
2. Check correct usage of `Wrap` and `WithMessage` for all `ierrors`.
3. Check the whole iota-core code base for correct error handling (panic, log error etc...)

PhilippGackstatter:
Fix error wrapping order.

Fixes #798.

See iotaledger/iota.go#709 for the rationale of the changes.

## Quick Guide

1. If you call another function that can return an error, you typically want to annotate that error with some context so the reader of the error can roughly understand the call chain. This is where we use `ierrors.Wrap` or `ierrors.Wrapf`. This **prepends** a message. For instance:

```go
if err := vm.ChainSTVF(vmParams, iotago.ChainTransitionTypeDestroy, inputChain, nil); err != nil {
	return ierrors.Wrapf(err, "invalid destruction for %s %s", inputChain.Output.Type(), chainID)
}
```

which gives us an error like

```
failed to execute transaction: invalid destruction for AccountOutput 0x6a22cbe0e578b4d88440fd1ccced1c8903848a00f6cbcecd12bc3c4a2d144524: block issuer feature has not expired: feature expires in slot 4294967295 but the slot of the commitment input is 110
```

2. If you detect an error condition and return a named error (e.g. `iotago.ErrBlockIssuerNotExpired`), you typically want to _append_ a message with more details using `ierrors.WithMessage` or `ierrors.WithMessagef`. For instance:

```go
if blockIssuerFeat.ExpirySlot >= vmParams.WorkingSet.Commitment.Slot {
    return ierrors.WithMessagef(iotago.ErrBlockIssuerNotExpired, "feature expires in slot %d but the slot of the commitment input is %d", blockIssuerFeat.ExpirySlot, vmParams.WorkingSet.Commitment.Slot)
}
```

Generally, in my opinion, the goal of the error message is that the further you read into the message, the more details you will see. In the above example you see an increasing level of detail (`transaction level -> output level -> feature level`). This makes it easy to understand that the error was a failed transaction, that an account output in the transaction was the culprit and that within that account, the block issuer feature was not expired.

To summarize:
- If you propagate an error upwards, you should **add context** to the error, like what that component tried and failed to do. Use `ierrors.Wrap` for that.
- If you return a named error you should **add details** to the error, like the commitment and expiry slot above. Use `ierrors.WithMessage` for that.

## Join

As mentioned in the iota.go PR `ierrors.Join` is not what we typically want but we should make a decision if we use the style in iota.go, change how `ierrors.Join` works or add a new API like maybe `ierrors.WithError`. That's why I did not restyle the errors in iota-core because it might be unnecessary work and leaving the current `ierrors.Join` in place makes it easier to search and replace later than the equivalent using `WithMessagef` and `%w`.


pkg/network/p2p/manager.go
```diff
@@ -113,11 +113,11 @@ func (m *Manager) DialPeer(ctx context.Context, peer *network.Peer) error {
 	}
 
 	if m.NeighborExists(peer.ID) {
-		return ierrors.Wrapf(network.ErrDuplicatePeer, "peer %s already exists", peer.ID)
+		return ierrors.Wrapf(network.ErrDuplicatePeer, "peer %s already exists", peer.ID.String())
 	}
 
 	if !m.allowPeer(peer.ID) {
-		return ierrors.Wrapf(network.ErrMaxAutopeeringPeersReached, "peer %s is not allowed", peer.ID)
+		return ierrors.Wrapf(network.ErrMaxAutopeeringPeersReached, "peer %s is not allowed", peer.ID.String())
```
cyberphysic4l: Shouldn't these be WithMessagef() rather than Wrapf()?

pkg/network/p2p/manager.go
```diff
@@ -126,28 +126,28 @@ func (m *Manager) DialPeer(ctx context.Context, peer *network.Peer) error {
...
 
-		return ierrors.Wrapf(err, "failed to update peer %s", peer.ID)
+		return ierrors.Wrapf(err, "failed to update peer %s", peer.ID.String())
 	}
 
 	if err := m.addNeighbor(ctx, peer, ps); err != nil {
 		m.closeStream(stream)
 
-		return ierrors.Errorf("failed to add neighbor %s: %s", peer.ID, err)
+		return ierrors.Errorf("failed to add neighbor %s: %s", peer.ID.String(), err.Error())
```
cyberphysic4l: Why Errorf here rather than Wrapf?

pkg/protocol/engine/mempool/v1/mempool.go
```diff
@@ -352,28 +352,22 @@ func (m *MemPool[VoteRank]) storeTransaction(signedTransaction mempool.SignedTra
 
 	if m.lastCommittedSlot >= blockID.Slot() {
 		// block will be retained as invalid, we do not store tx failure as it was block's fault
-		return nil, false, false, ierrors.Errorf("blockID %d is older than last evicted slot %d", blockID.Slot(), m.lastCommittedSlot)
+		return nil, false, false, ierrors.Errorf("blockID %d is older than last evicted slot %d", blockID, m.lastCommittedSlot)
```
cyberphysic4l:
```suggestion
		return nil, false, false, ierrors.Errorf("blockID %s is older than last evicted slot %d", blockID, m.lastCommittedSlot)
```

pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
```diff
@@ -111,7 +111,7 @@ func (o *SybilProtection) TrackBlock(block *blocks.Block) {
 
 	accountData, exists, err := o.ledger.Account(block.ProtocolBlock().Header.IssuerID, block.SlotCommitmentID().Slot())
 	if err != nil {
-		o.errHandler(ierrors.Wrapf(err, "error while retrieving account from account %s in slot %d from accounts ledger", block.ProtocolBlock().Header.IssuerID, block.SlotCommitmentID().Slot()))
+		o.errHandler(ierrors.Wrapf(err, "error while retrieving account %s in slot %d from accounts ledger", block.ProtocolBlock().Header.IssuerID, block.SlotCommitmentID().Slot()))
```
cyberphysic4l:
```suggestion
		o.errHandler(ierrors.Wrapf(err, "error while retrieving data for account %s in slot %d from accounts ledger", block.ProtocolBlock().Header.IssuerID, block.SlotCommitmentID().Slot()))
```

pkg/requesthandler/blocks.go
```diff
@@ -15,7 +15,7 @@ import (
 func (r *RequestHandler) BlockByID(blockID iotago.BlockID) (*iotago.Block, error) {
 	block, exists := r.protocol.Engines.Main.Get().Block(blockID)
 	if !exists {
-		return nil, ierrors.Wrapf(echo.ErrNotFound, "block not found: %s", blockID.ToHex())
+		return nil, ierrors.WithMessagef(echo.ErrNotFound, "block %s not found", blockID.ToHex())
```
cyberphysic4l: I notice we sometimes use blockID.ToHex() and other times just blockID. Which is preferred?

cyberphysic4l (approved):
Look good. Just a few small suggestions.

pkg/requesthandler/blocks.go
PhilippGackstatter: `%s` with just `blockID` is preferred because block ID implements aliases, so we want to call the `String()` method on the ID (which is what `%s` will result in) and not `ToHex()`. That way we get the alias printed in error messages when we have an alias set (such as within tests). Thanks for pointing that out!

pkg/network/p2p/manager.go
PhilippGackstatter: Thanks. I just looked at too many errors and couldn't see the problems anymore I think :sweat_smile:.

PR #837, opened 2024-03-14
Unify Wallets: Part 1

### LINKED ISSUE #789: Unify wallets in testsuite and dockerframework
In order to avoid duplication of testing utilities in both the testsuite and the docker test framework, we want to modify the testsuite wallet so it can use the nodeclient to connect with a node rather than having direct access to a node via a struct field.

cyberphysic4l:
This is the first part of the process to unify the wallets of the testsuite and the docker test framework. This part deals with the testsuite side (the mock wallet) by removing the direct connection to the node within the testsuite. Instead, the mock wallet uses a client interface which matches the nodeclient used in the docker tests, but this mock client simply connects directly to the node. 

The ultimate goal is to allow a common wallet to be used by the testsuite and the docker wallet so transaction and block creation tools do not need to be replicated in both frameworks.


pkg/testsuite/mock/wallet_blocks.go
```diff
@@ -7,6 +7,8 @@ import (
 	"github.com/iotaledger/iota-core/pkg/protocol/engine/blocks"
 )
 
-func (w *Wallet) IssueBasicBlock(ctx context.Context, blockName string, opts ...options.Option[BasicBlockParams]) *blocks.Block {
-	return w.BlockIssuer.IssueBasicBlock(ctx, blockName, w.Node, opts...)
+func (w *Wallet) IssueBasicBlock(ctx context.Context, blockName string, opts ...options.Option[BasicBlockParams]) (*blocks.Block, error) {
+	block, err := w.BlockIssuer.IssueBasicBlock(ctx, blockName, opts...)
```
muXxer: Why not simply return here?

pkg/testsuite/mock/wallet.go
```diff
@@ -22,71 +54,83 @@ type Wallet struct {
...
-
-	return iotago.Ed25519PublicKeyHashBlockIssuerKeyFromPublicKey(hiveEd25519.PublicKey(pub))
+func (w *Wallet) SetDefaultClient(client Client) {
+	w.Client = client
+	w.BlockIssuer.Client = client
 }
 
-func (w *Wallet) SetDefaultNode(node *Node) {
-	w.Node = node
+func (w *Wallet) SetCurrentSlot(slot iotago.SlotIndex) {
```
muXxer, line 121: Wouldn't it make sense to set that in the mocked client instead?
This way you also don't have to do the type switch in the "CurrentSlot()".

pkg/requesthandler/transaction.go
```diff
@@ -38,3 +47,12 @@ func (r *RequestHandler) TransactionMetadataByID(txID iotago.TransactionID) (*ap
 
 	return txMetadata, nil
 }
+
+func (r *RequestHandler) BlockMetadataFromTransactionID(txID iotago.TransactionID) (*api.BlockMetadataResponse, error) {
```
muXxer: Move the function above the `TransactionMetadataFromTransactionID`, so we have everything block related first.

pkg/testsuite/blocks.go
```diff
@@ -34,21 +30,22 @@ func (t *TestSuite) AssertBlock(block *blocks.Block, node *mock.Node) *model.Blo
 	return loadedBlock
 }
 
-func (t *TestSuite) AssertBlocksExist(blocks []*blocks.Block, expectedExist bool, nodes ...*mock.Node) {
-	mustNodes(nodes)
+func (t *TestSuite) AssertBlocksExist(blocks []*blocks.Block, expectedExist bool, clients ...mock.Client) {
+	//mustNodes(nodes)
```
muXxer:
```suggestion
```

pkg/testsuite/mock/block_params.go
```diff
@@ -93,12 +93,6 @@ func WithIssuer(issuer wallet.Account) func(builder *BlockHeaderParams) {
 	}
 }
 
-func WithSkipReferenceValidation(skipReferenceValidation bool) func(builder *BlockHeaderParams) {
```
muXxer, line 96: So how do we set the `ReferenceValidation` parameter now?

pkg/testsuite/storage_settings.go
```diff
@@ -48,27 +52,24 @@ func (t *TestSuite) AssertLatestCommitment(commitment *iotago.Commitment, nodes
 	}
 }
 
-func (t *TestSuite) AssertCommitmentSlotIndexExists(slot iotago.SlotIndex, nodes ...*mock.Node) {
-	mustNodes(nodes)
+func (t *TestSuite) AssertCommitmentSlotIndexExists(slot iotago.SlotIndex, clients ...mock.Client) {
+	//mustNodes(nodes)
```
muXxer:
```suggestion
```

pkg/testsuite/mock/blockissuer.go
```diff
@@ -47,101 +43,109 @@ type BlockIssuer struct {
...
 func (i *BlockIssuer) BlockIssuerKeys() iotago.BlockIssuerKeys {
 	return iotago.NewBlockIssuerKeys(i.BlockIssuerKey())
 }
 
-// Shutdown shuts down the block issuer.
-func (i *BlockIssuer) Shutdown() {
-	i.workerPool.Shutdown()
-	i.workerPool.ShutdownComplete.Wait()
+func (i *BlockIssuer) Address() iotago.Address {
+	_, pub := i.keyManager.KeyPair()
```
muXxer: It looks weird that in the constructor you can pass an `addressIndex`, so the account address changes, but here it does not. Is this correct?

pkg/testsuite/mock/blockissuer.go
```diff
@@ -47,101 +43,109 @@ type BlockIssuer struct {
...
-		blockParams.BlockHeader.SlotCommitment, err = i.getAddressableCommitment(apiForBlock, *blockParams.BlockHeader.IssuingTime, node)
-		if err != nil && ierrors.Is(err, ErrBlockTooRecent) {
-			commitment, parentID, err := i.reviveChain(*blockParams.BlockHeader.IssuingTime, node)
+		commitment := blockIssuanceInfo.LatestCommitment
+		blockSlot := apiForBlock.TimeProvider().SlotFromTime(*blockParams.BlockHeader.IssuingTime)
+		if blockSlot > commitment.Slot+protoParams.MaxCommittableAge() {
+			var parentID iotago.BlockID
+			var err error
+			commitment, parentID, err = i.reviveChain(*blockParams.BlockHeader.IssuingTime, node)
+			fmt.Printf("revive chain commitment: slot %d, parentID %s\n", commitment.Slot, parentID)
```
muXxer: Do we need this print?

pkg/testsuite/mock/blockissuer.go
```diff
@@ -47,101 +43,109 @@ type BlockIssuer struct {
...
+			ID:           accountID,
+			AddressIndex: addressIndex,
+			Address:      accountAddress,
+		},
 	}, opts)
 }
 
 func (i *BlockIssuer) BlockIssuerKey() iotago.BlockIssuerKey {
-	return iotago.Ed25519PublicKeyHashBlockIssuerKeyFromPublicKey(hiveEd25519.PublicKey(i.PublicKey))
+	_, pub := i.keyManager.KeyPair()
```
muXxer: Is always only the first index used?

pkg/testsuite/mock/blockissuer.go
```diff
@@ -320,226 +294,129 @@ func (i *BlockIssuer) IssueActivity(ctx context.Context, wg *sync.WaitGroup, sta
...
 	}
 
 	return nil
 }
 
-func (i *BlockIssuer) IssueBlock(block *model.Block, node *Node) error {
-	if err := node.Protocol.IssueBlock(block); err != nil {
+func (i *BlockIssuer) SubmitBlock(ctx context.Context, block *model.Block) error {
+	defer i.mutex.Unlock()
+	i.mutex.Lock()
```
muXxer:
```suggestion
	i.mutex.Lock()
	defer i.mutex.Unlock()
```

pkg/testsuite/mock/blockissuer.go
```diff
@@ -320,226 +294,129 @@ func (i *BlockIssuer) IssueActivity(ctx context.Context, wg *sync.WaitGroup, sta
...
 }
 
-func WithRateSetterEnabled(enabled bool) options.Option[BlockIssuer] {
-	return func(i *BlockIssuer) {
-		i.optsRateSetterEnabled = enabled
+func (i *BlockIssuer) LatestBlockIssuanceResponse(context context.Context) *api.IssuanceBlockHeaderResponse {
+	defer i.mutex.Unlock()
+	i.mutex.Lock()
+
+	if i.blockIssuanceResponseUsed {
```
muXxer, line 419: I don't get why we change it in this function, maybe you can add some comments how the logic of this `blockIssuanceResponseUsed` is supposed to work?

muXxer (changes requested):

pkg/requesthandler/accounts.go
```diff
@@ -103,7 +113,18 @@ func (r *RequestHandler) ValidatorByAccountAddress(accountAddress *iotago.Accoun
 	}, nil
 }
 
-func (r *RequestHandler) RewardsByOutputID(outputID iotago.OutputID, slot iotago.SlotIndex) (*api.ManaRewardsResponse, error) {
+func (r *RequestHandler) RewardsByOutputID(outputID iotago.OutputID, optSlot ...iotago.SlotIndex) (*api.ManaRewardsResponse, error) {
+	var slot iotago.SlotIndex
+	if len(optSlot) > 0 {
+		genesisSlot := r.LatestAPI().ProtocolParameters().GenesisSlot()
+		if optSlot[0] < genesisSlot {
+			return nil, ierrors.Wrapf(echo.ErrBadRequest, "slot index (%d) before genesis slot (%d)", slot, genesisSlot)
```
jkrvivian:
```suggestion
			return nil, ierrors.WithMessagef(echo.ErrBadRequest, "slot index (%d) before genesis slot (%d)", slot, genesisSlot)
```

pkg/requesthandler/blocks.go
```diff
@@ -70,8 +60,26 @@ func (r *RequestHandler) BlockIssuance() (*api.IssuanceBlockHeaderResponse, erro
...
 			block, exists := r.protocol.Engines.Main.Get().Block(blockID)
 			if !exists {
+				// check if this is genesis block
+				if blockID == r.CommittedAPI().ProtocolParameters().GenesisBlockID() {
+					continue
+				}
+				// or root block
+				rootBlocks, err := r.protocol.Engines.Main.Get().Storage.RootBlocks(blockID.Slot())
+				if err != nil {
+					return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to get root blocks for slot %d: %s", blockID.Slot(), err)
```
jkrvivian:
```suggestion
					return nil, ierrors.WithMessagef(echo.ErrInternalServerError, "failed to get root blocks for slot %d: %s", blockID.Slot(), err)
```

pkg/requesthandler/blocks.go
```diff
@@ -70,8 +60,26 @@ func (r *RequestHandler) BlockIssuance() (*api.IssuanceBlockHeaderResponse, erro
...
+				}
+				isRootBlock, err := rootBlocks.Has(blockID)
+				if err != nil {
+					return nil, ierrors.Wrapf(echo.ErrInternalServerError, "failed to check if block %s is root block: %s", blockID.ToHex(), err)
```
jkrvivian:
```suggestion
					return nil, ierrors.WithMessagef(echo.ErrInternalServerError, "failed to check if block %s is root block: %s", blockID.ToHex(), err)
```

pkg/requesthandler/transaction.go
```diff
@@ -38,3 +47,12 @@ func (r *RequestHandler) TransactionMetadataByID(txID iotago.TransactionID) (*ap
...
+	blockID, err := r.BlockIDFromTransactionID(txID)
+	if err != nil {
+		return nil, ierrors.Wrapf(echo.ErrNotFound, "failed to get block ID from transaction ID: %v", err)
```
jkrvivian:
```suggestion
		return nil, ierrors.WithMessagef(echo.ErrNotFound, "failed to get block ID from transaction ID: %v", err)
```

pkg/testsuite/mock/blockissuer.go
```diff
@@ -47,101 +43,109 @@ type BlockIssuer struct {
 	Name      string
 	Validator bool
 
-	AccountID  iotago.AccountID
-	OutputID   iotago.OutputID
-	PublicKey  ed25519.PublicKey
-	privateKey ed25519.PrivateKey
+	keyManager *wallet.KeyManager
+	Client     Client
+	// LatestBlockIssuanceResp is the cached response from the latest query to the block issuance endpoint.
```
jkrvivian:
```suggestion
	// latestBlockIssuanceResp is the cached response from the latest query to the block issuance endpoint.
```

pkg/testsuite/mock/blockissuer.go
```diff
@@ -320,226 +294,129 @@ func (i *BlockIssuer) IssueActivity(ctx context.Context, wg *sync.WaitGroup, sta
...
 }
 
-func WithTipSelectionRetryInterval(interval time.Duration) options.Option[BlockIssuer] {
-	return func(i *BlockIssuer) {
-		i.optsTipSelectionRetryInterval = interval
-	}
-}
+func (i *BlockIssuer) GetNewBlockIssuanceResponse() *api.IssuanceBlockHeaderResponse {
+	defer i.mutex.Unlock()
+	i.mutex.Lock()
```
jkrvivian:
```suggestion
	i.mutex.Lock()
	defer i.mutex.Unlock()
```

pkg/testsuite/mock/blockissuer.go
```diff
@@ -320,226 +294,129 @@ func (i *BlockIssuer) IssueActivity(ctx context.Context, wg *sync.WaitGroup, sta
...
+
+	return i.latestBlockIssuanceResp
 }
 
-func WithRateSetterEnabled(enabled bool) options.Option[BlockIssuer] {
-	return func(i *BlockIssuer) {
-		i.optsRateSetterEnabled = enabled
+func (i *BlockIssuer) LatestBlockIssuanceResponse(context context.Context) *api.IssuanceBlockHeaderResponse {
+	defer i.mutex.Unlock()
+	i.mutex.Lock()
```
jkrvivian:
```suggestion
	i.mutex.Lock()
	defer i.mutex.Unlock()
```

pkg/testsuite/mock/node.go
```diff
@@ -336,10 +350,19 @@ func (n *Node) MainEngineSwitchedCount() int {
...
-func (n *Node) IssueValidationBlock(ctx context.Context, alias string, opts ...options.Option[ValidationBlockParams]) *blocks.Block {
+func (n *Node) IssueValidationBlock(ctx context.Context, alias string, opts ...options.Option[ValidationBlockParams]) (*blocks.Block, error) {
 	if n.Validator == nil {
 		panic("node is not a validator")
 	}
 
 	return n.Validator.IssueValidationBlock(ctx, alias, n, opts...)
 }
+
+func ClientsForNodes(nodes []*Node) []Client {
```
jkrvivian: There's a `ClientsForNodes` method in Testsuite, do we still need this function?

pkg/testsuite/testsuite.go
```diff
@@ -276,11 +275,34 @@ func (t *TestSuite) Nodes(names ...string) []*mock.Node {
 	return nodes
 }
 
+func (t *TestSuite) ClientsForNodes(names ...string) []mock.Client {
+	if len(names) == 0 {
+		t.mutex.RLock()
+		defer t.mutex.RUnlock()
+
+		clients := make([]mock.Client, 0, t.nodes.Size())
+		t.nodes.ForEach(func(_ string, node *mock.Node) bool {
```
jkrvivian: Is there a `t.Nodes()` method that returns all nodes?
If yes, we could replace the forEach here and remove the RLock above.

jkrvivian (changes requested):

muXxer (approved):

PR #839, opened 2024-03-15
Extend mempool invalid signatures test

### LINKED ISSUE #794: Extend the `TestMempoolInvalidSignature` test case with valid attachments
We should extend the test with issuance of a valid attachment after the behavior with the invalid attachment was fully checked in the tests. After that valid attachment, we should also issue another invalid attachment again.

PhilippGackstatter:
Simplify & extend the mempool invalid signatures test. The implicit account creation was a leftover from an earlier test that also tested the problem fixed by iotaledger/iota.go#698, which is no longer necessary now. Hence the test can be simplified to use Basic Outputs.

Two outstanding TODOs in the comments that need to be resolved.

Fixes #794.


muXxer:
Could you please add a similar check to the docker network tests, using the transaction retainer api to check the state?

muXxer (approved):

PR #840, opened 2024-03-15
Add BlockRetainer cache

### LINKED ISSUE #815: Add cache to BlockRetainer
We should add a cache to the block retainer similar to the one in the transaction retainer to avoid heavy disc IO.

daria305:
Fixes #815 
- blocks are stored under it's `blockID` slot index
 - up until slot commitment blocks metadata are kept in a shrinking map
 - on commitment of slot, all in-mem data is deleted up until slot and stored in the slot-based storage
 - unlike in tx retainer, block metadata can still change after the commitment - on confirmation
 - Reset func clears the shrinking maps and reset all the confirmed statuses to accepted for `slots=(lattestFinalisedSlot, commitedSlot]` as we are not sure if after resetting future slots confirmation will still hold


pkg/requesthandler/blocks.go
```diff
@@ -24,8 +24,8 @@ func (r *RequestHandler) BlockByID(blockID iotago.BlockID) (*iotago.Block, error
 func (r *RequestHandler) BlockMetadataByBlockID(blockID iotago.BlockID) (*api.BlockMetadataResponse, error) {
 	blockMetadata, err := r.protocol.Engines.Main.Get().BlockRetainer.BlockMetadata(blockID)
 	if err != nil {
-		if ierrors.Is(err, kvstore.ErrKeyNotFound) {
-			return nil, ierrors.WithMessagef(echo.ErrNotFound, "block %s not found", blockID)
+		if ierrors.Is(err, blockretainer.ErrEntryNotFound) {
+			return nil, ierrors.Wrapf(echo.ErrNotFound, "block not found: %s", blockID.ToHex())
```
muXxer: Please use `WithMessagef` to maintain the correct error order in this case. We go from less detailed to more detailed errors.

pkg/requesthandler/blocks.go
```diff
@@ -24,8 +24,8 @@ func (r *RequestHandler) BlockByID(blockID iotago.BlockID) (*iotago.Block, error
 func (r *RequestHandler) BlockMetadataByBlockID(blockID iotago.BlockID) (*api.BlockMetadataResponse, error) {
 	blockMetadata, err := r.protocol.Engines.Main.Get().BlockRetainer.BlockMetadata(blockID)
 	if err != nil {
-		if ierrors.Is(err, kvstore.ErrKeyNotFound) {
-			return nil, ierrors.WithMessagef(echo.ErrNotFound, "block %s not found", blockID)
+		if ierrors.Is(err, blockretainer.ErrEntryNotFound) {
```
muXxer: Why is there a specific NotFound error in the blockretainer? Either it's in the cache, or we have to look in the database and then we can also use the kvstore error here.

pkg/retainer/blockretainer/block_retainer.go
```diff
@@ -14,32 +18,35 @@ import (
...
 
 type (
-	StoreFunc               func(iotago.SlotIndex) (*slotstore.BlockMetadataStore, error)
-	LatestCommittedSlotFunc func() iotago.SlotIndex
-	FinalizedSlotFunc       func() iotago.SlotIndex
+	StoreFunc         func(iotago.SlotIndex) (*slotstore.BlockMetadataStore, error)
+	FinalizedSlotFunc func() iotago.SlotIndex
 )
 
+var ErrEntryNotFound = ierrors.New("block metadatanot found")
```
muXxer:
```suggestion
var ErrEntryNotFound = ierrors.New("block metadata not found")
```

pkg/retainer/blockretainer/block_retainer.go
```diff
@@ -99,51 +110,72 @@ func NewProvider() module.Provider[*engine.Engine, retainer.BlockRetainer] {
...
 func (r *BlockRetainer) blockState(blockID iotago.BlockID) (api.BlockState, error) {
-	blockMetadata, err := r.getBlockMetadata(blockID)
-	if err != nil {
-		return api.BlockStateUnknown, err
+	var state api.BlockState
+	state, found := r.cache.blockMetadataByID(blockID)
+	if !found {
+		// block is not committed yet, should be in cache
+		if blockID.Slot() > r.latestCommittedSlot {
+			return api.BlockStateUnknown, ErrEntryNotFound
```
muXxer: you can also return the `kvstore.ErrKeyNotFound` here instead. I think this is a valid simplification.

pkg/retainer/blockretainer/block_retainer.go
```diff
@@ -99,51 +110,72 @@ func NewProvider() module.Provider[*engine.Engine, retainer.BlockRetainer] {
...
+		}
+
+		blockMetadata, err := r.getBlockMetadata(blockID)
+		if err != nil {
+			// use custm error to align with the cache
+			if ierrors.Is(err, kvstore.ErrKeyNotFound) {
+				err = ierrors.Join(err, ErrEntryNotFound)
+			}
+
```
muXxer:
```suggestion
```

pkg/retainer/blockretainer/block_retainer.go
```diff
@@ -99,51 +110,72 @@ func NewProvider() module.Provider[*engine.Engine, retainer.BlockRetainer] {
...
+		if err != nil {
+			r.errorHandler(ierrors.Wrapf(err, "could not get retainer store for slot %d", slot))
+			continue
+		}
 
-	data, err := store.BlockMetadata(blockID)
-	if err != nil {
-		return nil, ierrors.Wrapf(err, "block %s not found", blockID.String())
+		// reset all block confirmations
+		store.ResetConfirmations()
```
muXxer: I'm not sure we need this magic at all. Let's discuss this in a short huddle with @piotrm50

pkg/retainer/blockretainer/block_retainer.go
```diff
@@ -152,7 +184,21 @@ func (r *BlockRetainer) blockState(blockID iotago.BlockID) (api.BlockState, erro
...
 	}
 
-	return blockMetadata.State, nil
+	return state, nil
+}
+
+func (r *BlockRetainer) getBlockMetadata(blockID iotago.BlockID) (*slotstore.BlockMetadata, error) {
+	store, err := r.store(blockID.Slot())
+	if err != nil {
+		return nil, err
```
muXxer:
```suggestion
		return nil, ierrors.Wrapf(err, "could not get retainer store for slot %d", blockID.Slot())
```

pkg/retainer/blockretainer/block_retainer.go
```diff
@@ -167,37 +213,79 @@ func (r *BlockRetainer) OnBlockBooked(block *blocks.Block) error {
...
+	defer r.Unlock()
+
+	slots := make(map[iotago.SlotIndex]map[iotago.BlockID]api.BlockState)
+	r.cache.uncommittedBlockMetadataChanges.ForEach(func(cacheSlot iotago.SlotIndex, blocks map[iotago.BlockID]api.BlockState) bool {
+		if cacheSlot <= committedSlot {
+			slots[committedSlot] = blocks
+		}
+
+		return true
+	})
```
muXxer: Why do we need this `slots` variable? Can't we combine the following logic directly with this `ForEach` loop?

pkg/retainer/blockretainer/block_retainer_cache.go
```diff
@@ -0,0 +1,49 @@
...
+		if found {
+			return blockMetadata, true
+		}
+	}
+
+	return api.BlockStateUnknown, false
+}
+
+func (c *cache) setBlockMetadata(blockID iotago.BlockID, state api.BlockState) {
+	blocks, exists := c.uncommittedBlockMetadataChanges.Get(blockID.Slot())
```
muXxer: Use `GetOrCreate`

pkg/retainer/blockretainer/block_retainer_cache.go
```diff
@@ -0,0 +1,49 @@
...
+	prevState, ok := blocks[blockID]
+	if ok && state == api.BlockStateDropped {
+		if prevState == api.BlockStateAccepted || state == api.BlockStateConfirmed {
+			// do not overwrite acceptance with the local congestion dropped event
+			return
+		}
+	}
+
+	blocks[blockID] = state
+	c.uncommittedBlockMetadataChanges.Set(blockID.Slot(), blocks)
```
muXxer: I think you might not need the `Set` if you use `GetOrCreate`, because you are pointing to the same map anyway.
Without it, this might be racy.

pkg/storage/prunable/slotstore/block_metadata.go
```diff
@@ -59,50 +59,31 @@ func NewBlockMetadataStore(slot iotago.SlotIndex, store kvstore.KVStore) *BlockM
...
-	blockMetadata, err := r.blockMetadataStore.Get(blockID)
-	if err != nil {
-		return err
-	}
-
-	blockMetadata.State = api.BlockStateConfirmed
+func (r *BlockMetadataStore) ResetConfirmations() {
+	err := r.blockMetadataStore.Iterate(kvstore.EmptyPrefix, func(key iotago.BlockID, value *BlockMetadata) bool {
+		// we can safely assum the status is accepted for the committed bocks
+		// but we are not sure if the confirmation would still hold if the proceeding slots were cleared out
```
muXxer:
```suggestion
		// we can safely assume the status is accepted for the committed blocks
		// but we are not sure if the confirmation would still hold if the proceeding slots were cleared out
```

pkg/storage/prunable/slotstore/block_metadata.go
```diff
@@ -59,50 +59,31 @@ func NewBlockMetadataStore(slot iotago.SlotIndex, store kvstore.KVStore) *BlockM
...
 
-func (r *BlockMetadataStore) StoreBlockConfirmed(blockID iotago.BlockID) error {
-	blockMetadata, err := r.blockMetadataStore.Get(blockID)
-	if err != nil {
-		return err
-	}
-
-	blockMetadata.State = api.BlockStateConfirmed
+func (r *BlockMetadataStore) ResetConfirmations() {
+	err := r.blockMetadataStore.Iterate(kvstore.EmptyPrefix, func(key iotago.BlockID, value *BlockMetadata) bool {
```
muXxer:
```suggestion
	if err := r.blockMetadataStore.Iterate(kvstore.EmptyPrefix, func(key iotago.BlockID, value *BlockMetadata) bool {
```

pkg/storage/prunable/slotstore/block_metadata.go
```diff
@@ -59,50 +59,31 @@ func NewBlockMetadataStore(slot iotago.SlotIndex, store kvstore.KVStore) *BlockM
...
-	if err != nil {
-		return err
-	}
-
-	blockMetadata.State = api.BlockStateDropped
-
-	return r.blockMetadataStore.Set(blockID, blockMetadata)
-}
+		return true
+	})
```
muXxer:
```suggestion
	}); err != nil {
```

muXxer (changes requested):

pkg/retainer/blockretainer/block_retainer_cache.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/iotaledger/hive.go/runtime/options"
+	iotago "github.com/iotaledger/iota.go/v4"
+	"github.com/iotaledger/iota.go/v4/api"
+)
+
+type cache struct {
+	uncommittedBlockMetadataChanges *shrinkingmap.ShrinkingMap[iotago.SlotIndex, map[iotago.BlockID]api.BlockState]
+}
+
+func newCache(opts ...options.Option[cache]) *cache {
```
jkrvivian: We don't need options for cache.

pkg/retainer/blockretainer/block_retainer_cache.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"github.com/iotaledger/hive.go/ds/shrinkingmap"
+	"github.com/iotaledger/hive.go/runtime/options"
+	iotago "github.com/iotaledger/iota.go/v4"
+	"github.com/iotaledger/iota.go/v4/api"
+)
+
+type cache struct {
+	uncommittedBlockMetadataChanges *shrinkingmap.ShrinkingMap[iotago.SlotIndex, map[iotago.BlockID]api.BlockState]
```
jkrvivian: `uncommittedBlockMetadata` would be clear enough :stuck_out_tongue:

pkg/retainer/blockretainer/block_retainer.go
```diff
@@ -99,51 +110,72 @@ func NewProvider() module.Provider[*engine.Engine, retainer.BlockRetainer] {
...
 		BlockState: blockStatus,
 	}, nil
 }
 
 func (r *BlockRetainer) blockState(blockID iotago.BlockID) (api.BlockState, error) {
-	blockMetadata, err := r.getBlockMetadata(blockID)
-	if err != nil {
-		return api.BlockStateUnknown, err
+	var state api.BlockState
+	state, found := r.cache.blockMetadataByID(blockID)
```
jkrvivian:
```suggestion
	state, found := r.cache.blockMetadataByID(blockID)
```

jkrvivian (changes requested):

muXxer (approved):

PR #841, opened 2024-03-15
Refactor and add more event API docker tests

### LINKED ISSUE #779: Events from `block-metadata` mqtt topics don't always relate to block metadata
Subscribing to mqtt topics `block-metadata/accepted` & `block-metadata/confirmed` sometimes return information of `pending` block states twice, instead of a normal behaviour `pending` -> `confirmed`. But fetching from the `client.getBlockMetadata(blockId)` method from the sdk returns the block state correctly.
double-pending_720
pending-confirmed

### LINKED ISSUE #797: Add "visualizer testcase" to the event API docker network test
Add a testcase for #779 after #751 was merged.

jkrvivian:
* Create EventAPIDockerTestFramework that wraps the DockerTestFramework with event API testing variables. 
* Test Transaction Metadata topics 
* Test Transaction with Tagged Data Payload by tag
* Fix #779 by exposing the accepted flag in BlockMetadata
* Test, if accepted/confirmed blocks received from event API, are indeed accepted/confirmed, close #797


pkg/retainer/blockretainer/block_retainer.go
```diff
@@ -124,11 +124,6 @@ func (r *BlockRetainer) BlockMetadata(blockID iotago.BlockID) (*api.BlockMetadat
 		return nil, ierrors.Wrapf(err, "block %s not found", blockID)
 	}
 
-	// we do not expose accepted flag
```
muXxer, line 127: do you know why we did that in the first place?
maybe @daria305 remembers?

tools/docker-network/tests/eventapiframework.go
```diff
@@ -0,0 +1,617 @@
...
+	}()
+}
+
+func (e *EventAPIDockerTestFramework) AssertAnchorOutput(ctx context.Context, eventClt *nodeclient.EventAPIClient, anchorId iotago.AnchorID) {
+	outputMetadataChan, subInfo := eventClt.OutputsWithMetadataByAnchorID(anchorId)
+	require.Nil(e.Testing, subInfo.Error())
+
+	go func() {
+		defer subInfo.Close()
+		e.assertOutputMetadataTopics(ctx, "AssertNFTOutput", outputMetadataChan, func(resp *api.OutputWithMetadataResponse) bool {
```
muXxer: wrong comment

tools/docker-network/tests/eventapiframework.go
```diff
@@ -0,0 +1,617 @@
...
+func (e *EventAPIDockerTestFramework) AwaitEventAPITopics(t *testing.T, cancleFunc context.CancelFunc, numOfTopics int) error {
+	counter := 0
+	timer := time.NewTimer(e.optsWaitFor)
+	defer timer.Stop()
+
+	for {
+		select {
+		case <-timer.C:
+			cancleFunc()
+			return ierrors.New("Timeout, did not receive signals from all  topics")
```
muXxer: two white spaces
```suggestion
			return ierrors.New("Timeout, did not receive signals from all topics")
```

tools/docker-network/tests/eventapi_test.go
```diff
@@ -415,19 +385,51 @@ func test_NFTTransactionBlocks(t *testing.T, d *DockerTestFramework) {
...
+
+		// wait until all topics starts listening
+		err := e.AwaitEventAPITopics(t, cancel, totalTopics)
 		require.NoError(t, err)
+
+		// issue blocks
+		e.SubmitDataBlockStream(account, 5*time.Minute)
+
+		// time's up, cancel the context
+		time.Sleep(5 * time.Minute)
```
muXxer: why do we wait here?
Shouldn't the `SubmitDataBlockStream` return earliest after the 5 minutes anyway?

muXxer (changes requested):

tools/docker-network/tests/eventapi_test.go
jkrvivian: Ah, this is for debugging, to let the network run a bit longer so I can check with core API. I'll remove it.

muXxer (approved):

PR #850, opened 2024-03-19
Breaking: Add network health endpoint and flag

### LINKED ISSUE #806: Node reports healthy but it is not
Network is at 102692 but `latestFinalizedSlot` is at 90115:
```json
{
  "name": "iota-core",
  "version": "1.0.0-alpha.8",
  "status": {
    "isHealthy": true,
    "acceptedTangleTime": "1709622328498843859",
    "relativeAcceptedTangleTime": "1709622329066738472",
    "confirmedTangleTime": "1709496705999999999",
    "relativeConfirmedTangleTime": "1709567311749196564",
    "latestCommitmentId": "0x1ad4e6230d8c0dbffefe067fa9f1a180c364569236ba6738bfa8eda89e6fc2721a910100",
    "latestFinalizedSlot": 90115,
    "latestAcceptedBlockSlot": 102685,
    "latestConfirmedBlockSlot": 90122,
    "pruningEpoch": 0
  },
```

but the node reports itself as healthy what imho it is not.

muXxer:
Companion PR to https://github.com/iotaledger/iota.go/pull/710.
We need to re-release all other plugins as well because this is a breaking change.


pkg/protocol/engine/syncmanager/trivialsyncmanager/syncmanager.go
```diff
@@ -126,26 +134,18 @@ func New(subModule module.Module, e *engine.Engine, latestCommitment *model.Comm
...
-	defer s.lastAcceptedBlockSlotLock.RUnlock()
-	defer s.lastConfirmedBlockSlotLock.RUnlock()
-	defer s.latestCommitmentLock.RUnlock()
-	defer s.latestFinalizedSlotLock.RUnlock()
-	defer s.lastPrunedEpochLock.RUnlock()
+	lastPrunedEpoch, hasPruned := s.LastPrunedEpoch()
 
 	return &syncmanager.SyncStatus{
-		NodeSynced:             s.IsNodeSynced(),
 		NodeBootstrapped:       s.IsBootstrapped(),
```
alexsporn: This changes the locking behaviour actually. Because you don't read all the values once you got all locks, you instead read (and potentially wait) for each single value.

alexsporn (approved):

PR #852, opened 2024-03-20
Fix chain engine traversal

### LINKED ISSUE #842: crash with SIGSEGV
We have a node that crashed with SIGSEGV.

I attached part of the log file, maybe it helps to figure out what happened or how to prevent it in future.
[log.log](https://github.com/iotaledger/iota-core/files/14617696/log.log)

muXxer:
This PR solves two bugs.

First there could have been an infinite loop because we used the wrong variable to traverse the parent chains. See first commit.

Second there could have been an edge case where an engine was found in the loop header because the assignment is executed after the loop condition, but the check for the parent chain inside the loop was nil, which caused the function to return nil instead of the newly found engine, which ultimately resulted in a nil pointer exception.

Should fix #842


alexsporn (approved):

hmoog (approved):
LGTM

PR #855, opened 2024-03-20
Make sure user knows that block was filtered out due to wallclock drift

### LINKED ISSUE #833: Create protocol event for wallclock time filter and hook in requesthandler
The wallclock time filter has been moved out of the engine and now it does not trigger any event when a block is in the future relative to the node's wallclock time. Instead the block is simply not processed any further. The result of this from the perspective of users issuing blocks via the API, this means their request will simply time out and they will get no information about why their block was dropped.

We should add an event to the protocol network events for BlockFiltered with a block failure reason and hook this event in the request handler to ensure the user is informed of why their block was dropped.

daria305:
Fixes #833 
WallClock drift check is moved from `Network.OnBlockReceived` further into the flow, to `ProcessResponse` where error is returned via `BlockFiltered` protocol event, so user will know about the error immediately.
ProcessResponse happens before engine, so that chain switching cannot be trigger by forcing commitment into the future


piotrm50 (changes requested):
This cannot be moved back there because we need that filter on the protocol level to filter out blocks with arbitrary future commitments.

alexsporn:
> This cannot be moved back there because we need that filter on the protocol level to filter out blocks with arbitrary future commitments.

@piotrm50 tbh, I don't like the current solution we have where the OnBlockReceived is overwritten. I find it very difficult to follow the flow of things how it is wired right now.

pkg/protocol/network.go
```diff
@@ -42,13 +37,6 @@ func newNetwork(protocol *Protocol, networkEndpoint network.Endpoint) *Network {
 // OnBlockReceived overwrites the OnBlockReceived method of the core protocol to filter out invalid blocks.
 func (n *Network) OnBlockReceived(callback func(block *model.Block, src peer.ID)) (unsubscribe func()) {
 	return n.Protocol.OnBlockReceived(func(block *model.Block, src peer.ID) {
-		// filter blocks from the future
```
jonastheis, line 45: We need to have the wall clock check before a block reaches the `Protocol`. You moved the check now to the engine, but that is already to late, as the protocol also (or maybe only the protocol and not the engine) needs to have a perception of `current time`. 
If this is not the case, then a malicious actor could send a block with an arbitrary timestamp in the future and thus also commit to something far in the future and claim a massive weight (he just needs to be part of the committee once). Consequently, all nodes would see this chain as the heaviest and switch to it. It is therefore crucial to have a sense of `current time` also in the protocol when we sync chains.

daria305: I moved it to the `ProcessResponse`, it's less hidden there and still should be early enough

piotrm50 (approved):

PR #856, opened 2024-03-20
Adapt .golangci.yml to latest version and fix linter issues

muXxer:


pkg/protocol/engine/accounts/accounts.go
```diff
@@ -44,8 +44,7 @@ func NewAccountData(id iotago.AccountID, opts ...options.Option[AccountData]) *A
 
 func (a *AccountData) AddBlockIssuerKeys(blockIssuerKeys ...iotago.BlockIssuerKey) {
 	for _, blockIssuerKey := range blockIssuerKeys {
-		k := blockIssuerKey
-		a.BlockIssuerKeys.Add(k)
+		a.BlockIssuerKeys.Add(blockIssuerKey)
```
piotrm50, line 47: Isn't the pointer we add going to change in the next iteration?

piotrm50 (approved):

pkg/protocol/engine/accounts/accounts.go
alexsporn: they changed that with go 1.22

alexsporn: https://go.dev/blog/loopvar-preview

alexsporn (approved):

PR #857, opened 2024-03-21
Hook to OnTransactionAttached in submitBlock

### LINKED ISSUE #826: Check transactions are always retained when needed
We need to ensure that all transaction processing steps are completed and the transaction is definitely retained prior to returning in the SubmitBlock API. In some cases we may take longer for the mempool to retain the transaction.

jkrvivian:
Close #826 

* Return success from SubmitBlock when the transaction is retained in the transaction retainer if the block contains one.
* Check the transaction metadata from the core API in event API docker tests.


cyberphysic4l:
I think it is best to create a `TransactionRetained` event, similar to the `BlockRetained` event, and hook to that instead of `TransactionAttached`.

Additionally, I think it would be best to rename the method `submitBlockAndAwaitEvent` to `submitBlockAndAwaitRetainer`, then remove the last argument, because we are only ever concerned with awaiting the retainers now. Also, `SubmitBlockAndAwaitBooking` should be renamed to `SubmitBlockAndAwaitRetainer`, and `SubmitBlockWithoutAwaitingBooking` to `SubmitBlockWithoutAwaitingRetainer`.

Finally, to deal with reattachments, if a block contains a transaction, we should first check if this transaction is already retained. If it is already retained we can treat the block like an empty block and not await the transaction retainer. If it is not yet retained then we await the `TransactionRetained` event.

cyberphysic4l (approved):

PR #888, opened 2024-03-27
Fix module lifecycles

muXxer:


pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go
```diff
@@ -46,54 +46,61 @@ type SybilProtection struct {
...
+		apiProvider:             engine,
+		optsSeatManagerProvider: topstakers.NewProvider(),
+	}, opts, func(s *SybilProtection) {
+		s.seatManager = s.optsSeatManagerProvider(engine)
+
+		s.ShutdownEvent().OnTrigger(func() {
+			s.StoppedEvent().Trigger()
 		})
+
+		s.ConstructedEvent().Trigger()
```
piotrm50: Not constructed here because the Ledger, performance tracker and errHandler is set later and here it's still `nil`

pkg/protocol/engine/consensus/slotgadget/totalweightslotgadget/gadget.go
```diff
@@ -67,12 +59,28 @@ func NewProvider(opts ...options.Option[Gadget]) module.Provider[*engine.Engine,
...
+		events:                        slotgadget.NewEvents(),
+		slotTrackers:                  shrinkingmap.New[iotago.SlotIndex, *slottracker.SlotTracker](),
+		optsSlotFinalizationThreshold: 0.67,
+		errorHandler:                  engine.ErrorHandler("slotgadget"),
+	}, opts, func(g *Gadget) {
+		g.ShutdownEvent().OnTrigger(func() {
+			g.StoppedEvent().Trigger()
 		})
+
+		g.ConstructedEvent().Trigger()
```
piotrm50: Some properties are nil here and are only set later.

piotrm50 (changes requested):

piotrm50 (approved):

PR #889, opened 2024-03-27
Unify wallet: Part 2

### LINKED ISSUE #789: Unify wallets in testsuite and dockerframework

cyberphysic4l:
Following on from https://github.com/iotaledger/iota-core/pull/837

This PR removes the DockerWallet and replaces it with the mock Wallet used in the testsuite, but with a nodeclient for the docker network and using a real clock instead of the testsuite time.

closes [#789](https://github.com/iotaledger/iota-core/issues/789)


pkg/testsuite/mock/client.go
```diff
@@ -244,6 +244,10 @@ func (c *TestSuiteClient) SubmitBlock(ctx context.Context, block *iotago.Block)
 	return c.Node.RequestHandler.SubmitBlockAndAwaitRetainer(ctx, block)
 }
 
+func (c *TestSuiteClient) TransactionByID(_ context.Context, _ iotago.TransactionID) (*iotago.Transaction, error) {
+	panic("not implemented")
```
daria305, line 250: is it in some separate PR?

pkg/testsuite/mock/wallet.go
```diff
@@ -44,6 +45,22 @@ type AccountData struct {
 	OutputID iotago.OutputID
 }
 
+func NewAccountDataFromAccountID(accountID iotago.AccountID, addressIndex ...uint32) *AccountData {
```
daria305: Why not simply `NewAccountData`

pkg/testsuite/mock/wallet.go
```diff
@@ -44,6 +45,22 @@ type AccountData struct {
...
+	accountAddress, ok := accountID.ToAddress().(*iotago.AccountAddress)
+	if !ok {
+		panic(ierrors.Errorf("accountID %s is not an account address", accountID.ToHex()))
+	}
+	if len(addressIndex) == 0 {
```
daria305: Defining array is weird. Maybe more straightforward, rename `addressIndex` to `optAddrressIndex`:
```
var addressIndex uint32
if len(optAddrressIndex) > 0 {
    addressIndex = optAddressIndex[0]
}
```

pkg/testsuite/mock/wallet.go
```diff
@@ -76,15 +93,19 @@ type Wallet struct {
...
-	IssuerAccount *AccountData
+	BlockIssuer *BlockIssuer
 
 	outputs      map[string]*OutputData
+	outputsByID  map[iotago.OutputID]*OutputData
 	transactions map[string]*iotago.Transaction
-	clock        WalletClock
+
+	accounts     map[iotago.AccountID]*AccountData
+	accountsLock sync.RWMutex
```
daria305: Why do we need mutex for accounts but not for outputs and txs?

pkg/testsuite/mock/wallet.go
```diff
@@ -176,6 +206,40 @@ func (w *Wallet) TransactionID(alias string) iotago.TransactionID {
...
+
+func (w *Wallet) Accounts(accountIDs ...iotago.AccountID) []*AccountData {
+	w.accountsLock.RLock()
+	defer w.accountsLock.RUnlock()
+
+	accounts := make([]*AccountData, 0)
+	if len(accountIDs) == 0 {
+		for _, acc := range w.accounts {
+			accounts = append(accounts, acc)
+		}
```
daria305:
```suggestion
		return lo.Values(w.accounts)
```

tools/docker-network/tests/dockerframework.go
```diff
@@ -66,16 +66,28 @@ func (n *Node) AccountAddress(t *testing.T) *iotago.AccountAddress {
 	return accAddress
 }
 
+type DockerWalletClock struct {
+	client mock.Client
+}
+
+func (c *DockerWalletClock) SetCurrentSlot(slot iotago.SlotIndex) {
```
daria305, line 73: I assume it should never be called, maybe panic here?

tools/docker-network/tests/dockerframework.go
```diff
@@ -369,192 +412,192 @@ func (d *DockerTestFramework) StopIssueCandidacyPayload(nodes ...*Node) {
...
-
-	return block, signedTx, input
+	return block, signedTx, fundsOutputData
 }
 
 // CreateDelegationBlockFromInput consumes the given basic output, then build a block of a transaction that includes a delegation output, in order to delegate the given validator.
-func (d *DockerTestFramework) CreateDelegationBlockFromInput(issuerID iotago.AccountID, accountAdddress *iotago.AccountAddress, inputID iotago.OutputID) (iotago.DelegationID, iotago.OutputID, *iotago.Block) {
-	issuer := d.wallet.Account(issuerID)
+func (d *DockerTestFramework) CreateDelegationBlockFromInput(issuerID iotago.AccountID, accountAdddress *iotago.AccountAddress, input *mock.OutputData) (iotago.DelegationID, iotago.OutputID, *iotago.Block) {
 	ctx := context.TODO()
```
daria305, line 453: Create and pass context directly in line 457 if not used anywhere else.
Also maybe use everywhere `context.Background()` instead of `context.TODO`, but I believe we never decided and mix them a lot

pkg/testsuite/testsuite.go
```diff
@@ -461,7 +461,7 @@ func (t *TestSuite) DefaultWallet() *mock.Wallet {
 }
 
 func (t *TestSuite) AddWallet(name string, node *mock.Node, accountID iotago.AccountID, keyManager ...*wallet.KeyManager) *mock.Wallet {
-	newWallet := mock.NewWallet(t.Testing, name, &mock.TestSuiteClient{Node: node}, keyManager...)
+	newWallet := mock.NewWallet(t.Testing, name, &mock.TestSuiteClient{Node: node}, &mock.TestSuiteWalletClock{}, keyManager...)
```
daria305: What about having `mock.NewTestSuiteClient(node)` and `mock.NewTestSuiteWalletClock()` just fora anicer look

pkg/testsuite/mock/client.go
cyberphysic4l: It is not, but it is not actually used in the testsuite, so not required to implement it at the moment.

pkg/testsuite/mock/wallet.go
cyberphysic4l: We should probably have that for outputs as well. I will add that. It was not required in testsuite, but probably required for docker tests.

pkg/testsuite/mock/client.go
PhilippGackstatter: Can you add a comment in the panic that it's not currently used, so a future reader knows it was intended rather than forgotten.

pkg/testsuite/mock/wallet.go
```diff
@@ -176,6 +237,38 @@ func (w *Wallet) TransactionID(alias string) iotago.TransactionID {
 	return w.Transaction(alias).MustID()
 }
 
+func (w *Wallet) Account(accountID iotago.AccountID) *AccountData {
+	w.mutex.RLock()
+	defer w.mutex.RUnlock()
+
+	acc, exists := w.accounts[accountID]
+	if !exists {
+		panic(ierrors.Errorf("account %s not registered in wallet", accountID.ToHex()))
```
PhilippGackstatter:
```suggestion
		panic(ierrors.Errorf("account %s not registered in wallet", accountID))
```

Nit: It's better to use `%s` with IDs that support aliasing (like block IDs, tx IDs, account IDs, ...) so the alias is printed rather than the hex string. Especially helpful in a testing context.

pkg/testsuite/mock/wallet.go
```diff
@@ -176,6 +237,38 @@ func (w *Wallet) TransactionID(alias string) iotago.TransactionID {
...
+
+	accounts := make([]*AccountData, 0)
+	if len(accountIDs) == 0 {
+		return lo.Values(w.accounts)
+	}
+
+	for _, id := range accountIDs {
+		acc, exists := w.accounts[id]
+		if !exists {
+			panic(ierrors.Errorf("account %s not registered in wallet", id.ToHex()))
```
PhilippGackstatter:
```suggestion
			panic(ierrors.Errorf("account %s not registered in wallet", id))
```

pkg/testsuite/mock/wallet_transactions.go
```diff
@@ -425,19 +431,83 @@ func (w *Wallet) TransitionImplicitAccountToAccountOutput(transactionName string
...
+		SetCreationSlot(currentSlot).
+		AddBlockIssuanceCreditInput(&iotago.BlockIssuanceCreditInput{AccountID: issuer.ID}).
+		AddCommitmentInput(&iotago.CommitmentInput{CommitmentID: w.GetNewBlockIssuanceResponse().LatestCommitment.MustID()}).
+		AddTaggedDataPayload(&iotago.TaggedData{Tag: []byte("foundry")}).
+		AllotAllMana(currentSlot, issuer.ID, 0).
+		Build()
+	require.NoError(w.Testing, err)
+
+	foundryOutputID := iotago.OutputIDFromTransactionIDAndIndex(signedTx.Transaction.MustID(), 1)
+	w.AddOutput("", &OutputData{
```
PhilippGackstatter: Not sure I understand why this is necessary, but can you give this a more descriptive name regardless?

tools/docker-network/tests/api_core.go
```diff
@@ -226,20 +227,21 @@ func (d *DockerTestFramework) prepareAssets(totalAssetsNum int) (coreAPIAssets,
 		d.AwaitTransactionPayloadAccepted(ctx, signedTx.Transaction.MustID())
 
 		// issue reattachment after the fisrt one is already included
```
PhilippGackstatter:
```suggestion
		// issue reattachment after the first one is already included
```

tools/docker-network/tests/rewards_test.go
```diff
@@ -208,59 +210,58 @@ func Test_DelayedClaimingRewards(t *testing.T) {
...
+func issueValidationBlockInBackground(wg *sync.WaitGroup, wallet *mock.Wallet, startSlot, endSlot iotago.SlotIndex, blocksPerSlot int) {
 	wg.Add(1)
 
 	go func() {
 		defer wg.Done()
-		fmt.Println("Issuing validation block for account", accountID, "in the background...")
-		defer fmt.Println("Issuing validation block for account", accountID, "in the background......done")
-		clt := d.wallet.DefaultClient()
+		fmt.Println("Issuing validation block for account", wallet.Name, "in the background...")
+		defer fmt.Println("Issuing validation block for account", wallet.Name, "in the background......done")
```
PhilippGackstatter: This should perhaps be modified to say wallet rather than account or print the actual account id.

PhilippGackstatter (approved):
Looks good to me. Minor comments.

There are a lot of occurences of creating a basic block where the alias is set to `""`. I think it would be good to set aliases there in case this test ever fails and we look at an error message containing that alias, then a string like "_tagged_data_block" is easier to find and debug than "". Just a thought. If you feel this is unnecessary, feel free to skip.

tools/docker-network/tests/rewards_test.go
```diff
@@ -179,27 +176,32 @@ func Test_DelayedClaimingRewards(t *testing.T) {
...
 		// delegate funds to V2
-		delegationOutputID, _ := d.DelegateToValidator(account.ID, d.Node("V2").AccountAddress(t))
-		d.AwaitCommitment(delegationOutputID.CreationSlot())
+		delegationOutputData := d.DelegateToValidator(delegatorWallet, d.Node("V2").AccountAddress(t))
+		d.AwaitCommitment(delegationOutputData.ID.CreationSlot())
 
 		// check if V2 received the delegator stake
 		v2Resp, err := clt.Validator(ctx, d.Node("V2").AccountAddress(t))
 		require.NoError(t, err)
+		fmt.Printf("Pool stake: %d, Validator stake: %d\n", v2Resp.PoolStake, v2Resp.ValidatorStake)
```
muXxer: Is this a debug print or is that still needed?

tools/docker-network/tests/rewards_test.go
```diff
@@ -179,27 +176,32 @@ func Test_DelayedClaimingRewards(t *testing.T) {
...
+		apiForSlot := clt.APIForSlot(currentSlot)
+		latestCommitmentSlot := delegatorWallet.GetNewBlockIssuanceResponse().LatestCommitment.Slot
+		delegationEndEpoch := getDelegationEndEpoch(apiForSlot, currentSlot, latestCommitmentSlot)
+		delegationOutputData = d.DelayedClaimingTransition(ctx, delegatorWallet, delegationOutputData)
+		d.AwaitCommitment(delegationOutputData.ID.CreationSlot())
 
 		// the delegated stake should be removed from the validator, so the pool stake should equal to the validator stake
 		v2Resp, err = clt.Validator(ctx, d.Node("V2").AccountAddress(t))
 		require.NoError(t, err)
+		fmt.Printf("Pool stake: %d, Validator stake: %d\n", v2Resp.PoolStake, v2Resp.ValidatorStake)
```
muXxer: Is this a debug print or is that still needed?

pkg/testsuite/mock/client.go
cyberphysic4l: Done for all unimplemented methods

PR #896, opened 2024-03-29
Improve Force Commitment logic

piotrm50:
Improve Force Commitment logic and apply other optimizations.
1. Only one ForceCommit process can be happening at a time.
2. Remove account mana monitoring.
3. Force commit only when the chain manager is not aware of a newer commitment.


muXxer (approved):

pkg/testsuite/mock/blockissuer_acceptance_loss.go
```diff
@@ -14,6 +14,13 @@ func (i *BlockIssuer) reviveChain(issuingTime time.Time, node *Node) (*iotago.Co
 
 	issuingSlot := apiForSlot.TimeProvider().SlotFromTime(issuingTime)
 
+	// If the chain manager is aware of a commitments on the main chain, then do not force commit.
```
jonastheis, line 17: Did you also implement this on the `inx-validator`?

jonastheis (approved):

PR #909, opened 2024-04-15
Feat: Orphanage before cleanup

hmoog:
This PR contains the changes of the orphange PR including the changes to the tip selection that automatically reference payloads in weak arrows if necessary.

We decided to not add additional code complexity to make "attacks easier".

If somebody wants to test a certain attack then he needs to implement his own tip selection strategy that references the otherwise unsolid transactions of deep nested spams, but then this complexity can live outside the core node implementation.

To not loose the already done work, we simply park the changes here.


hmoog:
These additional changes are no longer needed and are kept in this Branch only for achive reasons.

PR #943, opened 2024-04-26
Properly export and rollback accounts that were created after the target slot

alexsporn:
Also extend the AccountsLedger snapshot tests to verify the tree roots


pkg/protocol/engine/accounts/accountsledger/manager.go
```diff
@@ -314,7 +314,11 @@ func (m *Manager) Rollback(targetSlot iotago.SlotIndex) error {
 		}
 	}
 
-	return m.accountsTree.Commit()
+	if err := m.accountsTree.Commit(); err != nil {
+		return err
```
muXxer:
```suggestion
		return ierrors.Wrap(err, "unable to commit account tree")
```

piotrm50 (approved):

muXxer (approved):

PR #946, opened 2024-04-29
Fix scheduler bug and add panics

### LINKED ISSUE #934: OOM Crash - failed to store on BlockDropped in retainer
We have two nodes that crashed on out of memory.

It seems they started to log this error message:
```
Protocol.Engine0    	engine error (err=blockRetainer: failed to store on BlockDropped in retainer: cannot update block metadata for block BlockID(0xbc718142f4c3957f2e7484dec30b891a9edfc09b2d50c8faa8d753d09bb8dc12d4830000:33748) with state dropped as block is already committed)
```

About 50k times per hour.

Memory inflated at the time:
image

We have a log file when it started:
[faucet.h.iota2-alphanet_2024-04-24-09.log](https://github.com/iotaledger/iota-core/files/15106326/faucet.h.iota2-alphanet_2024-04-24-09.log)

Unfortunately it happened at night, so we have no memory profile of this node.

But we have profile of another node that started at the same time but "recovered" later on (while memory usage still is high)
image

[pprof.validator-2_20240425-075134_all.zip](https://github.com/iotaledger/iota-core/files/15106310/pprof.validator-2_20240425-075134_all.zip)

Maybe it shows something :see_no_evil:

### LINKED ISSUE #936: shutdown hanging, somehow prometheus related
(happened with iota-core alpha.18)

We monitor go metrics like heap and noticed that on some nodes prometheus is not accessible anymore, trying to `curl` the metrics leads to a timeout.

As expected, in this cases the nodes cannot be shutdown gracefully.

Node was healthy + network-healthy and other things like the `info` endpoint were working.

Here is a profile of iota-core while it is waiting for `Close p2p peer database, P2P, Protocol`.
[pprof.validator-11_20240425-125923_all.zip](https://github.com/iotaledger/iota-core/files/15110308/pprof.validator-11_20240425-125923_all.zip)

cyberphysic4l:
Possible fix for #936 
When issuer queues were removed, there was a possibility that the overall buffer size would not decrease by the correct amount do to use of PopFront() in the removeIssuer() function, where the issuer queue size was decreased, but not the overall buffer size. This is now fixed by not using PopFront in this place.

Also some panics have been added in places where we check for conditions that should never occur. Rather than return nil or some empty value, we should panic in such cases to detect the problem early.

### DIFF
--- a/pkg/protocol/engine/congestioncontrol/scheduler/drr/issuerqueue.go
+++ b/pkg/protocol/engine/congestioncontrol/scheduler/drr/issuerqueue.go
@@ -129,7 +129,7 @@ func (q *IssuerQueue) PopFront() *blocks.Block {
 
 	heapElement, isHeapElement := heap.Pop(&q.readyHeap).(*generalheap.HeapElement[timed.HeapKey, *blocks.Block])
 	if !isHeapElement {
-		panic("Unable to pop from a non-empty heap.")
+		panic("unable to pop from a non-empty heap.")
 	}
 	blk := heapElement.Value
 	q.size.Dec()
@@ -157,13 +157,13 @@ func (q *IssuerQueue) RemoveTail() *blocks.Block {
 		}
 	} else if heapTailIndex < 0 { // the heap is empty
 		// should never happen that the oldest submitted block does not exist and the heap is empty.
-		panic("Heap tail and oldest submitted block do not exist. Trying to remove tail of an empty queue?")
+		panic("heap tail and oldest submitted block do not exist. Trying to remove tail of an empty queue?")
 	}
 
 	// if the oldest ready block is older than the oldest non-ready block, drop it
 	heapElement, isHeapElement := heap.Remove(&q.readyHeap, heapTailIndex).(*generalheap.HeapElement[timed.HeapKey, *blocks.Block])
 	if !isHeapElement {
-		panic("Trying to remove a heap element that does not exist.")
+		panic("trying to remove a heap element that does not exist.")
 	}
 	blk := heapElement.Value
 	q.size.Dec()
--- a/pkg/protocol/engine/congestioncontrol/scheduler/drr/issuerqueue.go
+++ b/pkg/protocol/engine/congestioncontrol/scheduler/drr/issuerqueue.go
@@ -18,18 +18,18 @@ import (
 
 // IssuerQueue keeps the submitted blocks of an issuer.
 type IssuerQueue struct {
-	issuerID  iotago.AccountID
-	submitted *shrinkingmap.ShrinkingMap[iotago.BlockID, *blocks.Block]
-	inbox     generalheap.Heap[timed.HeapKey, *blocks.Block]
-	size      atomic.Int64
-	work      atomic.Int64
+	issuerID    iotago.AccountID
+	nonReadyMap *shrinkingmap.ShrinkingMap[iotago.BlockID, *blocks.Block]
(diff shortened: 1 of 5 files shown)


pkg/protocol/engine/congestioncontrol/scheduler/drr/issuerqueue.go
```diff
@@ -138,31 +138,32 @@ func (q *IssuerQueue) PopFront() *blocks.Block {
...
+		panic("Heap tail and oldest submitted block do not exist. Trying to remove tail of an empty queue?")
 	}
 
-	// if the tail exists and is older than the oldest submitted block, drop it
-	heapElement, isHeapElement := heap.Remove(&q.inbox, tail).(*generalheap.HeapElement[timed.HeapKey, *blocks.Block])
+	// if the oldest ready block is older than the oldest non-ready block, drop it
+	heapElement, isHeapElement := heap.Remove(&q.readyHeap, heapTailIndex).(*generalheap.HeapElement[timed.HeapKey, *blocks.Block])
 	if !isHeapElement {
-		return nil
+		panic("Trying to remove a heap element that does not exist.")
```
muXxer:
```suggestion
		panic("trying to remove a heap element that does not exist.")
```

pkg/protocol/engine/congestioncontrol/scheduler/drr/issuerqueue.go
```diff
@@ -138,31 +138,32 @@ func (q *IssuerQueue) PopFront() *blocks.Block {
...
-		return nil
+	heapTailIndex := q.heapTail()
+	// if heap tail (oldest ready block) does not exist or is newer than oldest non-ready block, unsubmit the oldest non-ready block
+	if oldestNonReadyBlock != nil && (heapTailIndex < 0 || q.readyHeap[heapTailIndex].Key.CompareTo(timed.HeapKey(oldestNonReadyBlock.IssuingTime())) > 0) {
+		if q.Unsubmit(oldestNonReadyBlock) {
+			return oldestNonReadyBlock
+		}
+	} else if heapTailIndex < 0 { // the heap is empty
+		// should never happen that the oldest submitted block does not exist and the heap is empty.
+		panic("Heap tail and oldest submitted block do not exist. Trying to remove tail of an empty queue?")
```
muXxer:
```suggestion
		panic("heap tail and oldest submitted block do not exist. Trying to remove tail of an empty queue?")
```

pkg/protocol/engine/congestioncontrol/scheduler/drr/issuerqueue.go
```diff
@@ -114,22 +114,22 @@ func (q *IssuerQueue) IDs() (ids []iotago.BlockID) {
...
-	if q.inbox.Len() == 0 {
+	if q.readyHeap.Len() == 0 {
 		return nil
 	}
 
-	heapElement, isHeapElement := heap.Pop(&q.inbox).(*generalheap.HeapElement[timed.HeapKey, *blocks.Block])
+	heapElement, isHeapElement := heap.Pop(&q.readyHeap).(*generalheap.HeapElement[timed.HeapKey, *blocks.Block])
 	if !isHeapElement {
-		return nil
+		panic("Unable to pop from a non-empty heap.")
```
muXxer:
```suggestion
		panic("unable to pop from a non-empty heap.")
```

muXxer (approved):

PR #947, opened 2024-04-29
Fix scheduler sizes and shutdown

### LINKED ISSUE #934: OOM Crash - failed to store on BlockDropped in retainer

### LINKED ISSUE #936: shutdown hanging, somehow prometheus related

muXxer:


pkg/protocol/engine/congestioncontrol/scheduler/drr/issuerqueue.go
```diff
@@ -18,19 +18,36 @@ import (
 
 // IssuerQueue keeps the submitted blocks of an issuer.
 type IssuerQueue struct {
-	issuerID    iotago.AccountID
-	nonReadyMap *shrinkingmap.ShrinkingMap[iotago.BlockID, *blocks.Block]
-	readyHeap   generalheap.Heap[timed.HeapKey, *blocks.Block]
-	size        atomic.Int64
-	work        atomic.Int64
+	issuerID        iotago.AccountID
```
cyberphysic4l, line 21: Why group them this way? It seems nices to group the readyHeap and the nonReadyMap, and then the size and work

pkg/protocol/engine/congestioncontrol/scheduler/drr/scheduler.go
```diff
@@ -408,31 +387,11 @@ func (s *Scheduler) selectBlockToScheduleWithLocking() {
 	s.bufferMutex.Lock()
 	defer s.bufferMutex.Unlock()
 
-	s.validatorBuffer.buffer.ForEach(func(_ iotago.AccountID, validatorQueue *ValidatorQueue) bool {
-		if s.selectValidationBlockWithoutLocking(validatorQueue) {
-			s.validatorBuffer.size.Dec()
-		}
+	s.validatorBuffer.ScheduleNext()
```
cyberphysic4l: I think it's easier to understand that we are scheduling something from every validator's individual queue if you leave it as a loop here. The ScheduleNext() for the whole validator buffer is confusing.

cyberphysic4l (approved):

PR #952, opened 2024-05-02
Fixed prunable rollback deleting the rewards

alexsporn:
If we rolled back to the last slot of an epoch (e.g. when loading a snapshot) on resuming from storage or starting warpsync we wrongly deleted the rewards


hmoog (approved):
LGTM

